PR #318, opened 2021-12-12
Mvn delete

### LINKED ISSUE #277: Maven: delete dependency and property

MSaguer:
Hi,
I propose this PR to close #277 

It adds two maven services
- delete existing version property using its identifier from maven pom
- delete dependency from pom using its package and artifacts id. May content other fields such as scope

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
+++ b/src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
@@ -12,7 +12,7 @@ public interface ProjectRepository {
   void template(Project project, String source, String sourceFilename, String destination, String destinationFilename);
 
   void replaceText(Project project, String source, String sourceFilename, String oldText, String newText);
-  void replaceInFile(Project project, String source, String sourceFilename, String regexpText, String newText);
+  void replaceRegexp(Project project, String source, String sourceFilename, String regexpText, String newText);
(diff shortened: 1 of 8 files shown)


src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
```diff
@@ -12,6 +12,7 @@
   void template(Project project, String source, String sourceFilename, String destination, String destinationFilename);
 
   void replaceText(Project project, String source, String sourceFilename, String oldText, String newText);
+  void replaceInFile(Project project, String source, String sourceFilename, String regexpText, String newText);
```
pascalgrimaud, line 15: I would rename this: replaceInFile -> replaceRegexp
It is to not confuse with replaceText

pascalgrimaud (changes requested):

pascalgrimaud:
Thanks a lot @MSaguer for your contribution. And congrats for keeping the coverage at the max :-)

I added a minor comment. What do you think?

src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
MSaguer: You're right. I forgot to rename it after my last modification. Fix has been pushed.

pascalgrimaud (approved):

Issue #284, opened 2021-12-01
Java: add Sonar configuration


pascalgrimaud:

swarajsaaj:
I propose following approach:-
1. Add plugin configuration in pom.xml
2. Add sonar.properties
3. Add sonar.yml in src/main/docker for sonarqube community server.

@pascalgrimaud would you suggest anything more to be done?

pascalgrimaud:
Yes it's the good approach I think

Anyway:
- I don't know if we have all method to add the configuration in pom.xml. We need to check this part first and do a 1st PR separately if needs
- then, I wonder where to put this. Maybe in server/java/sonar. Not sure it's specific to java, as it could be similar if it was kotlin. Maybe simply server/sonar ?

swarajsaaj:
> I don't know if we have all method to add the configuration in pom.xml. We need to check this part first and do a 1st PR separately if needs

Yes, would enhance the `BuildToolDomainService` with required configurations in plugins

> then, I wonder where to put this. Maybe in server/java/sonar. Not sure it's specific to java, as it could be similar if it was kotlin. Maybe simply server/sonar ?

I agree, its not just `java`, as per current structure it should be `server/sonar`.
Then I think its not limited to only server-side applications, then even `server/javatool` should not be in `server`. Isn't it?

pascalgrimaud:
oh by `server`, it means backend, so I think javatool is in the right place. Anyway, it would be easy to move if needed.

About sonar, we could have another / similar configuration for "client", like Angular, React or Vue later.

So I'm suggesting this:
- put it into `server/sonar` for now, and then later we could move it if it's necessary

What do you think?

swarajsaaj:
Yes, let's start with `server/sonar` and move it later if needed :)
You may assign this ticket to me.

pascalgrimaud:
Not sure if it can help: https://github.com/pascalgrimaud/jhipster-lite/commit/cfd245cfbf042c016f6fec46995ae508be98582b

I did this for JHLite, and to reduce the code smells, before working on pushing an analysis to Sonarcloud

PR #353, opened 2021-12-19
Maven: Add Plugin management and configuration

swarajsaaj:
- Allow maven plugins to be added in `pluginManagement`
- Allow deeper configuration for maven plugins
  - Executions
    - Phase
    - Goals
  - Configuration

This PR is pre-requisite to #284

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Plugin.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Plugin.java
@@ -1,5 +1,7 @@
 package tech.jhipster.lite.generator.buildtool.generic.domain;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import tech.jhipster.lite.error.domain.Assert;
 
@@ -8,6 +10,8 @@ public class Plugin {
   private final String groupId;
   private final String artifactId;
   private final Optional<String> version;
+  private List<PluginExecution> executions;
+  private List<PluginConfiguration> configurations;
 
   private Plugin(PluginBuilder builder) {
     Assert.notBlank("groupId", builder.groupId);
@@ -16,6 +20,8 @@ public class Plugin {
     this.groupId = builder.groupId;
     this.artifactId = builder.artifactId;
     this.version = optionalNotBlank(builder.version);
+    this.executions = builder.executions;
+    this.configurations = builder.configurations;
   }
 
   private Optional<String> optionalNotBlank(String value) {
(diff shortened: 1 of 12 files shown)


pascalgrimaud:
Does it cover all possible configuration? PluginManagement is the most complex part of Maven, and modeling this part, is not easy.

If it's too difficult, there are some alternatives:
- use git patch like https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/springboot/mvc/security/jwt/jwt-security-annotation-processor.patch
- add a needle like you did `<!-- jhipster-needle-maven-add-plugin-management -->` and simply use replace: https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/common/domain/FileUtils.java#L127-L134

swarajsaaj:
> Does it cover all possible configuration? PluginManagement is the most complex part of Maven, and modeling this part, is not easy.

Hi @pascalgrimaud, Yes, I realized its complex :), but wanted to try something better than patch/text-replacement.

This is an attempt to generate the configuration using Java code. Mainly covers `<execution>` and `<configuration>` sections. For `configuration` I have tried to model it into two types of nodes:-
- `PluginConfigurationValue` - Meant for simpler value nodes like `<source.dir>/tmp</source.dir>`
- `PluginConfigurationContainer` - For complex nested tree-like configuration of any plugin, (the `children` of this node can again be of type ``PluginConfigurationValue` or `PluginConfigurationContainer` to allow complicated configurations).

Currently it does not cover `reportSets`, `inherited` and nested `dependencies` within plugin. What do you think of this strategy?

pascalgrimaud:
When I look at the possible generated `pom.xml`, I don't think it's doable in Java: https://github.com/jhipster/jhipster-sample-app/blob/main/pom.xml :-D

So my advice would be to keep as simple as possible: regexp + text block (with Java 17) and it's done! :)

swarajsaaj:
Though its closely possible with this approach as seen in this test case, but maybe its an overkill for something we won't really need too much control on (a replace text might be just okay) :)
 https://github.com/jhipster/jhipster-lite/blob/3a9840296b371d823fa2262f55d97a42609c37b1/src/test/java/tech/jhipster/lite/generator/buildtool/maven/domain/maven/MavenTest.java#L227-L264

pascalgrimaud:
Yes sure it's possible. But as you said, maybe it's overkill. I'd vote for using replace

swarajsaaj:
I ll close this PR, in favour of using a regex-replace method. Will raise a new PR for that.

PR #364, opened 2021-12-21
Fix Code smells

pascalgrimaud:


src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
```diff
@@ -217,7 +217,7 @@ public static String getPlugin(Plugin plugin, int indentation) {
     plugin
       .getAdditionalElements()
       .ifPresent(additionalElements ->
-        result.append(plugin.getAdditionalElements().get().indent(4 * indentation).replaceAll("\n", System.lineSeparator()))
+        result.append(plugin.getAdditionalElements().get().indent(4 * indentation).replace("\n", System.lineSeparator()))
```
pascalgrimaud, line 220: @swarajsaaj : just apply this fix for Sonar code smell. Can you check quickly if it's correct plz?

swarajsaaj: Yes this seems to work fine.

PR #374, opened 2021-12-21
Test: introduce code smells

pascalgrimaud:


pascalgrimaud:
Strange... I don't think it works well with SonarCloud...

Issue #197, opened 2021-11-13
Spring Boot core: add Spring Boot DevTools


pascalgrimaud:

pascalgrimaud:
Closed by https://github.com/jhipster/jhipster-lite/pull/400

PR #400, opened 2021-12-24
Add Spring Boot Developer Tools

pblanchardie:
#197 

I added this dependency as optional, as recommended:

> Flagging the dependency as optional in Maven or using compileOnly in Gradle is a best practice that prevents devtools from being transitively applied to other modules using your project.

In JHipster, devtools were only present with the dev profile. but I wonder if it's really needed, as the documentation says:

> Developer tools are automatically disabled when running a fully packaged application. If your application is launched using java -jar or if it’s started using a special classloader, then it is considered a “production application”.

and

> repackaged archives do not contain devtools by default.

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
@@ -7,6 +7,7 @@ public class Dependency {
 
   private final String groupId;
   private final String artifactId;
+  private final boolean optional;
   private final Optional<String> version;
   private final Optional<String> scope;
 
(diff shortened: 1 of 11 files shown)


pascalgrimaud:
So happy to see you contributing to this project @pblanchardie ! Thanks so much.

I think we could activate it only for fast profile.
See https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/resources/config/application-dev.yml#L24

My ideas in JHipster Lite is to remove dev/prod profile, as it is so confusing. It's not real prod profile.
So I replaced it by:
- default application.properties
- fast profile with application-fast.properties which is used only during development. It will contain log debug, async liquibase, and maybe devtools too etc

pblanchardie:
Hi @pascalgrimaud, you're welcome and congrats for the great work so far.

I was not sure about the goal of the fast profile, thanks for the explanation.

Let's disable both restart and reload in the default and enable them in fast.

Had trouble with prettier, will fix.

pblanchardie:
forgot ResourceIT and BeanConfigurationIT

pascalgrimaud (approved):

pascalgrimaud:
Thanks for your contribution @pblanchardie

Issue #275, opened 2021-12-01
Spring Boot: Async Configuration


pascalgrimaud:

pascalgrimaud:
Closed by https://github.com/jhipster/jhipster-lite/pull/403

PR #403, opened 2021-12-24
add Spring Boot Async Configuration

pblanchardie:
#275 

Will work on that after christmas if it's ok.

Not sure if draft PRs are welcome.

Merry Christmas!


pascalgrimaud:
Everything is welcome!
Merry Christmas you too :-)

pblanchardie:
The second commit (https://github.com/jhipster/jhipster-lite/pull/403/commits/60c508247ebb508896a206cb622bf34061d4824b) adds `ExceptionHandlingAsyncTaskExecutorTest`. It depends on `LogbackRecorder` test utility, which was also used by Liquibase tests.

I moved `LogbackRecorder` to the base package, and created a `SpringBootCommonService` (inspired by the shared Properties service) that is responsible for addition of this shared utility. It is now used by both Async and Liquibase.
If `LogbackRecorder` was already created, it is silently overriden (default behavior).

I don't know if it's the preferred way of doing things, but alternatives were not satisfying:
- duplicate LogbackRecorder in async test package: leads to code duplication, and this utility could be used by other tests
- omit LogBackRecorder in async testing: not a good reason to omit testing
- simply duplicate the same code in async and liquibase domain services to generate LogBackRecorder: that what I've done at the beginning, it's simple, but sharing became natural once I added a SOURCE constant for the path...

Btw, I fixed some `public` fields that should have been `private` in liquibase services.

pascalgrimaud:
Thanks a lot @pblanchardie !
Could you call the API in https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh plz?

I'll review your work soon :-)

src/main/java/tech/jhipster/lite/generator/server/springboot/async/infrastructure/primary/rest/SpringBootAsyncResource.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import tech.jhipster.lite.generator.project.domain.Project;
+import tech.jhipster.lite.generator.project.infrastructure.primary.dto.ProjectDTO;
+import tech.jhipster.lite.generator.server.springboot.async.application.SpringBootAsyncApplicationService;
+
+@RestController
+@RequestMapping("/api/servers/spring-boot/async")
+@Tag(name = "Spring Boot - Async")
```
pascalgrimaud, line 16: what do you think about putting this under "Spring Boot" tag ? As I don't think we'll have another alternative under "async" tag

src/main/resources/generator/server/springboot/async/test/AsyncConfigurationIT.java.mustache
```diff
@@ -0,0 +1,20 @@
+package {{packageName}}.technical.infrastructure.secondary.async;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.mycompany.myapp.IntegrationTest;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
```
pascalgrimaud, line 8:
```suggestion
import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.ApplicationContext;
import {{packageName}}.IntegrationTest;
```

src/main/resources/generator/server/springboot/async/test/ExceptionHandlingAsyncTaskExecutorTest.java.mustache
```diff
@@ -0,0 +1,267 @@
+package {{packageName}}.technical.infrastructure.secondary.async;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.mockito.Mockito.*;
+
+import com.mycompany.myapp.LogbackRecorder;
+import com.mycompany.myapp.LogbackRecorder.Event;
```
pascalgrimaud, line 8:
```suggestion
import {{packageName}}.LogbackRecorder;
import {{packageName}}.LogbackRecorder.Event;
```

src/main/resources/generator/server/springboot/async/test/ExceptionHandlingAsyncTaskExecutorTest.java.mustache
```diff
@@ -0,0 +1,267 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.core.task.AsyncTaskExecutor;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
+
```
pascalgrimaud, line 20:
```suggestion
import org.springframework.core.task.SimpleAsyncTaskExecutor;
import {{packageName}}.UnitTest;

@UnitTest
```

src/main/resources/generator/server/springboot/async/src/ExceptionHandlingAsyncTaskExecutor.java.mustache
```diff
@@ -0,0 +1,123 @@
...
+package {{packageName}}.technical.infrastructure.secondary.async;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.core.task.AsyncTaskExecutor;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
```
pascalgrimaud, line 29:
```suggestion
import java.util.concurrent.Callable;
import java.util.concurrent.Future;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.DisposableBean;
import org.springframework.beans.factory.InitializingBean;
import org.springframework.core.task.AsyncTaskExecutor;
```

pascalgrimaud (changes requested):

src/main/resources/generator/server/springboot/async/src/ExceptionHandlingAsyncTaskExecutor.java.mustache
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * <p>ExceptionHandlingAsyncTaskExecutor class.</p>
+ */
+public class ExceptionHandlingAsyncTaskExecutor implements AsyncTaskExecutor,
```
pascalgrimaud, line 34: plz, indent with 2 spaces:

```java
public class ExceptionHandlingAsyncTaskExecutor implements AsyncTaskExecutor, InitializingBean, DisposableBean {

  static final String EXCEPTION_MESSAGE = "Caught async exception";

  private final Logger log = LoggerFactory.getLogger(ExceptionHandlingAsyncTaskExecutor.class);

  private final AsyncTaskExecutor executor;

  /**
   * <p>Constructor for ExceptionHandlingAsyncTaskExecutor.</p>
   *
   * @param executor a {@link org.springframework.core.task.AsyncTaskExecutor} object.
   */
  public ExceptionHandlingAsyncTaskExecutor(AsyncTaskExecutor executor) {
    this.executor = executor;
  }

  /** {@inheritDoc} */
  @Override
  public void execute(Runnable task) {
    executor.execute(createWrappedRunnable(task));
  }

  /** {@inheritDoc} */
  @Override
  public void execute(Runnable task, long startTimeout) {
    executor.execute(createWrappedRunnable(task), startTimeout);
  }

  private <T> Callable<T> createCallable(final Callable<T> task) {
    return () -> {
      try {
        return task.call();
      } catch (Exception e) {
        handle(e);
        throw e;
      }
    };
  }

  private Runnable createWrappedRunnable(final Runnable task) {
    return () -> {
      try {
        task.run();
      } catch (Exception e) {
        handle(e);
      }
    };
  }

  /**
   * <p>handle.</p>
   *
   * @param e a {@link java.lang.Exception} object.
   */
  protected void handle(Exception e) {
    log.error(EXCEPTION_MESSAGE, e);
  }

  /** {@inheritDoc} */
  @Override
  public Future<?> submit(Runnable task) {
    return executor.submit(createWrappedRunnable(task));
  }

  /** {@inheritDoc} */
  @Override
  public <T> Future<T> submit(Callable<T> task) {
    return executor.submit(createCallable(task));
  }

  /** {@inheritDoc} */
  @Override
  public void destroy() throws Exception {
    if (executor instanceof DisposableBean) {
      DisposableBean bean = (DisposableBean) executor;
      bean.destroy();
    }
  }

  /** {@inheritDoc} */
  @Override
  public void afterPropertiesSet() throws Exception {
    if (executor instanceof InitializingBean) {
      InitializingBean bean = (InitializingBean) executor;
      bean.afterPropertiesSet();
    }
  }
}
```

pascalgrimaud (changes requested):

pascalgrimaud:
Agree with your refactoring for `LogbackRecorder` :-)

src/main/java/tech/jhipster/lite/generator/server/springboot/async/infrastructure/primary/rest/SpringBootAsyncResource.java
pblanchardie: Sure

pblanchardie:
After all, perhaps we could simply abandon ExceptionHandlingAsyncTaskExecutor to stick with standard behaviors and avoid custom implementations.
see https://github.com/jhipster/generator-jhipster/issues/2714

> Is this code still useful ? This was added at the beginning of the project and I don't even know if this is useful.

> Maybe today there is no need for such executor as it is easy to set unhandled exception handler

It was added in 2014: https://github.com/jhipster/generator-jhipster/commit/c0f3b37538243659dd14b09ea7374042b39f041c

pascalgrimaud:
@pblanchardie : sorry it's totally my fault. I created the ticket with the goal to generate step by step what generator-jhipster generates.
I didn't check each class.

Indeed, if this `ExceptionHandlingAsyncTaskExecutor` is not necessary, we can forgot it.

I have the same problem with a lot of developers: I don't understand all classes generated by geneator-jhipster. So by coding JHipster Lite, I'll understand step by step what I strictly need to generate for my projects.

pblanchardie:
Yes it's a good opportunity to do some cleanup in our JHipster projects!

Not sure if you want to go more into details, but talking about "step-by-step" and understanding of what happens, we could have separated Async and Scheduling.

If this issue is about Async task execution, only EnableAsync is relevent.

Task scheduling is apart, with EnableScheduling and 2 specific properties.

pascalgrimaud:
Yes it can come together. Later if needed, we can easily separate it.

pascalgrimaud (approved):

PR #431, opened 2021-12-28
add Spring Boot Ehcache (XML configuration)

pblanchardie:
#121 (alternative to #432 )

Here is a version with JSR-107 Ehcache XML configuration.

[Getting JSR-107 caches configured through Ehcache XML](https://www.ehcache.org/documentation/3.0/107.html#getting-jsr-107-caches-configured-through-ehcache-xml)

[Spring Boot Caching with JCache](https://docs.spring.io/spring-boot/docs/current/reference/html/io.html#io.caching.provider.jcache)

There are many pros with XML configuration on both the generator and generated sides, but there are probably cons.

Generator:
- [x] Ehcache
- [x] Rename to xml-configuration

Generated:
- [x] ehcache.xml with template for entities
- [x] jhipster-needle-ehcache-add-entity
note the explicit "entity" needle because it uses a template to have a shared ttl and max-entries
- [ ] ~~integration test of the configuration? see [Supplement JSR-107’s configurations](https://www.ehcache.org/documentation/3.0/107.html#supplement-jsr-107-configurations)~~

---

This is based on a common branch: https://github.com/pblanchardie/jhipster-lite/tree/add-spring-boot-jcache

- [x] Spring Boot Cache Commons (spring-boot-starter-cache + CacheConfiguration)
- [x] Spring Boot JCache Commons (jcache)

Issue #121, opened 2021-10-09
Spring Boot Cache: ehcache support


pascalgrimaud:

pblanchardie:
Work in progress.

pascalgrimaud:
More work than expected, so increasing the bounty
- cache/common shared kernel
- cache/simple
- cache/ehcache

pascalgrimaud:
Thanks a lof for all your work here @pblanchardie !

pblanchardie:
You're welcome, thanks for your help!

pblanchardie:
@pascalgrimaud https://opencollective.com/generator-jhipster/expenses/60551

pascalgrimaud:
@pblanchardie : approved, thanks so much for helping this project! <3

PR #432, opened 2021-12-29
add Spring Boot Ehcache (Java Configuration)

pblanchardie:
#121 (alternative to #431)

This is the version with Java configuration.

Quite verbose, will less flexibility than the XML configuration but a common needle for all JCache implementations.

Generator:

- [x] Ehcache
- [x] rename to java-configuration

Generated:

- [x] EhcacheConfiguration
- [x] jhipster-needle-**jcache**-add-**entity**
note 1: "jcache" needle because we can add entities at a single place for all JCache providers
note 2: explicit "entity" needle because it uses a template to have a shared ttl and max-entries
- [x] integration test of the configuration (2 properties...)

---

This is based on a common branch: https://github.com/pblanchardie/jhipster-lite/tree/add-spring-boot-jcache

- [x] Spring Boot Cache Commons (spring-boot-starter-cache + CacheConfiguration)
- [x] Spring Boot JCache Commons (jcache)
- [x] Spring Boot JCache Java Config (Java configuration)

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
@@ -1,0 +1,32 @@
+package tech.jhipster.lite.generator.server.springboot.cache.ehcache.infrastructure.primary.rest;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import tech.jhipster.lite.generator.project.domain.Project;
+import tech.jhipster.lite.generator.project.infrastructure.primary.dto.ProjectDTO;
+import tech.jhipster.lite.generator.server.springboot.cache.ehcache.application.EhcacheApplicationService;
+
+@RestController
+@RequestMapping("/api/servers/spring-boot/cache/ehcache")
+@Tag(name = "Spring Boot - Cache")
+class EhcacheResource {
+
+  private final EhcacheApplicationService ehcacheApplicationService;
+
+  public EhcacheResource(EhcacheApplicationService ehcacheApplicationService) {
+    this.ehcacheApplicationService = ehcacheApplicationService;
+  }
+
+  @Operation(summary = "Add Ehcache")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding Ehcache")
+  @PostMapping
+  public void init(@RequestBody ProjectDTO projectDTO) {
+    Project project = ProjectDTO.toProject(projectDTO);
+    ehcacheApplicationService.init(project);
+  }
+}
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
@@ -24,9 +24,9 @@ class EhcacheResource {
 
   @Operation(summary = "Add Ehcache")
   @ApiResponse(responseCode = "500", description = "An error occurred while adding Ehcache")
-  @PostMapping
-  public void init(@RequestBody ProjectDTO projectDTO) {
+  @PostMapping("java-configuration")
+  public void initJavaConfiguration(@RequestBody ProjectDTO projectDTO) {
     Project project = ProjectDTO.toProject(projectDTO);
-    ehcacheApplicationService.init(project);
+    ehcacheApplicationService.initJavaConfiguration(project);
   }
 }
--- a/src/main/resources/generator/server/springboot/cache/ehcache/test/EhcacheConfigurationIT.java.mustache
+++ b/src/main/resources/generator/server/springboot/cache/ehcache/test/EhcacheConfigurationIT.java.mustache
@@ -1,0 +1,51 @@
+package {{packageName}}.technical.infrastructure.secondary.cache.ehcache;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import {{packageName}}.IntegrationTest;
+import {{packageName}}.technical.infrastructure.secondary.cache.JCacheConfigurer;
+import {{packageName}}.technical.infrastructure.secondary.cache.JCacheCreator;
+import javax.cache.Cache;
+import javax.cache.CacheManager;
+import org.ehcache.config.CacheRuntimeConfiguration;
+import org.ehcache.config.ResourceType;
+import org.ehcache.jsr107.Eh107Configuration;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ApplicationContext;
+
+@IntegrationTest
+@SpringBootTest(properties = { "application.cache.ehcache.max-entries=1000", "application.cache.ehcache.time-to-live-seconds=86400" })
+class EhcacheConfigurationIT {
+
+  @Autowired
+  ApplicationContext applicationContext;
+
+  @Autowired
+  CacheManager cm;
+
+  @Autowired
+  JCacheCreator jCacheCreator;
+
+  @Test
+  void shouldEhcacheConfigurer() {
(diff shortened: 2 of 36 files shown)


src/main/resources/generator/server/springboot/cache/jcache/src/CacheProperties.java.mustache
```diff
@@ -0,0 +1,6 @@
+package {{packageName}}.technical.infrastructure.secondary.cache;
+
+public interface CacheProperties {
+  long getMaxEntries();
+  int getTimeToLiveSeconds();
```
pblanchardie, line 5: pas besoin de ces 2 propriétés dans l'interface, je vais laisser la liberté en fonction de l'implémentation

pblanchardie:
will rebase

src/main/resources/generator/server/springboot/cache/jcache/src/JCacheConfiguration.java.mustache
```diff
@@ -0,0 +1,22 @@
+package {{packageName}}.technical.infrastructure.secondary.cache;
+
+import org.springframework.boot.autoconfigure.cache.JCacheManagerCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JCacheConfiguration {
+
+  private final JCacheCreator jCacheCreator;
```
pascalgrimaud, line 10:
```suggestion
  @SuppressWarnings("java:S1068") // Not used without entity
  private final JCacheCreator jCacheCreator;
```

pascalgrimaud (approved):

src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
```diff
@@ -0,0 +1,32 @@
...
+
+  private final EhcacheApplicationService ehcacheApplicationService;
+
+  public EhcacheResource(EhcacheApplicationService ehcacheApplicationService) {
+    this.ehcacheApplicationService = ehcacheApplicationService;
+  }
+
+  @Operation(summary = "Add Ehcache")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding Ehcache")
+  @PostMapping
```
pascalgrimaud, line 27:
```suggestion
  @PostMapping("/java-configuration")
```

src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
```diff
@@ -0,0 +1,32 @@
...
+  public void init(@RequestBody ProjectDTO projectDTO) {
```
pascalgrimaud, line 28:
```suggestion
  public void initJavaConfiguration(@RequestBody ProjectDTO projectDTO) {
```

src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
```diff
@@ -0,0 +1,32 @@
...
+    Project project = ProjectDTO.toProject(projectDTO);
+    ehcacheApplicationService.init(project);
```
pascalgrimaud, line 30:
```suggestion
    ehcacheApplicationService.initJavaConfiguration(project);
```

src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
pascalgrimaud: just in case you want to keep both solution: JavaConfiguration + XML

pascalgrimaud (approved):

pascalgrimaud:
I still prefer this version, over XML one :)
But we can keep both, see my comments

As soon as the last suppressWarning is added, I can merge it, unless you want to rename some methods for keeping both solutions.

src/main/java/tech/jhipster/lite/generator/server/springboot/cache/ehcache/infrastructure/primary/rest/EhcacheResource.java
pblanchardie: Sure, I'll do the appropriate refactoring

pblanchardie:
looks like action failed for a mysterious reason, re-running job could help

src/main/resources/generator/server/springboot/cache/ehcache/test/EhcacheConfigurationIT.java.mustache
```diff
@@ -0,0 +1,51 @@
...
+import org.ehcache.config.CacheRuntimeConfiguration;
+import org.ehcache.config.ResourceType;
+import org.ehcache.jsr107.Eh107Configuration;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ApplicationContext;
+
+@IntegrationTest
+@SpringBootTest(properties = { "application.cache.ehcache.max-entries=1000", "application.cache.ehcache.time-to-live-seconds=86400" })
```
pascalgrimaud, line 19: I think the problem comes from these 2 same annotations.
I'll submit another PR to improve IntegrationTest, so we can use:

```
@IntegrationTest(properties = { "application.cache.ehcache.max-entries=1000", "application.cache.ehcache.time-to-live-seconds=86400" })
```

pblanchardie: fixed by #453

pascalgrimaud:
@pblanchardie : I'm playing a little bit with this version, to add new custom cache and it was a little bit more complex than the one from generator-jhipster

My question is:
- why not using a single file for CacheConfiguration, which should look like: https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/java/io/github/jhipster/sample/config/CacheConfiguration.java
- because, with this version, I can:
  - add a new cache with specifc value: https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/java/io/github/jhipster/sample/config/CacheConfiguration.java#L24
  - build it in constructor: https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/java/io/github/jhipster/sample/config/CacheConfiguration.java#L29-L35
  - and create the cache: https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/java/io/github/jhipster/sample/config/CacheConfiguration.java#L56

pblanchardie:
> why not using a single file for CacheConfiguration

Short answer: sure we can, it would be simpler on the generated side.

Long answer: I wasn't comfortable with the single-CacheConfiguration.
- code and test duplication in the generator for each cache provider
- hard to test with private methods (make them package-private and it should be ok)
- mixing "declarative" configuration with logic

Yes, the drawback is that the generated code has more Java classes, but you can still add a custom caches the same way than before in JCacheConfiguration (at jhipster-needle-jcache...) by using the utility methods of the injected `jCacheCreator`, which is provider-agnostic but accepts any Configuration, and the code of this JCacheConfiguration class is shorter.
All the provider-specific code is isolated in the relevant subpackage, as we leverage JCache abstraction.
Indeed, if you want to have multiple cache configurations, then better go with the XML file or single CacheConfiguration so everything is at the same place.

The current implementation can seem overkill unless you want to support multiple cache providers.

We can get rid of this and switch to a single CacheConfiguration, but we'll have the same boilerplate code in each cache provider feature of the generator.
If the ultimate goal is to keep the generated code stupid simple, that's probably a better choice.

pascalgrimaud:
Not sure we'll have exactly the same thing in other cache provider. See the complex file at https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs

Honestly, in real project, I don't see any value to have multiple cache at the same time. It would be really strange

So 2 solutions I can see here:

Keep both options and add a 3rd one:
- xml
- java config complete
- java config single <---

I'm afraid it's too much but it's doable as JH Lite is done for that

2nd solution:
- xml
- replace multiple java config by a single one

Which solution do you prefer?
I can code it if you want

pblanchardie:
I kept pieces of my previous attempt with single CacheConfiguration. Will give it a try and submit a PR that you can adjust if needed.

I also think we should have only 1 java option as they have no differences in terms of functionality (the current design was mainly focused on improving the generator side).

pascalgrimaud:
> I kept pieces of my previous attempt with single CacheConfiguration. Will give it a try and submit a PR that you can adjust if needed

So thanks a lot.
I'm working on generator-jhipster release, as:
- jhipster-lite took me less than 10sec to launch the release and 3min to write the tweet
- generator-jhipster will take at least 4h...

Issue #279, opened 2021-12-01
Spring Boot: Spring Cloud Config Client


pascalgrimaud:

swarajsaaj:
In generator-jhipster, the Jhipster registry (https://github.com/jhipster/jhipster-registry) was shipped together. I think we can use the same (not just for config server, also for the eureka registry part). Any opinions for not using that and using https://hub.docker.com/r/springcloud/configserver instead?

swarajsaaj:
Instead of microservice, it can be `spring-boot/spring-cloud/config-client` and in addition following approach:-

- [ ] Add config-client dependency
- [ ] Setup `spring-cloud/config-client` context
- [ ] Add `jhipster-registry.yml` in `src/main/docker`
- [ ] Add properties for connecting to config-client

pascalgrimaud:
Look good to me, excepting there are 2 possibilities:
- jhipster registry as config server
- consul as config server

swarajsaaj:
For consul as config provider (with 'spring-cloud-starter-consul-config'), it has pre-requisite that consul is in place already as using only config KV without registry capabilities is unusual.
Can add the config as an option in Consul context. WDYT?

pascalgrimaud:
Mhhh it's related to another ticket.
Let's start with your 1st idea + jhipster registry, we'll see later for spring cloud consul

pascalgrimaud:
@swarajsaaj : so I had a look and indeed you're right - see https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/resources/config/application-dev.yml.ejs#L93-L114

There are 2 configs:
- spring.cloud.consul.config
- spring.cloud.config.uri

So this ticket should be related only to spring cloud config (not consul)
I wonder if we should implement this ticket in the same context than Eureka ?
- API to generate Eureka + Spring Cloud Config
- API to generate only Spring Cloud Config

It means this ticket https://github.com/jhipster/jhipster-lite/issues/280 can only be done after this one.
What do you think?

swarajsaaj:
@pascalgrimaud : 
>So this ticket should be related only to spring cloud config (not consul)

Yes agreed, as I have seen projects using only config-server for dynamic configuration without need of eureka at all (like in kubernetes based deployments, where you may not need additional registry but dynamic configuration is required instead of configmaps)

[...]

PR #435, opened 2021-12-30
Spring Boot: Spring Cloud Consul

### LINKED ISSUE #281: Spring Boot: Spring Cloud Consul

pblanchardie:
fixes #281 

- [x] Consul bootstrap, config and discovery base
- [x] Add Spring Cloud to dependency management
- [x] Add more Consul properties
- [x] Use the same group as #279 - "spring-cloud"
- [x] Add Docker Consul (without jhipster/consul-config-loader)

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
@@ -13,12 +13,24 @@ public class Consul {
 
   private static final String SPRING_CLOUD = "org.springframework.cloud";
 
-  // TODO
-  // currently hard-coded in pom template - move to commons cloud?
-  // private static final String SPRING_CLOUD_VERSION = "2021.0.0";
-  // public static String getSpringCloudVersion() {
-  //  return SPRING_CLOUD_VERSION;
-  //}
+  private static final String SPRING_CLOUD_VERSION = "2021.0.0";
+
+  private Consul() {}
+
+  public static String getSpringCloudVersion() {
+    return SPRING_CLOUD_VERSION;
+  }
+
+  public static Dependency springCloudDependencyManagement() {
+    return Dependency
+      .builder()
+      .groupId(SPRING_CLOUD)
+      .artifactId("spring-cloud-dependencies")
+      .version("\\${spring-cloud.version}")
+      .type("pom")
+      .scope("import")
+      .build();
+  }
 
   public static Dependency springCloudBootstrapDependency() {
     return Dependency.builder().groupId(SPRING_CLOUD).artifactId("spring-cloud-starter-bootstrap").build();
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
@@ -7,10 +7,6 @@ public class Consul {
   public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
   public static final String BOOTSTRAP_FAST_PROPERTIES = "bootstrap-fast.properties";
 
-  public static final String NEEDLE_BOOTSTRAP_PROPERTIES = "# jhipster-needle-bootstrap-properties";
-  public static final String NEEDLE_BOOTSTRAP_FAST_PROPERTIES = "# jhipster-needle-bootstrap-fast-properties";
-  public static final String NEEDLE_BOOTSTRAP_TEST_PROPERTIES = "# jhipster-needle-bootstrap-test-properties";
-
   private static final String SPRING_CLOUD = "org.springframework.cloud";
 
   private static final String SPRING_CLOUD_VERSION = "2021.0.0";
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
@@ -1,0 +1,34 @@
+package tech.jhipster.lite.generator.server.springboot.consul.domain;
+
+import tech.jhipster.lite.generator.buildtool.generic.domain.Dependency;
+
+public class Consul {
+
+  public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+  public static final String BOOTSTRAP_FAST_PROPERTIES = "bootstrap-fast.properties";
+
+  public static final String NEEDLE_BOOTSTRAP_PROPERTIES = "# jhipster-needle-bootstrap-properties";
+  public static final String NEEDLE_BOOTSTRAP_FAST_PROPERTIES = "# jhipster-needle-bootstrap-fast-properties";
+  public static final String NEEDLE_BOOTSTRAP_TEST_PROPERTIES = "# jhipster-needle-bootstrap-test-properties";
+
+  private static final String SPRING_CLOUD = "org.springframework.cloud";
+
+  // TODO
+  // currently hard-coded in pom template - move to commons cloud?
+  // private static final String SPRING_CLOUD_VERSION = "2021.0.0";
+  // public static String getSpringCloudVersion() {
+  //  return SPRING_CLOUD_VERSION;
+  //}
+
+  public static Dependency springCloudBootstrapDependency() {
+    return Dependency.builder().groupId(SPRING_CLOUD).artifactId("spring-cloud-starter-bootstrap").build();
+  }
+
+  public static Dependency springCloudConsulDiscoveryDependency() {
+    return Dependency.builder().groupId(SPRING_CLOUD).artifactId("spring-cloud-starter-consul-discovery").build();
+  }
+
+  public static Dependency springCloudConsulConfigDependency() {
+    return Dependency.builder().groupId(SPRING_CLOUD).artifactId("spring-cloud-starter-consul-config").build();
+  }
+}
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
@@ -7,11 +7,15 @@ public class Consul {
   public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
   public static final String BOOTSTRAP_FAST_PROPERTIES = "bootstrap-fast.properties";
 
+  private static final String DOCKER_CONSUL_IMAGE = "consul:1.11.1";
   private static final String SPRING_CLOUD = "org.springframework.cloud";
-
   private static final String SPRING_CLOUD_VERSION = "2021.0.0";
 
   private Consul() {}
+
+  public static String getDockerConsulImage() {
+    return DOCKER_CONSUL_IMAGE;
+  }
 
   public static String getSpringCloudVersion() {
     return SPRING_CLOUD_VERSION;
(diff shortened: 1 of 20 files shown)


src/main/java/tech/jhipster/lite/generator/server/springboot/consul/infrastructure/primary/rest/ConsulResource.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import tech.jhipster.lite.generator.project.domain.Project;
+import tech.jhipster.lite.generator.project.infrastructure.primary.dto.ProjectDTO;
+import tech.jhipster.lite.generator.server.springboot.consul.application.ConsulApplicationService;
+
+@RestController
+@RequestMapping("/api/servers/spring-boot/cloud/consul")
```
swarajsaaj, line 15: We can keep `cloud` for public clouds (azure,gcp,aws) and use spring-cloud here `/api/servers/spring-boot/spring-cloud/consul`?

pblanchardie: Done.

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
```diff
@@ -0,0 +1,87 @@
...
+    );
+    addProperty(project, "spring.cloud.consul.enabled", "false", TEST_RESOURCES, BOOTSTRAP_PROPERTIES, NEEDLE_BOOTSTRAP_TEST_PROPERTIES);
+  }
+
+  private Map<String, Object> properties() {
+    Map<String, Object> result = new LinkedHashMap<>();
+
+    result.put("spring.cloud.consul.discovery.health-check-path", "/management/health");
+    result.put("spring.cloud.consul.discovery.tags[0]", "version=@project.version@");
+    result.put("spring.cloud.consul.discovery.tags[1]", "context-path=\\$\\{server.servlet.context-path:\\}");
```
pblanchardie, line 69: taken from JHipster generator but not tested

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
```diff
@@ -0,0 +1,46 @@
...
+import tech.jhipster.lite.generator.buildtool.generic.domain.Dependency;
+
+public class Consul {
+
+  public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+  public static final String BOOTSTRAP_FAST_PROPERTIES = "bootstrap-fast.properties";
+
+  public static final String NEEDLE_BOOTSTRAP_PROPERTIES = "# jhipster-needle-bootstrap-properties";
+  public static final String NEEDLE_BOOTSTRAP_FAST_PROPERTIES = "# jhipster-needle-bootstrap-fast-properties";
+  public static final String NEEDLE_BOOTSTRAP_TEST_PROPERTIES = "# jhipster-needle-bootstrap-test-properties";
```
swarajsaaj, line 12: As these are generic (also needed for eureka and config-server), these can be used from `SpringBootPropertiesApplicationService` (moved this in #457)

pascalgrimaud: not sure about needle for bootstrap. Look at https://github.com/jhipster/jhipster-sample-app-microservice/blob/main/src/main/resources/config/bootstrap.yml

There is no needle here and I don't think we need its

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
```diff
@@ -0,0 +1,46 @@
+package tech.jhipster.lite.generator.server.springboot.consul.domain;
+
+import tech.jhipster.lite.generator.buildtool.generic.domain.Dependency;
+
+public class Consul {
+
+  public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+  public static final String BOOTSTRAP_FAST_PROPERTIES = "bootstrap-fast.properties";
```
pascalgrimaud, line 8: no need fast profile here

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
```diff
@@ -0,0 +1,87 @@
+package tech.jhipster.lite.generator.server.springboot.consul.domain;
+
+import static tech.jhipster.lite.common.domain.FileUtils.getPath;
+import static tech.jhipster.lite.generator.project.domain.Constants.MAIN_RESOURCES;
+import static tech.jhipster.lite.generator.project.domain.Constants.TEST_RESOURCES;
+import static tech.jhipster.lite.generator.project.domain.DefaultConfig.BASE_NAME;
+import static tech.jhipster.lite.generator.server.springboot.consul.domain.Consul.*;
+import static tech.jhipster.lite.generator.server.springboot.core.domain.SpringBootDomainService.CONFIG_FOLDER;
```
pascalgrimaud, line 8: strange: this import should not be possible with HegaxonalArchTest as springboot.core is a BusinessContext

Then, maybe I would suggest to move CONFIG_FOLDER to Constants ?

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
```diff
@@ -0,0 +1,87 @@
...
+    buildToolService.addDependency(project, springCloudConsulConfigDependency());
+    buildToolService.addDependency(project, springCloudConsulDiscoveryDependency());
+  }
+
+  @Override
+  public void addProperties(Project project) {
+    project.addDefaultConfig(BASE_NAME);
+
+    projectRepository.template(project, getPath(SOURCE, "src"), "bootstrap.properties", getPath(MAIN_RESOURCES, CONFIG_FOLDER));
+    projectRepository.template(project, getPath(SOURCE, "src"), "bootstrap-fast.properties", getPath(MAIN_RESOURCES, CONFIG_FOLDER));
```
pascalgrimaud, line 48: no need

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
```diff
@@ -0,0 +1,87 @@
...
+    projectRepository.template(project, getPath(SOURCE, "test"), "bootstrap.properties", getPath(TEST_RESOURCES, CONFIG_FOLDER));
+
+    properties().forEach((k, v) -> addProperty(project, k, v, MAIN_RESOURCES, BOOTSTRAP_PROPERTIES, NEEDLE_BOOTSTRAP_PROPERTIES));
+
+    addProperty(
```
pascalgrimaud, line 53: no need this as comment above. We should use bootstrap.properties.mustache and generate it like this

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
pblanchardie: I found it very convenient to be able to disable Consul during development.
Can you elaborate?

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
pblanchardie: yes, didn't notice, I thought I was importing SpringBootCommonDomainService which is a shared kernel.
up to you if we move that to Constants or SpringBootCommonDomainService or other.

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
pblanchardie: ho yes we should get rid of these needles, it will be simpler if everything is in the mustache template as I don't think we'll add anything in bootstrap properties in other contexts.

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
pblanchardie: absolutely

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
pascalgrimaud: If during development, you don't use consul, you won't see other microservices. So when you start your microservice1, you won't be able to communicate with microservice2

So you always need consul when you start you app.

Don't hesitate to tell me if I'm not clear enough

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/ConsulDomainService.java
pascalgrimaud: Constant like other MAIN and RESOURCES etc would be enough

src/main/java/tech/jhipster/lite/generator/server/springboot/consul/domain/Consul.java
pblanchardie: The problem is especially with config, not discovery.

During local development, it can be convenient to allow starting the app without Consul Config, and that was allowed by the `failFast` property that you'll probably prefer and which was in generator-jhipster.
https://cloud.spring.io/spring-cloud-consul/reference/html/#spring-cloud-consul-failfast

The difference is that Consul will still be enabled during local dev, but if it's not available for configuration, you'll still be able to start your app. If it's available, it will be used. I understand that it's a better choice.

pascalgrimaud: Oh I see. Then let's give it a try, and I'll test it on my real projects :)

pblanchardie: done, only for config.

pascalgrimaud:
so it looks ready, I'll review it tomorrow @pblanchardie

pascalgrimaud:
Just gave a quick try:

image

pascalgrimaud:
One issue I saw, is the indentation, which should be fixed in another PR than this one:

image

pascalgrimaud:
Missing this folder `src/main/docker/central-server-config/` with a `application.properties` file

This file can contain: 

```
configserver.name=Docker Consul Service
configserver.status=Connected to Consul Server running in Docker
``

Then, these information can be displayed when starting the app maybe ?

```java
        String configServerStatus = env.getProperty("configserver.status");
        if (configServerStatus == null) {
            configServerStatus = "Not found or not setup for this application";
        }
        log.info(
            "\n----------------------------------------------------------\n\t" +
            "Config Server: \t{}\n----------------------------------------------------------",
            configServerStatus
        );
```

pblanchardie:
> One issue I saw, is the indentation, which should be fixed in another PR than this one:

Yes, seen this indentation problem. 

> Missing this folder src/main/docker/central-server-config/ with a application.properties file

Soit's just to be sure that config works?
Is `configserver.name` used somewhere?

pascalgrimaud:
For indentation, the fix could be similar to https://github.com/jhipster/jhipster-lite/pull/412/files -> but I'm not sure

`configserver.name` is only used to be sure that the application pull the config from Consul.
See https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/docker/config/consul-config/application.yml.ejs#L19-L21

pblanchardie:
> For indentation, the fix could be similar to https://github.com/jhipster/jhipster-lite/pull/412/files -> but I'm not sure

see #478

pascalgrimaud (approved):

PR #436, opened 2021-12-30
Maven resources filtering config

pblanchardie:
apply filtering to `config/*.properties`
needed by #435 for `config/bootstrap.properties`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,21 @@
   </dependencies>
 
   <build>
+    <resources>
+      <resource>
+        <directory>${basedir}/src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>config/*.properties</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>${basedir}/src/main/resources</directory>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -136,6 +136,21 @@
...
+        <directory>${basedir}/src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>config/*.properties</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>${basedir}/src/main/resources</directory>
+        <excludes>
+          <exclude>config/*.properties</exclude>
```
pascalgrimaud, line 150: just wonder if it's correct, as both directories are identical, and include/exclude are identical too :D

pblanchardie: The second block tells maven to exclude other files from filtering, as it defaults to false, but to still process them. It is needed.
Some prefer to make it explicit by adding `<filtering>true</filtering>` but it's redundant, but it's not the case of spring-boot-started-parent.

pascalgrimaud (approved):

PR #445, opened 2021-12-31
Reduce logs during tests

pascalgrimaud:


pascalgrimaud:
Still a lot of logs during integration test

PR #439, opened 2021-12-30
MavenService.addDependencyManagement

pblanchardie:
needed by #435

PR #448, opened 2021-12-31
BuildToolService.addDependencyManagement

pblanchardie:
BuildToolService.addDependencyManagement was missing in #439 sry

Also added a missing test for optional dependencies

Needed by #435

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
@@ -8,9 +8,9 @@ public class Dependency {
   private final String groupId;
   private final String artifactId;
   private final boolean optional;
-  private final String version;
-  private final String scope;
-  private final String type;
+  private final Optional<String> version;
+  private final Optional<String> scope;
+  private final Optional<String> type;
 
   private Dependency(Dependency.DependencyBuilder builder) {
     Assert.notBlank("groupId", builder.groupId);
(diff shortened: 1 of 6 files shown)


src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
```diff
@@ -8,8 +8,9 @@
   private final String groupId;
   private final String artifactId;
   private final boolean optional;
-  private final Optional<String> version;
```
pascalgrimaud, line 11: Using optional here is intended and really important when working in Domain.

If the field is mandatory, use String xxx.
Otherwise, we must use an Optional.

By doing this, no need to check null later, I mean I can call dependency.getGroupId, I know it won't never be null.
But I can use directly dependency.getVersion as it can be empty

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/Dependency.java
pascalgrimaud: Another comment too: when looking all the fields, I can know immediatly which ones are mandatory or not

pblanchardie: I prefer when it's the getter's job to wrap nulls with Optional.ofNullable so Optional is never used as a field type (anti-pattern) and no need to check null later. Did you see?

I admit it allows to quickly see optional fields, and the private constructor secures the fields, and the class is not Serializable so "it works".
I can revert after you confirm the intended use of Optional on fields.

pascalgrimaud: In real project, the domain is more complex and I can have a lot of other methods in the Domain class. It can lead to the use of the field version directly  instead of getVersion() so it can lead immediatly to null pointer exception if the dev didn't check it

That's the reason why I prefer to protect all the fields of a domain class.

pascalgrimaud: So yes, plz revert this part. I'm still convinced of the use of Optional, we don't care about Serialization here

pblanchardie: Thanks for the explanation, reverting.

pascalgrimaud (approved):

Issue #440, opened 2021-12-30
Use maven to read and write pom.xml


pblanchardie:
hand-made regex-based pom.xml generation code is error-prone.

why not using [Maven](https://maven.apache.org) itself for proper pom.xml management and validation?

at least we can rely on base objects like [Dependency](https://maven.apache.org/ref/3.5.2/maven-model/apidocs/org/apache/maven/model/Dependency.html) instead of creating custom ones, with strings

see
[Model](https://maven.apache.org/ref/3.5.2/maven-model/apidocs/org/apache/maven/model/Model.html)
[ModelBuildingRequest](https://maven.apache.org/ref/3.5.2/maven-model-builder/apidocs/org/apache/maven/model/building/ModelBuildingRequest.html)
[Maven Model Builder](https://maven.apache.org/ref/3.8.4/maven-model-builder/index.html)
etc.

[MavenXpp3Reader](https://maven.apache.org/ref/3.5.2/maven-model/apidocs/org/apache/maven/model/io/xpp3/MavenXpp3Reader.html) also allows reading an existing POM to extract some useful information.

could be nice to play with, will perhaps try on spare time after but will focus on features first ;)

I guess there are the same utilities for #408

pascalgrimaud:
Yes, we could use Maven to read in pom.xml but the problem is: it's a little bit slow :-D
When you generate a lof of files, calling Maven in each methods is not acceptable.
We could give a try but I'm afraid it won't work nice.

Instead, we could play with regexp:
- check if dependency is already here or not
- then, try to find needle and add dependency at this place
- if there is no needle, add it at the end of section:

```
  </dependency>
</dependencies>
```

pblanchardie:
Yes after writing this I also reminded that  simple text processing is much faster.
As long as our use cases stay simple we won't need such tooling.

pascalgrimaud:
Let's close this one as the current code works well

Issue #302, opened 2021-12-05
CI with Windows


pascalgrimaud:

pblanchardie:
`git config --global core.autocrlf true` to properly test #480

pascalgrimaud:
- closed by https://github.com/jhipster/jhipster-lite/pull/509

PR #450, opened 2021-12-31
Fix maven deletion tests

pblanchardie:
On Windows with LF separators (forced by Git), Maven property and dependency deletions were failing because the deletion code expects CRLF.
Instead of relying on items already present in the pom test file, these tests pass by adding and removing dummy dependencies.


pascalgrimaud:
That's why https://github.com/jhipster/jhipster-lite/issues/302 is really important. I'll work on this in the next days

pascalgrimaud (approved):

pblanchardie:
> That's why #302 is really important. I'll work on this in the next days

in this very specific case, the problem was only on Windows+LF but non on Windows+CRLF, so CI would need to test both cases (that's why I thought about #440).

PR #453, opened 2021-12-31
Fix integrations tests

pblanchardie:
- fix duplicate `@IntegrationTest` on both enclosing class and inner class (probably not the cause of the problem, but...)
- limit hikari max pool size during tests (testcontainer-postgresql was complaining about too much connections during testing)

### DIFF
--- a/src/main/resources/generator/server/springboot/dbmigration/liquibase/test/LiquibaseConfigurationIT.java.mustache
+++ b/src/main/resources/generator/server/springboot/dbmigration/liquibase/test/LiquibaseConfigurationIT.java.mustache
@@ -9,15 +9,14 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import {{packageName}}.IntegrationTest;
 
-@IntegrationTest
 class LiquibaseConfigurationIT {
-
-  @Autowired
-  ApplicationContext applicationContext;
 
   @Nested
   @IntegrationTest(properties = { "application.liquibase.async=true" })
   class Async {
+
+    @Autowired
+    ApplicationContext applicationContext;
 
     @Test
     void shouldGetLiquibaseAsync() {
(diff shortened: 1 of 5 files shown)


src/main/resources/generator/server/springboot/dbmigration/liquibase/test/LiquibaseConfigurationIT.java.mustache
```diff
@@ -9,16 +9,15 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import {{packageName}}.IntegrationTest;
 
-@IntegrationTest
```
pascalgrimaud, line 12: If possible, I'd like to keep this
As later, I'd like to add this test: https://github.com/jhipster/jhipster-lite/blob/main/src/test/java/tech/jhipster/lite/AnnotationArchTest.java#L37

I don't think the IntegrationTest is the cause of issue. but if it's the case, I'll change my AnnotationArchTest

src/main/resources/generator/server/springboot/logging/logstash/tcp/test/LogstashTcpConfigurationIT.java.mustache
```diff
@@ -19,7 +19,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import {{packageName}}.IntegrationTest;
 
-@IntegrationTest
```
pascalgrimaud, line 22: Same comment than above

src/main/resources/generator/server/springboot/mvc/security/jwt/test/CorsFilterConfigurationIT.java.mustache
```diff
@@ -9,7 +9,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.filter.CorsFilter;
 import {{packageName}}.IntegrationTest;
 
-@IntegrationTest
```
pascalgrimaud, line 12: Same comment than above

pascalgrimaud (changes requested):

pascalgrimaud (approved):

PR #454, opened 2021-12-31
Improve app startup log

pblanchardie:
Fix https://github.com/jhipster/jhipster-lite/issues/475

log the appropriate information on startup (instead of localhost:null)

### DIFF
--- a/src/main/resources/generator/server/springboot/core/MainApp.java.mustache
+++ b/src/main/resources/generator/server/springboot/core/MainApp.java.mustache
@@ -22,30 +22,44 @@ public class {{mainClass}}App {
   }
 
   private static void logApplicationStartup(Environment env) {
-    String protocol = getProtocol(env.getProperty("server.ssl.key-store"));
     String serverPort = env.getProperty("server.port");
-    String contextPath = getContextPath(env.getProperty("server.servlet.context-path"));
-    String hostAddress = getHostAddress();
-    log.info(
-      """
-      
-      ----------------------------------------------------------
-        Application '{}' is running! Access URLs:
-        Local: \t{}://localhost:{}{}
-        External: \t{}://{}:{}{}
-        Profile(s): \t{}
-      ----------------------------------------------------------
-      """,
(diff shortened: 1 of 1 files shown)


src/main/resources/generator/server/springboot/core/MainApp.java.mustache
```diff
@@ -22,30 +22,44 @@ public class {{mainClass}}App {
...
-      protocol,
-      serverPort,
-      contextPath,
-      protocol,
-      hostAddress,
-      serverPort,
-      contextPath,
-      env.getActiveProfiles()
-    );
+    if (serverPort != null) {
```
pascalgrimaud, line 26: This big 'if' is not really elegant :-)
I have an idea of how to improve it. I'll do a suggestion soon

pblanchardie: Yes, it was late and I was a bit lazy so I gave up^^

pascalgrimaud: I wonder if we should:
- extract each part into specific static method, and use StringBuilder ?

swarajsaaj: Instead of doing a long condition ladder, wondering if it would be better to allow each business context to optionally add a startup message of their own, through a shared context maybe (list of messages that can be logged here?). Maybe later, WDYT?

pascalgrimaud: Several solutions here:

Doing everything here, in main:
Pros:
- easy for the user to custom
- 1 single place
Cons:
- not elegant code
- some part won't never be executed (once there is a server app, the if serverPort will never be null)

Applying a patch or something else, to add info in main:
Pros:
- nice generated code
Cons:
- harder to code in the generator part

pascalgrimaud:
@pblanchardie : I submit https://github.com/jhipster/jhipster-lite/pull/482

Are you ok to close this one?

PR #455, opened 2022-01-01
Put back IntegrationTest annotation with Nested

pascalgrimaud:
Following https://github.com/jhipster/jhipster-lite/pull/453


pascalgrimaud:
as it seems to work, I'm merging this, @pblanchardie

pblanchardie:
@pascalgrimaud yes I should only have commited the pool size limit.

Issue #461, opened 2022-01-02
Package Manager NPM


pascalgrimaud:
Following this: https://github.com/jhipster/jhipster-lite/issues/446

A little bit similar to Maven (backend part), this context is for the frontend part

PR #462, opened 2022-01-02
Add package manager context

pascalgrimaud:
Related to https://github.com/jhipster/jhipster-lite/issues/461


pascalgrimaud:
cc @swarajsaaj : here my ideas ->
- refactoring to put the API in this context
- add the new methods : we can't have needle here, so we need to do some hard replace

Issue #477, opened 2022-01-04
System.lineSeparator is not relevant in all cases


pblanchardie:
- won't make sense on a remote server (aka JHipster Online)
- is not appropriate on some local workstations (Windows with LF project: adds CRLF in LF files)

suggestions:
- only support LF? what is the problem with that?
- auto-detect file line separators when applying modifications?
- switch to a project config with default to LF?

pascalgrimaud:
Great ticket ! Important to fix these issue
Instead of using lineSeparator or hard code \n, we could use our own constant + dedicated method to do some replace when needed

Then, I don't know which is the best solution for our Windows users.
What do you think @swarajsaaj ?

pascalgrimaud:
I think there is 2 issues here:
- the tests in generator should work on Windows and Linux
- the generated code could support only LF. I think it is the choice in generator-jhipster, as it's forced with git, but I'm not sure :-D

swarajsaaj:
It would be really nice to have a consistent convention for line endings in pproject.
Even Java text blocks use only '\n' for new lines, irrespective of OS (currently to get our test cases working on all platforms, we replace it manually with System.lineSeparator() ).

Need to think more on how will it behave for non jhipster-lite projects, as we currently are open to even non-jhipster-lite generated projects, their project configuration on Windows may not have same line endings.

pblanchardie:
Yes @swarajsaaj.

Moreover, when you programmatically add an import in java, it's \n so I'm not sure we should bother with that for some files and not for other.

If we want to handle non-jhipster projects, better abandon the needle and opt pour xml processing with maven library or any other IMHO.

pascalgrimaud:
Handle non jhipster project is an idea and could be done really later. I don't think we should put our effort on this now

swarajsaaj:
> Handle non jhipster project is an idea and could be done really later. I don't think we should put our effort on this now

Great, then lets go ahead with the LF convention. (also maybe add the same in .gitattributes for sanitizing commits)

pblanchardie:
Almost done, will provide PR.

pascalgrimaud:
It should already be the case ?
https://github.com/jhipster/jhipster-lite/blob/main/.gitattributes

pblanchardie:
> It should already be the case ?
> https://github.com/jhipster/jhipster-lite/blob/main/.gitattributes

It only forces ending for bat, sh... and leave the rest up to the system, which is not sufficient indeed.

[...]

PR #478, opened 2022-01-04
fix maven dependency management indentation

pblanchardie:
the same approach could be used on plugin management and others

indentation:
- `Maven` only provides natural indentation
- `MavenDomainService` adds a level of indentation if needed

line separators:
- `Maven` only works with `\n` (like `String.indent`)
- `MavenDomainService` converts line separators, see #477

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
@@ -3,6 +3,7 @@ package tech.jhipster.lite.generator.buildtool.maven.domain;
 import static tech.jhipster.lite.common.domain.WordUtils.indent;
 
 import java.util.List;
+import java.util.stream.Collectors;
 import tech.jhipster.lite.common.domain.WordUtils;
 import tech.jhipster.lite.generator.buildtool.generic.domain.Dependency;
 import tech.jhipster.lite.generator.buildtool.generic.domain.Parent;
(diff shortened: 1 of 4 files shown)


pblanchardie:
codecov not up to date

src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
```diff
@@ -97,102 +98,82 @@ public static String getParent(Parent parent, int indentation) {
...
+    String begin = DEPENDENCY_BEGIN + "\n";
+
+    String content = new StringBuilder()
       .append(GROUP_ID_BEGIN)
       .append(dependency.getGroupId())
       .append(GROUP_ID_END)
-      .append(System.lineSeparator())
-      .append(indent(3, indentation))
+      .append("\n")
       .append(ARTIFACT_ID_BEGIN)
```
pascalgrimaud, line 113: I'm a little bit lot with issue on Windows
Should we use lineSeparator or \n?

pascalgrimaud (approved):

src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
pblanchardie: At this level only \n.

Now it's the job of the DomainService to convert separators if needed (eg by the project).

PR #480, opened 2022-01-04
improve line separator handling

### LINKED ISSUE #477: System.lineSeparator is not relevant in all cases

pblanchardie:
fix #477 

- [x] .gitattributes
- [x] .editorconfig
- [x] .prettierrc
- [x] Add a lineSeparator project configuration and use it where applicable
- [x] Remove all uses of System.lineSeparator
- [ ] Add utility method that replaces LF with custom lineSeparator

Some use of indentation constants were also committed as part of this PR

@pascalgrimaud supporting custom line separators would also required that we ensure applying them when modifying Java files (eg. updateExceptionTranslatorIT).

As LF is widely used in git repos by both Unix and Windows worlds, I'm not sure it's worth the effort after all. Let's re-think about forcing LF for everyone as good-practice. I don't know any advantage of using CRLF in source code, and it should never be used in Git.

Note that .gitattributes still overrides bat, cmd, ps1 and sh but it's not the case in editorconfig and prettierrc.

### DIFF
--- a/src/main/java/tech/jhipster/lite/common/domain/FileUtils.java
+++ b/src/main/java/tech/jhipster/lite/common/domain/FileUtils.java
@@ -1,14 +1,13 @@
 package tech.jhipster.lite.common.domain;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.Optional;
 import java.util.Scanner;
 import java.util.UUID;
 import java.util.regex.Pattern;
@@ -136,5 +135,24 @@ public class FileUtils {
 
   public static boolean isPosix() {
     return FileSystems.getDefault().supportedFileAttributeViews().contains("posix");
+  }
+
+  public static Optional<String> detectEndOfLine(String filename) throws IOException {
+    try (BufferedReader reader = new BufferedReader(new InputStreamReader(new FileInputStream(filename), StandardCharsets.UTF_8))) {
+      char previousChar = 0;
+      char currentChar;
+      int read;
+      while ((read = reader.read()) != -1) {
+        currentChar = (char) read;
+        if (currentChar == '\n') {
+          if (previousChar == '\r') {
+            return Optional.of(WordUtils.CRLF);
+          }
+          return Optional.of(WordUtils.LF);
+        }
+        previousChar = currentChar;
+      }
+    }
+    return Optional.empty();
   }
 }
--- a/src/main/java/tech/jhipster/lite/common/domain/WordUtils.java
+++ b/src/main/java/tech/jhipster/lite/common/domain/WordUtils.java
@@ -5,6 +5,9 @@ import tech.jhipster.lite.error.domain.Assert;
 
 public class WordUtils {
 
+  public static final String LF = "\n";
+  public static final String CRLF = "\r\n";
+
   public static final int DEFAULT_INDENTATION = 2;
   public static final String DQ = "\"";
 
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
@@ -1,5 +1,6 @@
 package tech.jhipster.lite.generator.buildtool.maven.domain;
 
+import static tech.jhipster.lite.common.domain.WordUtils.LF;
 import static tech.jhipster.lite.common.domain.WordUtils.indent;
 
 import java.util.List;
@@ -67,17 +68,17 @@ public class Maven {
   public static String getParentHeader(Parent parent, int indentation) {
     StringBuilder result = new StringBuilder()
       .append(PARENT_BEGIN)
-      .append(System.lineSeparator())
+      .append(LF)
       .append(indent(2, indentation))
       .append(GROUP_ID_BEGIN)
       .append(parent.getGroupId())
       .append(GROUP_ID_END)
-      .append(System.lineSeparator())
+      .append(LF)
       .append(indent(2, indentation))
       .append(ARTIFACT_ID_BEGIN)
       .append(parent.getArtifactId())
       .append(ARTIFACT_ID_END)
-      .append(System.lineSeparator());
+      .append(LF);
     return result.toString();
   }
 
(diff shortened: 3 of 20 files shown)


src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
```diff
@@ -1,5 +1,6 @@
 package tech.jhipster.lite.generator.buildtool.maven.domain;
 
+import static org.apache.commons.lang3.StringUtils.LF;
```
pascalgrimaud, line 3: in domain, we should not depend on external lib, so I'd suggest to:
- add this constant in common/domain/WordUtils (or another domain class)
- use this constant in all our domain classes

Not need to use external lib for that :-) we can code it!
image

pblanchardie: Sure

pblanchardie:
It's now cleaner with `project.getEndOfLine()`

@pascalgrimaud Should we add endOfLine in ProjectDTO so the user can force it or leave it null for "automatic" mode?
Or only rely on some arbitrary file in the project to find eol so we don't ask dumb questions and find the answer ourselves?^^

For the moment, it's forced to LF with a TODO.

src/main/java/tech/jhipster/lite/generator/init/domain/InitDomainService.java
```diff
@@ -49,14 +50,34 @@ public void addReadme(Project project) {
...
     projectRepository.add(project, SOURCE, "gitignore", ".", ".gitignore");
-    projectRepository.add(project, SOURCE, "gitattributes", ".", ".gitattributes");
+
+    project.addConfig(
+      "gitEndOfLine",
+      switch (project.getEndOfLine()) {
+        case CRLF -> "crlf";
+        case CR -> "cr";
+        default -> "lf";
+      }
```
pblanchardie, line 63: or toLowerCase

src/main/java/tech/jhipster/lite/generator/init/domain/InitDomainService.java
```diff
@@ -49,14 +50,34 @@ public void addReadme(Project project) {
...
   public void addEditorConfiguration(Project project) {
     project.addDefaultConfig(PRETTIER_DEFAULT_INDENT);
 
+    project.addConfig(
+      "editorConfigEndOfLine",
+      switch (project.getEndOfLine().toLowerCase()) {
+        case CRLF -> "crlf";
+        case CR -> "cr";
+        default -> "lf";
+      }
```
pblanchardie, line 78: or toLowerCase

src/main/java/tech/jhipster/lite/generator/project/domain/Project.java
```diff
@@ -162,9 +169,15 @@ public void checkFolder() {
     }
   }
 
+  public String detectEndOfLine() {
+    // TODO read some arbitrary files to detect end of line?
```
pblanchardie, line 173: read an arbitrary file in the folder and rely on it?

src/main/java/tech/jhipster/lite/common/domain/WordUtils.java
```diff
@@ -5,6 +5,10 @@
 
 public class WordUtils {
 
+  public static final String LF = "\n";
+  public static final String CR = "\r";
```
pblanchardie, line 9: CR alone is a joke, remove it?

pascalgrimaud: Yes I think lol

src/main/java/tech/jhipster/lite/generator/project/domain/Project.java
pascalgrimaud: Yes it's a good idea

pblanchardie:
git patch fails with CRLF, maybe we have to force .patch files to be lf, or definitely force LF for everyone as we'll run into multiple issues with CRLF without bringing value.

hard to make a static method do throw an exception (powermockito?):
https://github.com/jhipster/jhipster-lite/pull/480/checks?check_run_id=4703426160

src/main/java/tech/jhipster/lite/common/domain/FileUtils.java
```diff
@@ -137,4 +136,24 @@ public static String replace(String text, String regexp, String replacement) {
...
+      char currentChar;
+      int read;
+      while ((read = reader.read()) != -1) {
+        currentChar = (char) read;
+        if (currentChar == '\n') {
+          if (previousChar == '\r') {
+            return Optional.of(WordUtils.CRLF);
+          } else {
+            return Optional.of(WordUtils.LF);
+          }
```
pascalgrimaud, line 152:
```suggestion
          }
          return Optional.of(WordUtils.LF);
```

pascalgrimaud:
I didn't expect you provide a 1st solution so fast!

> As LF is widely used in git repos by both Unix and Windows worlds, I'm not sure it's worth the effort after all. Let's re-think about forcing LF for everyone as good-practice. I don't know any advantage of using CRLF in source code, and it should never be used in Git.

I agree with this, maybe we should simply force LF for project

pblanchardie:
Don't worry, the most important is not lost ;)

The code is now cleaner and platform independant, all tests pass and we can reuse detection to ensure LF: it will be better than failing on the git patch, as the error it produces is very hard to understand and can be very frustrating for the user.

We really needed to investigate on this windows related stuff anyway.

Changing gitattributes is not sufficient, as core.autocrlf is responsible for converting the working copy to crlf and is a global config. We could also check it to be false when runni g on local machine...

pascalgrimaud:
@pblanchardie : similar issue here https://github.com/jhipster/generator-jhipster/issues/17115

pascalgrimaud:
@pblanchardie : can I merge this?

pblanchardie:
You can, but it may need further improvements.

pascalgrimaud:
Ok, so I'll wait and merge when the PR will be ready :-)

pblanchardie:
fixing conflicts!

pblanchardie:
git patch now works everywhere when *.patch files are forced to stay with lf

pblanchardie:
> @pblanchardie : similar issue here [jhipster/generator-jhipster#17115](https://github.com/jhipster/generator-jhipster/issues/17115)

I did read and was about to answer as I have some advice about what not to do, but I'll get back to it tomorrow ;)

this PR can be merged as-is, more to come in other PR.

pascalgrimaud:
Ok @pblanchardie : I'll merge it tomorrow, as I think I need to use your work in existing code:
- I merge some PR, 1 hour ago and I used hard code \n I think

pblanchardie:
Ok @pascalgrimaud 

Detection should be improved to ignore sh, vsproj, bat... that are not relevant because they are forced to lf/crlf, and binary files. Not big deal, planned.
Good night!

pascalgrimaud (approved):

pascalgrimaud:
Huge work, thanks a lot @pblanchardie

PR #488, opened 2022-01-05
Init prettier xml

pascalgrimaud:


pblanchardie:
@pascalgrimaud when using "Reformat with prettier", it properly goes to `<relativePath />`.
But when using IntelliJ "Reformat Code" it reverts to `<relativePath/>`, even if I "Apply prettier code style rules" by right-clicking prettierrc.
So I guess we must only use "Reformat with prettier" instead of "Reformat Code"?

pascalgrimaud:
Oh I see. The format code by IntelliJ is always different, for Java (import), for Xml.

So a good practice is to avoid using reformat code with IntelliJ (or use the prettier plugin), you can use it, but it will be changed when you launch git add & git commit

pblanchardie:
Thanks!

PR #491, opened 2022-01-06
harmonization of pom.xml indentation and eol management

pblanchardie:
handle maven parent & plugin handling the same way as dependencies

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
@@ -1,7 +1,6 @@
 package tech.jhipster.lite.generator.buildtool.maven.domain;
 
 import static tech.jhipster.lite.common.domain.WordUtils.LF;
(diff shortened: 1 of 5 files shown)


src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/Maven.java
```diff
@@ -66,37 +65,36 @@ public static String getParentHeader(Parent parent) {
...
-      .append(indent(2, indentation))
+    String header = getParentHeader(parent, indentation);
+
+    String additionalContent = new StringBuilder()
       .append(VERSION_BEGIN)
       .append(parent.getVersion())
       .append(VERSION_END)
       .append(LF)
-      .append(indent(2, indentation))
       .append("<relativePath />")
```
pblanchardie, line 93: @pascalgrimaud I've seen that you added a space before `/>`  but my IDE wants to revert to <relativePath/> so I guess we have a mismatch in editorconfig for that, as prettier does not seem to be the responsible?

PR #496, opened 2022-01-06
Add springboot user

### LINKED ISSUE #426: Spring Boot PostgreSQL: User / Authority Entities
Can be done after https://github.com/jhipster/jhipster-lite/issues/425

- create UserEntity and AuthorityEntity
- create UserJpaRepository, AuthorityJpaRepository
- add new xml entry in Liquibase: need to think about flyway, once the support is added, but for now, let's focus on liquibase

MSaguer:
To close #426 

- [x] First working instance after user generation
- [x] Refactor of generated : temporary generated package
- [x] Refactor of generation code : package modification ? Separate the generation in many end points?
- [x] Add tests in generator
- [x] Add tests in generated project


MSaguer:
Hi @pascalgrimaud 
Here is a first shot of generated files in tech.jhipster.beer.technical.infrastructure.secondary.user package
image

Is it the right target package or do it need to be moved?

Same question on generation files. I've created source files in new tech.jhipster.lite.generator.server.springboot.user package.

Last question on required end points. For the moment, I have only one end point /api/servers/spring-boot/user/init that creates both user and authority files. I could split it in /api/servers/spring-boot/user/init and /api/servers/spring-boot/user/authority but they are not fully independant. The second one would require the first one to work.

Nota : I've seen that an Arch test is failling. I'll fix it when the target package will be defined.

pascalgrimaud:
The target should be to have a specific user context, so a dedicated folder `user`, so it could be:

image

Another solution would be to add secondary/postgresql, in case of you have different database in your architecture:

image

pascalgrimaud:
For this PR, you can ignore primary, application and domain, as it will be done in another PR

MSaguer:
The second option seems good for me.

when reading [User.java.ejs](https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/java/package/domain/User.java.ejs), I can see that user class is the same for all slq database (using databaseTypeSql variable).
What do you think to replace postgresql package by sqldatabase or something like that?

pascalgrimaud:
So maybe just 'sql' ?

MSaguer:
Forget my last remark. It takes me a long time to fully understand what you said. I agree now it could be usefull to be able to generate user entities for many different databases.
So I change my end point into "/api/servers/spring-boot/user/postgresql" that calls the following services 

```
    springBootUserApplicationService.addSqlJavaUsers(project, sqlDatabaseName);
    springBootUserApplicationService.addSqlJavaAuthority(project, sqlDatabaseName);
    springBootUserApplicationService.addSqlJavaAuditEntity(project, sqlDatabaseName);
    springBootUserApplicationService.addSqlLiquibaseConfiguration(project, sqlDatabaseName);
```

Should be simplier to add more service for different databases.

pascalgrimaud:
Mhhh let me think, but I don't think liquibase should be part of this PR.
Imagine the user don't use liquibase but flyway, or nothing. So we should not include the liquibase part, in this PR.

IMO, creating user.xml and authority.xml can be done in another PR and should be in liquibase context, not in this one.

What do you think?

Later, at the end of this workflow, we could publish a spring event, and a database migration context can consume this event, and create liquibase diff or flyway diff, according to the project config. It's just an idea so I'm not sure about this

MSaguer:
user.xml and autority.xml are bounded to both user and liquibase services. That's why I planned to have a separated end point to let user decides or not if this should be generated.
I don't mind hosting this generation in liquibase or in user package.

I understand that liquibase:diff cmd could be the future target stragegy for changelog generation? If so, Is it still usefull to keep the xml generation I created ? 
Do you want me to move this to the liquibase package and to propose it in an other PR?

pascalgrimaud:
No, I don't think liquibase:diff will be used by JHipster Lite to generate changelog. It depends too much on external services (database, base code etc), so it's not a good idea. The end-user will use it if he/she wants.

About this PR.
You can keep user.xml and authority.xml in this pull request, no need to do it in another PR if it's simpler for you.

I wanted to point it because of this change:
image

Liquibase should stay as BusinessContext, it's not a shared kernel.
It means:
- user.xml / authority.xml should be in liquibase context
- methods which add these 2 files should be in liquibase context too
- user context should not call liquibase

Don't hesitate if I'm not clear enough. My english is really good.
Or send private message on Twitter ;)

pascalgrimaud:
Just a comment, according to your last commits.
In my opinion, it's important to have `Entity` suffix, like UserEntity or AuthorityEntity, used in secondary -> your commits are good :-)

So User and Authority can be used in domain, which more important

That's one of reason why, when using generator-jhipster, I try to add Entity suffix.

pascalgrimaud:
Thanks a lot @MSaguer : it's a huge work but it is needed. I'll review it soon !

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/liquibase/domain/LiquibaseDomainService.java
```diff
@@ -55,7 +56,9 @@ public void addLiquibase(Project project) {
 
   @Override
   public void addChangelogMasterXml(Project project) {
-    projectRepository.add(project, getPath(SOURCE, "resources"), MASTER_XML, getPath(MAIN_RESOURCES, CONFIG_LIQUIBASE));
+    projectRepository.add(project, getPath(SOURCE, "resources"), LIQUIBASE_MASTER_XML, getPath(MAIN_RESOURCES, CONFIG_LIQUIBASE));
+    projectRepository.add(project, getPath(SOURCE, "resources"), "sequence.xml", getPath(MAIN_RESOURCES, CHANGELOG));
+    projectRepository.add(project, getPath(SOURCE, "resources"), "date_time_wrapper.xml", getPath(MAIN_RESOURCES, CHANGELOG));
```
pascalgrimaud, line 61: not sure it should be added by default:

- sequence -> not needed by MySQL https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs#L28-L32
- date_time_wrapper -> https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs#L167-L180

This was added for https://github.com/jhipster/jhipster-sample-app/tree/5a3ca402497db3b4e9b88f4186b4183e364d02ba/src/test/java/io/github/jhipster/sample/repository/timezone but I'm not sure it is needed by default when adding liquibase

src/main/resources/generator/server/springboot/user/sqldatabase/AuthorityRepository.java.mustache
```diff
@@ -0,0 +1,8 @@
+package {{packageName}}.user.infrastructure.secondary.{{sqlDatabaseName}};
+
+import org.springframework.data.jpa.repository.JpaRepository;
+
+/**
+ * Spring Data JPA repository for the {@link Authority} entity.
```
pascalgrimaud, line 6:
```suggestion
 * Spring Data JPA repository for the {@link AuthorityEntity} entity.
```

pascalgrimaud (changes requested):

pascalgrimaud:
Very good job @MSaguer ! I tested and it seems to work very well.
I have a few comments, specially about date_time_wrapper -> opened to discuss about it :)

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/liquibase/application/LiquibaseApplicationService.java
```diff
@@ -36,4 +36,12 @@ void addConfigurationJava(Project project) {
   void addLoggerInConfiguration(Project project) {
     liquibaseService.addLoggerInConfiguration(project);
   }
+
+  public void addSqlUserChangelog(Project project, String sqlDabaseName) {
```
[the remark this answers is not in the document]
MSaguer: Global modification to user DatabaseType instead of string

src/main/java/tech/jhipster/lite/generator/server/springboot/user/domain/SpringBootUserDomainService.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public SpringBootUserDomainService(ProjectRepository projectRepository) {
+    this.projectRepository = projectRepository;
+  }
+
+  @Override
+  public void addSqlJavaUser(Project project, String sqlDatabaseName) {
+    project.addDefaultConfig(PACKAGE_NAME);
+    String packageNamePath = project.getPackageNamePath().orElse(getPath("com/mycompany/myapp"));
+    project.addConfig(USER_DATABASE_KEY, sqlDatabaseName);
```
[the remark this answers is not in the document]
MSaguer: Use id field instead of name that is already defined in Enumerate class

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/liquibase/domain/LiquibaseDomainService.java
MSaguer: Remove these files

src/main/resources/generator/server/springboot/user/sqldatabase/UserConstants.java.mustache
```diff
@@ -0,0 +1,12 @@
+package {{packageName}}.user.infrastructure.secondary.{{sqlDatabaseName}};
```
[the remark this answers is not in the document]
MSaguer: New package generated

src/main/resources/generator/server/springboot/user/sqldatabase/UserEntity.java.mustache
```diff
@@ -0,0 +1,243 @@
...
+import org.hibernate.annotations.BatchSize;
+import org.hibernate.annotations.Cache;
+import org.hibernate.annotations.CacheConcurrencyStrategy;
+
+/**
+ * A user.
+ */
+@Entity
+@Table(name = "jhi_user")
+@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
```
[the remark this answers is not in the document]
MSaguer: Remove all cache options

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/liquibase/domain/LiquibaseDomainService.java
```diff
@@ -57,8 +57,6 @@ public void addLiquibase(Project project) {
   @Override
   public void addChangelogMasterXml(Project project) {
     projectRepository.add(project, getPath(SOURCE, "resources"), LIQUIBASE_MASTER_XML, getPath(MAIN_RESOURCES, CONFIG_LIQUIBASE));
-    projectRepository.add(project, getPath(SOURCE, "resources"), "sequence.xml", getPath(MAIN_RESOURCES, CHANGELOG));
```
pascalgrimaud: I was not sure about my last comment or maybe I was not clear enough.

When using liquibase + init, we don't need a sequence.
But when adding User with PostgreSQL, we need a sequence, as the UserEntity needs a sequence.
So maybe, we should add a sequence in User.xml and use a specific name to this sequence (instead of a global one): sequence_jhi_user for example.

What do you think?

MSaguer: You're right. My cut was a little rough yesterday :)

I see two solutions here
- Add new service LiquibaseApplicationService.addSequenceGenerator(Project, DatabaseType) that could be called in user/postgresql. This would create a global sequence that could be used by all generated entities.
That's what is done in current jhipster generator
- Add the sequence definition in the user.xml file that would be specific to the user entity. Then each entity should define its own sequence.
I think I remember that jhipster generator firstly used a dedicated sequence for each generated entity.

I personnaly prefer the second solution (the one you proposed) just because data are strictly separated. One entity = one sequence.
The benefit of the first option would be to avoid mistakes when switching by error two different table fk.
Just for my informatin, was it the rationale for the generator modification or is there anything else?

to resume the fix, just to be sure :)
- date_time_wrapper is not needed and should stayed deleted
- specific sequence generator is declared in user.xml file

pascalgrimaud: Agree with your resume! Let's try separate sequence, linked to each entity

pascalgrimaud:
Thanks, will do the final review soon :-)

pascalgrimaud (approved):

pascalgrimaud:
Very good job @MSaguer 
There are still have some minor fix, but let's merge it first and do some minor PRs to fix it

MSaguer:
thanks @pascalgrimaud for having fixed the latest problems and for the bounty.
I've created the request https://opencollective.com/generator-jhipster/expenses/61903

pascalgrimaud:
@MSaguer : approved. Thanks for your hard work :)

Issue #497, opened 2022-01-06
Typescript: init


pascalgrimaud:
- dependency in package.json
- .eslintrc.js
- jest.config.js
- tsconfig.json

Maybe it can be used just before init Vue

Inspired by https://gitlab.ippon.fr/twitch/live-coding-fr/-/tree/master/diamond-ts, coded by @nonomoho

PR #498, opened 2022-01-06
Typescript: add dependency, script and config files

pascalgrimaud:
Fix https://github.com/jhipster/jhipster-lite/issues/497

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/typescript/domain/Typescript.java
+++ b/src/main/java/tech/jhipster/lite/generator/typescript/domain/Typescript.java
@@ -7,7 +7,7 @@ public class Typescript {
 
   private Typescript() {}
 
-  public static Map<String, String> dependencies() {
+  public static Map<String, String> devDependencies() {
     // @formatter:off
     return Map.of(
       "@types/jest", "27.4.0",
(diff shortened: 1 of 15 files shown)


src/main/java/tech/jhipster/lite/generator/typescript/domain/Typescript.java
```diff
@@ -0,0 +1,48 @@
+package tech.jhipster.lite.generator.typescript.domain;
+
+import java.util.List;
+import java.util.Map;
+
+public class Typescript {
+
+  private Typescript() {}
+
+  public static Map<String, String> dependencies() {
```
nonomoho, line 10: In the package.json, you can provide two kind of dependencies (dependencies and devDependencies).  
The difference between these two, is that devDependencies are modules which are only required during development, while dependencies are modules which are also required at runtime.

I think that all your dependencies here are, in fact, devDependencies.

pascalgrimaud: Indeed, having it in devDependencies works fine too:

image

image

nonomoho (approved):

Issue #424, opened 2021-12-28
Build Docker Image


pascalgrimaud:
I don't know what is the best solution :
- use JIB
- add a Dockerfile at root folder (which will contain the build part too)

Maybe we can provide 2 options

swarajsaaj:
I think atleast `JIB` must be there as its the simpler+efficient way to build java container images.
Will do a PR for jib soon.

For Dockerfile, I think it would not serve anything extra over Jib, unless the project needs some custom commands/build process etc, which ofcourse will require changes by user later on

pascalgrimaud:
@swarajsaaj : I would say `springboot/docker/jib` as we could add other stuff later under `springboot/docker/xxxx` :-)

swarajsaaj:
Thanks @pascalgrimaud , yes the url would be 'springboot/docker/jib' but all this under a new common docker business context in folder 'springboot/docker' or you suggest a dedicated context for jib too?

pascalgrimaud:
I would suggest these folders, but I'm not totally sure, so don't hesitate to suggest if you have a better idea :-)

![image](https://user-images.githubusercontent.com/9156882/148655252-bc345554-386a-4f41-920b-a88e4fff6fbe.png)

pascalgrimaud:
@swarajsaaj : don't forget your bounty here :-)

swarajsaaj:
Thanks @pascalgrimaud for reminder :)
Submitted at https://opencollective.com/generator-jhipster/expenses/63108

pascalgrimaud:
@swarajsaaj : in Open Collective, the title refers to another pull request. Can you update it plz ?

swarajsaaj:
Oops... fixed.

pascalgrimaud:
@swarajsaaj : approved :-)

PR #539, opened 2022-01-16
Maven: Add version suffix manually

swarajsaaj:
- Remove .version suffix in all properties added by addProperty() method (Reflects the work done by method and allows adding properties more flexibly)
- There are other properties that needs to be added in pom without (".version suffix) (https://github.com/jhipster/jhipster-sample-app/blob/main/pom.xml)

Found specifically while adding `jib` properties `jib-maven-plugin.image` and `jib-maven-plugin.architecture` (#424)

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/BuildToolDomainService.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/generic/domain/BuildToolDomainService.java
@@ -72,9 +72,9 @@ public class BuildToolDomainService implements BuildToolService {
   }
 
   @Override
-  public void addProperty(Project project, String key, String version) {
+  public void addProperty(Project project, String key, String value) {
     if (project.isMavenProject()) {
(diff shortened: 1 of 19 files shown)


pascalgrimaud:
I think it's a good idea, but I would do it differently.

The new addProperty should add the value like this
The old addProperty should be renamed to addPropertyVersion, so no need to think about adding `.version`
Another method could be: addPropertyImage, which will suffix with image

What do you think?

swarajsaaj:
Hi @pascalgrimaud ,
I started out in a similar way, but did not want to duplicate too much code here, because there are `removeProperty`, `getProperty` as well for each `addProperty` (and then `addVersionProperty`,`removeVersionProperty`,`getVersionProperty`).

I can either just add `addVersionProperty` and `addImageProperty` in `BuildToolService` which further call `addProperty`. WDYT?

pascalgrimaud:
Ok fine, let's keep like this so :)
Thanks

PR #538, opened 2022-01-16
Add SpringDoc

Bolo89:
Hello,
- Add spring doc : fix https://github.com/jhipster/jhipster-lite/issues/219

Example : 

```json
{
  "folder": "/home/bolo/Projects/my-project-test",
  "generator-jhipster": {
    "projectName": "Beer Project",
    "baseName": "beer",
    "prettierDefaultIndent": 2,
    "packageName": "tech.jhipster.beer"
    "apiTitle": "My API",
    "apiDescription": "My API description",
    "apiLicenseName": "Custom 5.0",
    "apiLicenseUrl": "http://license-url.fr",
    "apiExternalDocDescription": "External API doc description",
    "apiExternalDocUrl": "http://external-api-doc.fr"
  }
}
```

PR #557, opened 2022-01-19
Rename spring-doc to springdoc

pascalgrimaud:
- Following https://github.com/jhipster/jhipster-lite/pull/538


pascalgrimaud:
cc @Bolo89 : hope you're ok with this change

Issue #234, opened 2021-11-26
Spring Boot Database Migration: flyway support


pascalgrimaud:

Bolo89:
Hello,

I can develop this feature. The steps could be :  

Add the dependencies in the pom.xml

Add in the application.properties file the values : 
- spring.flyway.enabled=true
- spring.flyway.driver-class-name=
- spring.flyway.url=
- spring.flyway.user=
- spring.flyway.password=

And add in the Project domain (in the config map, without default values)
- flywayDriverClassName=
- flywayUrl=
- flywayUser=
- flywayPassword=

It's ok ? or missing somethings ? :)

pascalgrimaud:
Thanks a lot @Bolo89 !

About default values, I'm not sure it is necessary, as after generation, the user should edit, if needed.
It's better than providing all real values (user/password) in the API

Then, another question:
- spring.flyway.url, spring.flyway.user, spring.flyway.password depend on the database, right ?
- which value will you put by default?
- maybe, one suggestion would be to use the sames than spring.datasource.url, spring.datasource.user, etc
- it could be something like: spring.flyway.url=${spring.datasource.url}

What do you think?

Bolo89:
I will use directly the properties of the datasource. Thanks.

PR #570, opened 2022-01-22
Add Flyway

Bolo89:
- fix https://github.com/jhipster/jhipster-lite/issues/234


src/main/java/tech/jhipster/lite/technical/infrastructure/primary/global/ClockConfiguration.java
```diff
@@ -0,0 +1,14 @@
+package tech.jhipster.lite.technical.infrastructure.primary.global;
```
pascalgrimaud, line 1:
```suggestion
package tech.jhipster.lite.technical.infrastructure.config;
```

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/infrastructure/primary/rest/FlywayResource.java
```diff
@@ -0,0 +1,32 @@
...
+
+  private final FlywayApplicationService flywayApplicationService;
+
+  public FlywayResource(FlywayApplicationService flywayApplicationService) {
+    this.flywayApplicationService = flywayApplicationService;
+  }
+
+  @Operation(summary = "Add Flyway")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding Flyway")
+  @PostMapping
```
pascalgrimaud, line 27: It's because we'll have another API in this class to add user and authority sql files

```suggestion
  @PostMapping("/init")
```

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/domain/Flyway.java
```diff
@@ -0,0 +1,24 @@
+package tech.jhipster.lite.generator.server.springboot.dbmigration.flyway.domain;
+
+import tech.jhipster.lite.generator.buildtool.generic.domain.Dependency;
+
+public class Flyway {
+
+  public static final Boolean DEFAULT_FLYWAY_ENABLED = true;
+  public static final String DEFAULT_SQL_FILES_FOLDER = "db/migration";
```
pascalgrimaud, line 8: Liquibase generates changelog in `src/main/resources/config/liquibase`

I know `db/migration` is the default folder of flyway. I wonder if we should let like this, or rename to have consistency with liquibase, maybe add `flyway`in the path ?

Some solutions:
- let like this (default config)
- rename to be like liquibase: `src/main/resources/config/flyway`
- rename Flyway to: `src/main/resources/flyway` and rename Liquibase to `src/main/resources/liquibase`

Which one do you prefer ? Do you have another suggestion ?

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/domain/FlywayDomainService.java
```diff
@@ -0,0 +1,83 @@
...
+  public FlywayDomainService(
+    BuildToolService buildToolService,
+    ProjectRepository projectRepository,
+    SpringBootCommonService springBootCommonService,
+    Clock clock
+  ) {
+    this.buildToolService = buildToolService;
+    this.projectRepository = projectRepository;
+    this.springBootCommonService = springBootCommonService;
+    this.clock = clock;
```
pascalgrimaud, line 37: we'll need this later, when adding new changelog (for user and authority) so need to keep this

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/domain/FlywayDomainService.java
```diff
@@ -0,0 +1,83 @@
...
+    buildToolService.addProperty(project, "flyway.version", Flyway.flywayVersion());
+    buildToolService.addDependency(project, Flyway.flywayDependency());
+  }
+
+  @Override
+  public void addChangelogSql(Project project) {
+    projectRepository.add(
+      project,
+      getPath(SQL_INIT_FILE_SOURCE, "resources"),
+      "V${datetime}__init.sql",
```
pascalgrimaud, line 58: to be simpler, I think we can named it to: `V00000000000000__init.sql` (similar to https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/resources/config/liquibase/changelog/00000000000000_initial_schema.xml)

src/main/resources/generator/server/springboot/dbmigration/flyway/resources/V${datetime}__init.sql
```diff
@@ -0,0 +1,2 @@
+-- Delete the init change log in the table 'flyway_schema_history' if this file is modified;
```
pascalgrimaud, line 1: maybe rename this file to be simpler ? V00000000000000__init.sql

pascalgrimaud (changes requested):

pascalgrimaud:
Another comment: can you add the API in CI tests plz ?
We could replace liquibase here: https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L55

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/domain/Flyway.java
Bolo89: As discussed, I prefer to keep the default flyway path (liquibase and flyway don't have link each other)

src/main/java/tech/jhipster/lite/generator/server/springboot/dbmigration/flyway/infrastructure/primary/rest/FlywayResource.java
Bolo89: But the verb " POST " means that it's a creation ? So I think it's not useful, and we can use the verb " PUT " for update (add user and authority in this example) ?
I will do the change and use the "/init" to be "iso" with the existing.

Bolo89:
Hello,

Bounty claimed at : https://opencollective.com/generator-jhipster/expenses/63105

pascalgrimaud:
@Bolo89 : approved

PR #596, opened 2022-01-25
Spring-boot: use get version, managed by dependabot

DanielFran:


DanielFran:
Stuck with current approach for the tests as buildTool is empty.

image

@pascalgrimaud using project created from a pom.xml should define buildTool as maven service?

pascalgrimaud:
All these fields are currently not used, it's planned but not done yet today.
It will be needed once we want to orchestrate all different calls to generate the project in one single shot

pascalgrimaud:
If you need, I can have a look at your branch and take care of failed tests.
just tell me :)

DanielFran:
Feel free to add changes, it will speed up since I have few time...

PR #608, opened 2022-01-28
FileUtils: fix readline in classpath

pascalgrimaud:
It's probably the reason why https://github.com/jhipster/jhipster-lite/pull/596 failed

Once it's merged, can you rebase with this fix plz?


DanielFran:
@pascalgrimaud I was a little blocked because buildtool was not identifying process as a maven project... but did not have time to check the origin of this issue.
Will confirm if that was the issue

Issue #603, opened 2022-01-27
Split package.json related to dependencies


pascalgrimaud:
In this folder  https://github.com/jhipster/jhipster-lite/tree/main/src/main/resources/generator/dependencies, there is a package.json which is auto updated.

This file should be splitted into different files, related to each front which will be support

The idea would be:
- common
- angular
- react
- vue
- svelte (one day :-)

PR #610, opened 2022-01-28
Refactoring and move package.json to dependencies/common

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/603

it will impact your current branch for Angular, @qmonmert 
because you'l need to:
- create a specific package.json in dependencies/angular, which will contain the dependencies you need there
- add this file to dependabot, so it will be updated separately
- add a new method getVersionInAngular, in NpmService, and use this method instead

If it's not clear enough, ping me


qmonmert:
@pascalgrimaud done

Issue #619, opened 2022-01-29
UserJpaRepository: add missing integration test


pascalgrimaud:
Add Integration Tests for [UserJpaRepository](https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/springboot/user/sqldatabase/UserJpaRepository.java.mustache)

- clean a little bit the [unused variables](https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/springboot/user/sqldatabase/UserJpaRepository.java.mustache#L17-L19)
- add tests for each methods

beatfreaker:
@pascalgrimaud `UserJpaRepository` is added by `/postgres` or `/mysql` resource/api,
now to write integration test we need some datasource, for test we can use `h2databse`
but we are adding its dependency when `/liquibase` or `/flyway` api is called
so while writing `UserJpaRepositoryIT` can we assume that `h2databse` dependency will be available ? means user will first invoke `/liquibase` and then `/mysql` ?

pascalgrimaud:
We can assume there is already liquibase (or flyway), to create the table User
So you don't need to worry about that.
To be more precise, h2 is added with liquibase to test sync/async, not for real integration tests which are with PostgreSQL or MySQL + testcontainers instead

beatfreaker:
okay, so for this integration test we want to use testcontainer + MySQL docker container, or we can use in-memory db i.e. h2 ?

pascalgrimaud:
Sorry, my ticket is not clear enough :-)

The context for Integration Tests is already here, and well configured, it means you don't need to worry about that.

What is missing here is simply `UserJpaRepositoryIT.java`
It should:
- be annotated with `@IntegrationTest`
- test each requests in UserJpaRepository

Don't hesitate to tell me, if I'm not clear enough and I can show you some examples

beatfreaker:
@pascalgrimaud I think there is some issue whith how `liquibase` scripts are generated.

After generating the project if I try to save `UserEntity` I am getting this error
```
2022-02-02 19:57:06.663  WARN 25252 --- [           main] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 1364, SQLState: HY000
2022-02-02 19:57:06.664 ERROR 25252 --- [           main] o.h.engine.jdbc.spi.SqlExceptionHelper   : Field 'id' doesn't have a default value
```
[...]

PR #637, opened 2022-02-02
Fix generator sequence for User + PostgreSQL

pascalgrimaud:
- Related to https://github.com/jhipster/jhipster-lite/issues/619


pascalgrimaud:
cc @beatfreaker

Issue #591, opened 2022-01-25
Maven frontend-maven-plugin


pascalgrimaud:
Extract the frontend-maven-plugin to do it in a separated PR
- see https://github.com/jhipster/jhipster-lite/pull/559

pascalgrimaud:
Not sure yet about what to add here
- only npm install + npm build ?
- including tests too ?
- etc.

PR #653, opened 2022-02-07
Add Frontend Maven Plugin

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/591


pascalgrimaud:
cc @qmonmert : as it will be usefull in your branch

PR #724, opened 2022-02-15
Add axios http in infrastructure layer

### LINKED ISSUE #715: JHipster lite front: add axios
In order to make request to the backend. I propose to use axios. 

We need to create a new infrastructure "service" on the front for that.

nonomoho:
* close #715


src/main/webapp/app/infrastructure/AxiosHttp.ts
```diff
@@ -0,0 +1,21 @@
+import { AxiosInstance, AxiosResponse } from 'axios';
```
pascalgrimaud, line 1: Not sure about the folder, but maybe I'm wrong. Several suggestions from my side:
- `app/common/infrastructure`
- `app/technical/infrastructure`

I don't know if we'll have some common primary, but in case of, maybe add secondary?
- `app/common/infrastructure/secondary`
- `app/technical/infrastructure/secondary`

nonomoho: Yes, I think you're right. I never know how to organise these infrastructure files...  
Do you have a link or something to find best practices on it?

pascalgrimaud: It was a suggestion from the code I saw from @Gnuk: maybe you have an idea about this ?

Gnuk: Yes, it's in the secondary part.

Maybe we can discuss about things you want to have as tools but with common usages. Because it may not have sense to have an hexa architecture for stuff only technical or used as tools.

So, for example :

`/{app}/{bounded-context}/…`

* {app} => the app name
* {bounded-context} => the context

Hexa case:

* `/{app}/{bounded-context}/infrastructure/primary/…`
* `/{app}/{bounded-context}/infrastructure/secondary/…`
* `/{app}/{bounded-context}/domain/…`

Tool case possibility (used as a shared kernel, possible to extract as a library if the team want):

`/{app}/http/AxiosHttp.ts`

But the idea of this abstraction (for this specific case) is to inject things like the token for example.

nonomoho: So I use: `/{app}/http/AxiosHttp.ts`

pascalgrimaud: thanks for sharing your idea @Gnuk 
So +1 @nonomoho

src/test/javascript/spec/http/AxiosHttp.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { AxiosResponse } from 'axios';
+import { AxiosHttp } from '../../../../main/webapp/app/http/AxiosHttp';
```
pascalgrimaud, line 2: we really need to find a way to use `@` instead, but can be done in another PR

nonomoho: Yes, one of the next thing to do!

PR #749, opened 2022-02-16
Add styling in index html of vue generated project

### LINKED ISSUE #721: Vite/Vue: provide home page with minimal css
The Vite/Vue context should contain 2 apis:
- one with empty home page, without CSS (what it is currently)
- another one with minimal css (similar to what is generated by vite+vue)

The 1st one can be used for starting real project
The 2nd one is used for demo, as it's better to have minimal CSS than a simple text

matthieulapatate:
fix #721


src/main/java/tech/jhipster/lite/generator/client/vite/vue/core/infrastructure/primary/rest/ViveVueResource.java
```diff
@@ -31,4 +31,13 @@ public void addViteVue(@RequestBody ProjectDTO projectDTO) {
     Project project = ProjectDTO.toProject(projectDTO);
     viteVueApplicationService.addViteVue(project);
   }
+
+  @Operation(summary = "Add some style to index page")
```
seraphinandrieux, line 35: You should explain than you also init Vue + vite?

Here it suggests you already had init the project

src/test/java/tech/jhipster/lite/generator/client/vite/vue/core/domain/ViteVueDomainServiceTest.java
```diff
@@ -40,6 +40,14 @@ void shouldAddViteVue() {
     assertThatCode(() -> viteVueDomainService.addViteVue(project)).doesNotThrowAnyException();
   }
 
+  @Test
+  void shouldAddStyledViteVue() {
```
seraphinandrieux, line 44: This test is not really finished.

Here you check no exception, it could happens when the method do just nothing.
You should verify the correct port are called with arguments you specified in your method

src/test/java/tech/jhipster/lite/generator/client/vite/vue/core/infrastructure/primary/rest/ViveVueResourceIT.java
```diff
@@ -44,4 +44,15 @@ void shouldAddViteVue() throws Exception {
     assertRootFiles(project);
     assertAppFiles(project);
   }
+
+  @Test
+  void shouldAddStyledViteVue() throws Exception {
```
seraphinandrieux, line 49: Same here, especially an integration test, in this test you have to check if the specif files related to styled vue are generated.

So first step to check if the status is OK is done perfect, now go further and assert the files are generated with correct content :)

seraphinandrieux (changes requested):

pascalgrimaud:
I'm closing this, in favor of https://github.com/jhipster/jhipster-lite/pull/762
I took your commits in my other PR

pascalgrimaud:
The problem is because of `main` branch, so I couldn't commit into your branch to improve it

PR #766, opened 2022-02-19
[Angular] Improve style (same as Vue)

qmonmert:
image


pascalgrimaud:
@qmonmert : I'd like to have a new red color instead. Do you know how to do it at https://github.com/jhipster/jhipster-artwork/tree/main/logos/lite and use this new `JHipster-Lite-neon-red.png` here ?
If not, I can ask the author to do it :-)

pascalgrimaud:
Then, I think we can keep the original Angular CSS, as it's nicer than the Vue one.
We could only add the JHipster Lite logo + display the {{baseName}} instead. What do you think?

pascalgrimaud:
See the future React one, which is different from Vue:

image

qmonmert:
@pascalgrimaud 

first question: no
second question: I thought that we wanted same design for all fronts :/

pascalgrimaud:
Ok for the logo, I'll ask the author to have a red one.

About the front, I'm still thinking about it.
What I have in mind, is for each client front:
- Vue
  - init without CSS (already done)
  - init with minimal CSS which looks like to the one used in Vite+Vue (npm init vite@latest viteapp)
- Angular
  - init without CSS (to do)
  - init with minimal CSS, with a similar one from ng cli (already done here)
- React
  - init without CSS (to do after the other PR is merged)
  - init with minimal CSS, with the same one than Vite+React

But maybe I'm wrong and it could be too complex?
But I think as all the logic is in domain, changing small index.html / css, etc would be simple I think.

pascalgrimaud:
@qmonmert : just give me 1 day or 2, to think more about it, maybe it would be simpler to have 1 similar CSS for all front, as you suggested :-)

pascalgrimaud:
Can you use this version of logo instead plz ?
https://github.com/jhipster/jhipster-artwork/blob/main/logos/lite/JHipster-Lite-neon-red.png

pascalgrimaud:
I applied some fix, specially to display the appName correctly in the home page
Thanks for your work @qmonmert

pascalgrimaud (approved):

PR #802, opened 2022-02-22
improve absolut path + add test:watch

flcarre:


pascalgrimaud:
Can you have a look at this PR https://github.com/jhipster/jhipster-lite/pull/806/files ?
I'd like to have the alias @ = src/main/webapp/app, as it's similar to what is done in main generator-jhipster

pascalgrimaud (approved):

PR #829, opened 2022-02-24
Start script react vue

### LINKED ISSUE #821: Using same script to start front and use same port
Each front uses his own script:
- Angular: `npm run start`
- React: `npm run dev`
- Vue: `npm run dev`

I think we should have at least 1 common same script to start the front
So I suggest to add: `start` for React and Vue, so the users from generator-jhipster won't change
But for React and Vue, we can keep `dev`, which could be used too

Then, I suggest to change the port, and use the same one in generator-jhipster too, so it should be localhost:9000 for all fronts

matthieulapatate:
related to #821 
Changed the launching port and add start script to package json for react and vue.


pascalgrimaud:
this PR should not close the ticket as there are still some work to do (for Vue, for Angular)
So instead of writing `fix #xxxx`, it should be something like `related to #xxxx`

PR #844, opened 2022-02-24
JHLite: add start script and update port [ci skip]

pascalgrimaud:
- following https://github.com/jhipster/jhipster-lite/pull/840


pascalgrimaud:
@nonomoho : for your information, this PR will change the port, when you use `npm run dev`, it will be http://localhost:9000 instead

PR #880, opened 2022-02-28
Spring Boot Mvc: refactoring with constants

pascalgrimaud:


src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/web/application/SpringBootMvcApplicationServiceIT.java
```diff
@@ -241,7 +241,7 @@ private void assertExceptionHandlerProperties(Project project) {
   }
 
   private void assertExceptionHandlerFiles(Project project) {
-    String packagePath = "com/mycompany/myapp/technical/infrastructure/primary/exception";
+    String packagePath = getPath(project.getPackageNamePath().orElse("com/mycompany/myapp"), INFRA_PRIMARY, "exception");
```
Bolo89, line 244: I think that it's better to throw an expection when the 'mandatory' package name path does not exist.
```java
String packagePath = getPath(project.getPackageNamePath().orElseThrow(), INFRA_PRIMARY, "exception")
```
Or put "com/mycompany/myapp" in a constant for tests ?

Bolo89 (approved):

Issue #584, opened 2022-01-23
Reduce output log in generated projects


pascalgrimaud:
When launching tests in a generated projects, there are still a lot of useless logs
It means we should add more logger name to ignore in `logback.xml`

MSaguer:
I can work in this

pascalgrimaud:
Thanks @MSaguer 
If possible, small PR in each contexts would be cool :-)

MSaguer:
@pascalgrimaud, I'm struggling with the traces generated by LogstashTcpLifeCycleTest unit test.
```
22:24:02,504 |-WARN in net.logstash.logback.appender.LogstashTcpSocketAppender[ASYNC_LOGSTASH_TCP] - Log destination localhost/127.0.0.1:5000: connection failed. java.net.ConnectException: Connection refused: no further information
	at java.net.ConnectException: Connection refused: no further information
	at 	at java.base/sun.nio.ch.Net.pollConnect(Native Method)
	at 	at java.base/sun.nio.ch.Net.pollConnectNow(Net.java:672)
Net.java:672
	at 	at java.base/sun.nio.ch.NioSocketImpl.timedFinishConnect(NioSocketImpl.java:542)
NioSocketImpl.java:542
	at 	at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:597)
NioSocketImpl.java:597
	at 	at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:327)
SocksSocketImpl.java:327
	at 	at java.base/java.net.Socket.connect(Socket.java:633)
Socket.java:633
	at 	at net.logstash.logback.appender.AbstractLogstashTcpSocketAppender$TcpSendingEventHandler.openSocket(AbstractLogstashTcpSocketAppender.java:762)
AbstractLogstashTcpSocketAppender.java:762
	at 	at net.logstash.logback.appender.AbstractLogstashTcpSocketAppender$TcpSendingEventHandler.onStart(AbstractLogstashTcpSocketAppender.java:681)
AbstractLogstashTcpSocketAppender.java:681
	at 	at net.logstash.logback.appender.AsyncDisruptorAppender$EventClearingEventHandler.onStart(AsyncDisruptorAppender.java:382)
AsyncDisruptorAppender.java:382
	at 	at net.logstash.logback.encoder.com.lmax.disruptor.BatchEventProcessor.notifyStart(BatchEventProcessor.java:224)
[...]

PR #944, opened 2022-03-05
Reduce output log in generated projects

MSaguer:
Part of #584

logback conf to reduced the outputs log for tests


MSaguer:
As `LogstashTcpLifeCycle` requires too much work, this is the last MR to reduce unit test logs.

PR #948, opened 2022-03-06
Duplicate logback configuration to test application properties

MSaguer:
Part of #584


pascalgrimaud:
Just a question @MSaguer : do you think it's better to add logger in application.properties instead of logback?

Logback is used for unit test and integration test.
application.properties is used only for integration test.

So unless I'm wrong, using logback would be enought

MSaguer:
`logging.config=classpath:logback.xml` should be added in test application properties instead of duplicating all logger configuration in test application.properties.

I close this PR and a new one will be opened.

Issue #990, opened 2022-03-10
CSS to choose


pascalgrimaud:
Need to be discussed first.

The final app should have less than 5 pages I think:
- account for login, register, forgot password (only for the Cloud version)
- Spring Boot generator
- maybe one day Quarkus generator
- Micronaut generator
- etc

Several options here:
- build our own CSS, and embedded in the app -> most simplest
- use an existing CSS (but we won't use all elements, so maybe it's overkill?)
  - bootstrap
  - material
  - etc

Franceq34:
I'm open to discuss about it. 
My thoughts is that we shouldnt use bootstrap or materialize as they are too complex for our small needs. However, Tailwind could be great. 
I would like to mention @nonomoho and @seraphinandrieux here 🎨

seraphinandrieux:
@Franceq34 I'm glad to see you have thought to me for this subject ^^
On my side, i'm not an expert at all about CSS subjects. I only had the opportunity to use a custom Pattern Library.
However I agree on the idea to go to the simplest option for now.
Maybe Later we will have to refactoring this part...

pascalgrimaud:
After thinking about this, it took too much time to start from scratch.
So the idea would be to use an existing CSS.

I tried [quasar](https://quasar.dev/) and I'm really happy with this, so I think we should go in this direction

pascalgrimaud:
It was hard to write unit tests with Quasar, so the current choice is to choose simple bootstrap. We use it since several years with generator-jhipster, it's simple, it works well.
So I'm fine to use it

atomfrede:
I used tailwind and like it very much, but I guess bootstrap is the best choice as it is used widely troughout the industry and can be custimized very well if desired for the use case.

pascalgrimaud:
yes we choose bootstrap
I'm waiting @sabrinapayet to submit some pull requests to improve this part
Then, we'll be able to add more button/api to generate all the features

pascalgrimaud:
I think it was done!
Thanks @sabrinapayet for your work!

sabrinapayet:
Bug bounty claim : https://opencollective.com/generator-jhipster/expenses/73694

pascalgrimaud:
@sabrinapayet : approved

PR #1016, opened 2022-03-13
Add CSS

sabrinapayet:
- related to https://github.com/jhipster/jhipster-lite/issues/990


pascalgrimaud:
Thanks a lot @sabrinapayet !

@Franceq34 : I'm going to merge this, to avoid conflict when we'll add new buttons, etc. And I think it's already far better than what we have initially. Then, don't hesitate to improve it.

pascalgrimaud (approved):

PR #1017, opened 2022-03-13
Retrieve commands history from API

Franceq34:
- related https://github.com/jhipster/jhipster-lite/issues/1015


src/main/webapp/app/common/secondary/ProjectHistoryRepository.ts
```diff
@@ -0,0 +1,13 @@
...
+import { RestHistory, toHistory } from '@/common/secondary/RestHistory';
+import { History } from '@/common/domain/History';
+import { ProjectHistoryService } from '@/common/domain/ProjectHistoryService';
+import { AxiosHttp } from '@/http/AxiosHttp';
+
+export default class ProjectHistoryRepository implements ProjectHistoryService {
+  constructor(private axiosHttp: AxiosHttp) {}
+
+  async get(folder: Folder): Promise<History> {
+    return this.axiosHttp.get<RestHistory>(`api/projects/history?folder=${folder}`).then(response => toHistory(response.data));
```
seraphinandrieux, line 11: What do you think about request param folder into params attribute into get config ? Something like : `this.axiosHttp.get<RestHistory>('api/projects/history',{
params:{
folder}})`

seraphinandrieux, line 11: What do you think about request param folder into params attribute into get config ? Something like : `this.axiosHttp.get<RestHistory>('api/projects/history',{
params:{
folder}})`

src/test/javascript/spec/common/secondary/ProjectHistoryRepository.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+describe('ProjectRepository', () => {
+  it('should get project history', async () => {
+    const axiosHttpStub = stubAxiosHttp();
+    axiosHttpStub.get.resolves({ data: createRestHistory() });
+    const projectHistoryRepository = new ProjectHistoryRepository(axiosHttpStub);
+
+    const history: History = await projectHistoryRepository.get('folder/path');
+
+    const [uri] = axiosHttpStub.get.getCall(0).args;
```
seraphinandrieux, line 15: If you use config params, you could separate uri from params and do a separate assertion.
`const [uri,params] = axiosHttpStub.get.getCall(0).args;
expect(uri).toBe('api/projects/history');
expect(params).toBe({folder:'folder/path'});

`

seraphinandrieux, line 15: If you use config params, you could separate uri from params and do a separate assertion.
`const [uri,params] = axiosHttpStub.get.getCall(0).args;
expect(uri).toBe('api/projects/history');
expect(params).toBe({folder:'folder/path'});

`

pascalgrimaud (approved):

PR #1091, opened 2022-03-18
[Angular] Add account with authentication JWT

### LINKED ISSUE #898: Angular: account with authentication JWT

qmonmert:
Fix #898


pascalgrimaud:
@qmonmert : you need to add the new API in generator.sh, otherwise, it won't be tested in CI

qmonmert:
@pascalgrimaud where is this file? I don't see yet I rebase main branch

pascalgrimaud:
@qmonmert : here is the file https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L183

You need to call your new API here

qmonmert:
@pascalgrimaud okay it's `generate.sh` not `generator.sh` :p

pascalgrimaud:
My bad, I'm sorry for this typo :-p

pascalgrimaud:
I started to review it.
Globally, the authent works, but I think we need to reorganize all these files, as it does not totally respect Hexagonal Architecture.
I'll propose you how we can reorganize it, but it would be easier to propose it directly in a sample project, rather than the generated part.

qmonmert:
@pascalgrimaud ok

pascalgrimaud:
Here the repo I use for testing and reviewing: 
https://github.com/pascalgrimaud/tmp-jhlite-angularapp/pull/1

pascalgrimaud:
After studying the architecture, I don't have enough skill to propose a good folder to totally respect Hexagonal Architecture with Angular. So I think we can have features by package, a little bit similar to what we have in generator-jhipster.

Can you refactorize and use this plz?
image

In another PR, we could refactorize "common" to use another like "app", as it's not necessary to have 'primary' if we can't split correctly in primary/domain/secondary

Another comment:
- I use angular styled, followed by angular+jwt and I lost all styled in app.component.html

qmonmert:
@pascalgrimaud ok I will do that!

qmonmert:
@pascalgrimaud refacto done + fix angular styled jwt

src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularDomainService.java
```diff
@@ -18,6 +18,11 @@
   public static final String SOURCE_PRIMARY = getPath(SOURCE, APP);
   public static final String DESTINATION_PRIMARY = APP;
 
+  private static final String APP_MODULE = "app.module.ts";
+  private static final String APP_COMPONENT = "app.component.ts";
+  private static final String APP_COMPONENT_SPEC = "app.component.spec.ts";
+  private static final String APP_COMPONENT_HTML = "app.component.html";
```
pascalgrimaud, line 24: can be moved to Angular.java

src/main/resources/generator/client/angular/proxy.conf.json
```diff
@@ -0,0 +1,6 @@
+{
+  "/api": {
+    "target": "http://localhost:8080",
```
pascalgrimaud, line 3: mhh I think it should be replaced by serverPort property

src/test/java/tech/jhipster/lite/generator/client/angular/core/application/AngularAssert.java
```diff
@@ -68,6 +68,21 @@ public static void assertAppWithCss(Project project) {
...
+
+    assertFileExist(project, getPath(pathAuth, "account.model.ts"));
+    assertFileExist(project, getPath(pathAuth, "account.service.ts"));
+    assertFileExist(project, getPath(pathAuth, "account.service.spec.ts"));
+    assertFileExist(project, getPath(pathAuth, "auth-jwt.service.ts"));
+    assertFileExist(project, getPath(pathAuth, "auth-jwt.service.spec.ts"));
+    assertFileExist(project, getPath(pathAuth, "auth.interceptor.ts"));
+    assertFileExist(project, getPath(pathLogin, "login.service.ts"));
+    assertFileExist(project, getPath(pathLogin, "login.service.spec.ts"));
+    assertFileExist(project, getPath(pathLogin, "login.model.ts"));
```
pascalgrimaud, line 83: I'm pretty sure we can do better, by using Angular classes, with forEach

pascalgrimaud (approved):

PR #1092, opened 2022-03-19
Move dependabot job to run after all other jobs

### LINKED ISSUE #1081: Dependabot: auto merge before other workflows
- following https://github.com/jhipster/jhipster-lite/issues/1002
- 
It seems to not work as intended, as sometimes, it's merged before all other workflows run
So the best would be to move at the end of workflow

beatfreaker:


pascalgrimaud:
Thanks a lot @beatfreaker !
Let's try this solution

.github/workflows/github-actions.yml
```diff
@@ -219,3 +219,24 @@ jobs:
...
+  # Dependabot auto merge PR
+  #--------------------------------------------------
+  dependabot-auto-merge-pr:
+    name: dependabot-auto-merge-pr
+    needs: [cypress, generation]
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'jhipster/jhipster-lite' && github.event.pull_request.user.login == 'dependabot[bot]' }}
```
pascalgrimaud, line 232: Maybe we need to exclude branch main, as this workflow will run in branch main too and I don't know what happened for main branch

Issue #1130, opened 2022-03-22
Kafka: DummyProducerIT should not be generated


pascalgrimaud:
When adding Kafka, DummyProducerIT is generated, but it should not.
It should be generated only when adding Dummy Producer

See https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/server/springboot/broker/kafka/domain/KafkaDomainService.java#L138

PR #1131, opened 2022-03-22
Kafka: fix generation of DummyProducerIT

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/1130


fdelbrayelle (dismissed):
OK but `KafkaDomainServiceTest` has to be fixed

Issue #1132, opened 2022-03-23
Allow user to download project as zip file.


matthieulapatate:
The idea would be to allow users to upload their project as a zip by clicking a button in the webapp.

pascalgrimaud:
Do you mean download the generated project instead ?
Because I don't understand why the user should upload the code to JHipster Lite

matthieulapatate:
Yes dowload, it would be another option than just directly creating in their repository.

pascalgrimaud:
So, can you update the title plz?

pascalgrimaud:
The 1st step would be to add this to `ProjectRepository.java` :

```java
void zip(Project project);
```

Then, you need to work on the implementation

pascalgrimaud:
@matthieulapatate : I'm waiting this 1st small PR, which should contain only:
- the method in interface
- the implem in secondary
- the unit test
- maybe the lib in pom.xml

Nothing more.

pascalgrimaud:
closed by https://github.com/jhipster/jhipster-lite/pull/1266

PR #1146, opened 2022-03-24
Add backend secondary zip function

matthieulapatate:
- Related to #1132 
One part of this issue is dealed in this PR. It relies to springboot generation and API response with zip file.
A bounded countext has been created and a method has been added to ProjectRepository Interface.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,8 @@
     <problem-spring-web.version>${problem-spring.version}</problem-spring-web.version>
     <properties-maven-plugin.version>1.1.0</properties-maven-plugin.version>
     <sonar-maven-plugin.version>3.9.1.2184</sonar-maven-plugin.version>
+    <commons-io.version>2.11.0</commons-io.version>
+    <zt-zip.version>1.11</zt-zip.version>
 
     <frontend-maven-plugin.version>1.12.1</frontend-maven-plugin.version>
     <checksum-maven-plugin.version>1.11</checksum-maven-plugin.version>
@@ -74,6 +76,16 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>${commons-io.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.zeroturnaround</groupId>
+      <artifactId>zt-zip</artifactId>
+      <version>${zt-zip.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,6 @@
     <problem-spring-web.version>${problem-spring.version}</problem-spring-web.version>
     <properties-maven-plugin.version>1.1.0</properties-maven-plugin.version>
     <sonar-maven-plugin.version>3.9.1.2184</sonar-maven-plugin.version>
-    <commons-io.version>2.11.0</commons-io.version>
     <zt-zip.version>1.11</zt-zip.version>
 
     <frontend-maven-plugin.version>1.12.1</frontend-maven-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -78,11 +78,6 @@
       <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
-      <version>${commons-io.version}</version>
-    </dependency>
-    <dependency>
       <groupId>org.zeroturnaround</groupId>
       <artifactId>zt-zip</artifactId>
       <version>${zt-zip.version}</version>
--- a/src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
+++ b/src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
@@ -25,4 +25,6 @@ public interface ProjectRepository {
   void gitInit(Project project);
   void gitAddAndCommit(Project project, String message);
   void gitApplyPatch(Project project, String patchFilename);
+
+  void zip(Project project);
 }
--- a/src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
+++ b/src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
@@ -209,8 +209,7 @@ public class ProjectLocalRepository implements ProjectRepository {
     try {
       ZipUtil.pack(workingDir, new File(workingDir + ".zip"));
     } catch (ZipException e) {
-      log.error("This folder doesn't exist");
-      throw (e);
+      throw new GeneratorException("Error when zipping " + project.getFolder());
     }
   }
 
--- a/src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
+++ b/src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
@@ -209,7 +209,7 @@ public class ProjectLocalRepository implements ProjectRepository {
     try {
       ZipUtil.pack(workingDir, new File(workingDir + ".zip"));
     } catch (ZipException e) {
-      throw new GeneratorException("Error when zipping " + project.getFolder());
+      throw new GeneratorException("Error when zipping " + project.getFolder(), e);
     }
   }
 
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -72,8 +72,14 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.11.0</version>
```
pascalgrimaud, line 77: version should not be here, but in properties instead

src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
```diff
@@ -1,5 +1,7 @@
 package tech.jhipster.lite.generator.project.domain;
 
+import org.springframework.http.ResponseEntity;
```
pascalgrimaud, line 3: it's totally forbidden. Domain should not use external framework like springframework

src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
```diff
@@ -13,10 +14,17 @@
...
 import java.util.HashSet;
 import java.util.Set;
+import java.util.UUID;
+import org.apache.commons.io.IOUtils;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
```
pascalgrimaud, line 24: we are in secondary, so there should not be import like this, which are related to primary (rest)

pom.xml
```diff
@@ -72,8 +72,14 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
pascalgrimaud, line 76: If it's just used for converting inputstream to byte array, we don't need this

src/main/resources/generator/dependencies/common/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "JHipster Lite : used for Dependencies",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@prettier/plugin-xml": "1.2.0",
+    "@prettier/plugin-xml": "2.0.1",
```
pascalgrimaud, line 7: not the scope of this PR

src/test/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepositoryTest.java
```diff
@@ -392,4 +396,27 @@ void shouldNotRename() {
 
     assertFileNotExist(project, POM_XML);
   }
+
+  @Test
+  void shouldZip() {
+    Project project = tmpProjectWithPomXml();
+    assertNotNull(repository.zip(project));
```
pascalgrimaud, line 403: use assertj instead to keep consistency with the other tests. Then, it would be better to test the filename of zip, for example

src/test/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepositoryTest.java
```diff
@@ -392,4 +396,27 @@ void shouldNotRename() {
...
+  }
+
+  @Test
+  void shouldNotZipWithNonExistingFolder() {
+    Project project = tmpProject();
+    assertThat(repository.zip(project).getStatusCode() == HttpStatus.INTERNAL_SERVER_ERROR);
```
pascalgrimaud, line 409: there should not be httpstatus in secondary, so this test needs to be rework

src/test/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepositoryTest.java
```diff
@@ -392,4 +396,27 @@ void shouldNotRename() {
...
+  }
+
+  @Test
+  void shouldNotZip() {
+    Project project = tmpProjectWithPomXml();
+
+    try (MockedStatic<IOUtils> ioUtilsMockedStatic = Mockito.mockStatic(IOUtils.class)) {
+      ioUtilsMockedStatic.when(() -> IOUtils.toByteArray(any(InputStream.class))).thenThrow(new IOException());
+
+      assertThat(repository.zip(project).getStatusCode() == HttpStatus.INTERNAL_SERVER_ERROR);
```
pascalgrimaud, line 419: there should not be httpstatus in secondary, so this test needs to be rework

pom.xml
matthieulapatate: Is there an existing method already doing it ?

src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
matthieulapatate: How can i describe what's returned ? is the responseEntity already implemented in return apis ?

pom.xml
pascalgrimaud: search byte array to inputstream or inputstream to byte array -> https://www.baeldung.com/convert-byte-array-to-input-stream

src/main/java/tech/jhipster/lite/generator/project/domain/ProjectRepository.java
pascalgrimaud: void or byte[]

pom.xml
```diff
@@ -54,6 +54,8 @@
     <problem-spring-web.version>${problem-spring.version}</problem-spring-web.version>
     <properties-maven-plugin.version>1.1.0</properties-maven-plugin.version>
     <sonar-maven-plugin.version>3.9.1.2184</sonar-maven-plugin.version>
+    <commons-io.version>2.11.0</commons-io.version>
```
pascalgrimaud, line 57: no need

src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
```diff
@@ -201,6 +203,17 @@ public void gitApplyPatch(Project project, String patchFilename) {
     }
   }
 
+  @Override
+  public void zip(Project project) {
+    File workingDir = new File(project.getFolder());
+    try {
+      ZipUtil.pack(workingDir, new File(workingDir + ".zip"));
+    } catch (ZipException e) {
+      log.error("This folder doesn't exist");
```
pascalgrimaud, line 212: it's not true, there can be another problem when zipping. Simply change to something like: 'Error when zipping {}', with the folder name as parameter

src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
```diff
@@ -201,6 +203,17 @@ public void gitApplyPatch(Project project, String patchFilename) {
...
+      throw (e);
```
pascalgrimaud, line 213: better to throw GeneratorException

pascalgrimaud (changes requested):

pascalgrimaud:
I suppose the work is finished but it's still in draft probably you forgot to change it.
Anyway I made some comments

src/main/java/tech/jhipster/lite/generator/project/infrastructure/secondary/ProjectLocalRepository.java
```diff
@@ -209,8 +209,7 @@ public void zip(Project project) {
     try {
       ZipUtil.pack(workingDir, new File(workingDir + ".zip"));
     } catch (ZipException e) {
-      log.error("This folder doesn't exist");
-      throw (e);
+      throw new GeneratorException("Error when zipping " + project.getFolder());
```
pascalgrimaud, line 212:
```suggestion
      throw new GeneratorException("Error when zipping " + project.getFolder(), e);
```

pascalgrimaud (changes requested):

matthieulapatate:
Is there any problem with codecov ? sonar analyse has 100% passed on local

pascalgrimaud:
@matthieulapatate : see the logs:

image

There was a problem when uploading analysis to codecov. I'm restarting the job

pascalgrimaud (approved):

PR #1179, opened 2022-03-28
Dependabot: Update sonar docker image

DanielFran:


pascalgrimaud:
Thanks @DanielFran 
But for your information, it won't do anything. Dependabot doesn't work with docker-compose files, only with Dockerfile and there is no Dockerfile in this folder for now :-)

DanielFran:
@pascalgrimaud you are right, little tired yesterday... :P

PR #1188, opened 2022-03-29
Add pinia for vue 3

### LINKED ISSUE #1117: [VueJS] Add Pinia into Vue generator + generated
We could add an API to add Pinia which replaces VueX for state management into a Vue app

seraphinandrieux:
- Fix #1117

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,8 @@
   },
   "dependencies": {
     "axios": "0.26.1",
+    "pinia": "2.0.12",
+    "pinia-plugin-persist": "1.0.0",
     "vue": "3.2.31",
     "vue-router": "4.0.14"
   },
--- a/src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/Vue.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/Vue.java
@@ -1,16 +1,34 @@
 package tech.jhipster.lite.generator.client.vue.core.domain;
 
+import java.util.Collection;
 import java.util.List;
 
 public class Vue {
 
   public static final String ROUTER_IMPORT = "import router from \'./router/router\';";
   public static final String ROUTER_PROVIDER = "app.use(router);";
+  public static final Collection<String> PINIA_IMPORTS = List.of(
+    "import {createPinia} from 'pinia';",
+    "import piniaPersist from 'pinia-plugin-persist'"
+  );
+  public static final Collection<String> PINIA_PROVIDERS = List.of(
+    "const pinia = createPinia();",
+    "pinia.use(piniaPersist);",
+    "app.use(pinia);"
+  );
 
   private Vue() {}
 
   public static List<String> dependencies() {
     return List.of("vue");
+  }
+
+  public static List<String> piniaDependencies() {
+    return List.of("pinia", "pinia-plugin-persist");
+  }
+
+  public static List<String> piniaDevDependencies() {
+    return List.of("@pinia/testing");
   }
 
   public static List<String> routerDependencies() {
--- a/src/main/java/tech/jhipster/lite/generator/client/vue/core/infrastructure/primary/rest/VueResource.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/vue/core/infrastructure/primary/rest/VueResource.java
@@ -41,4 +41,13 @@ class VueResource {
     Project project = ProjectDTO.toProject(projectDTO);
     vueApplicationService.addStyledVue(project);
   }
+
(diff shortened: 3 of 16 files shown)


package.json
```diff
@@ -31,6 +31,8 @@
   },
   "dependencies": {
     "axios": "0.26.1",
+    "pinia": "2.0.12",
+    "pinia-plugin-persist": "^1.0.0",
```
pascalgrimaud, line 35: always use fixed version

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/Vue.java
```diff
@@ -1,18 +1,66 @@
...
 
+import java.util.Collection;
 import java.util.List;
 
 public class Vue {
 
   public static final String ROUTER_IMPORT = "import router from \'./router/router\';";
   public static final String ROUTER_PROVIDER = "app.use(router);";
+  public static final Collection<String> PINIA_IMPORTS = List.of(
+    "import {createPinia} from \"pinia\";",
```
pascalgrimaud, line 11:
```suggestion
    "import {createPinia} from 'pinia';",
```

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/Vue.java
```diff
@@ -1,18 +1,66 @@
...
+  public static List<String> piniaDependencies() {
+    return List.of("pinia", "pinia-plugin-persist");
+  }
+
+  public static List<String> piniaDevDependencies() {
+    return List.of("@pinia/testing");
+  }
+
+  public static List<String> buttonForStoreExampleFiles() {
+    return List.of("Button.component.ts", "Button.html", "Button.vue", "index.ts");
```
pascalgrimaud, line 35: need to be discussed, not sure about the use of these files. I know it is used as example, but in the end, the user will delete it for sure

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/Vue.java
seraphinandrieux: Ok no problemo, however if I delete these ones, I'm asking if I have also to delete CounterStore file.
Without the button this store is useless. 
What do you think ?

pascalgrimaud: Yes we can delete the button. Anyway, this store will be used very soon, to store the token so the example is not really usefull imo

src/main/java/tech/jhipster/lite/generator/client/vue/core/infrastructure/primary/rest/VueResource.java
```diff
@@ -41,4 +41,13 @@ public void addStyledVue(@RequestBody ProjectDTO projectDTO) {
     Project project = ProjectDTO.toProject(projectDTO);
     vueApplicationService.addStyledVue(project);
   }
+
+  @Operation(summary = "Add Pinia", description = "Add pinia for state management")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding Pinia")
+  @PostMapping("/pinia")
+  @GeneratorStep(id = "vue-pinia")
```
pascalgrimaud, line 48: can you add and use new constant in GeneratorAction plz ?

src/main/java/tech/jhipster/lite/generator/client/vue/core/infrastructure/primary/rest/VueResource.java
```diff
@@ -41,4 +41,13 @@ public void addStyledVue(@RequestBody ProjectDTO projectDTO) {
     Project project = ProjectDTO.toProject(projectDTO);
     vueApplicationService.addStyledVue(project);
   }
+
+  @Operation(summary = "Add Pinia", description = "Add pinia for state management")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding Pinia")
+  @PostMapping("/pinia")
```
pascalgrimaud, line 47: it would be important to call this api in generator.sh, so it will be really test in CI

pascalgrimaud (changes requested):
few last comments and it would be good

pascalgrimaud (approved):

seraphinandrieux:
$200 bounty claimed => https://opencollective.com/generator-jhipster/expenses/70882

pascalgrimaud:
@seraphinandrieux : approved

PR #1247, opened 2022-04-04
Add secondary for project download

matthieulapatate:


src/main/webapp/app/springboot/secondary/ProjectRepository.ts
```diff
@@ -25,4 +25,12 @@ export default class ProjectRepository implements ProjectService {
     const restProject: RestProject = toRestProject(project);
     await this.axiosHttp.post('api/servers/java/base', restProject);
   }
+
+  async download(project: Project): Promise<Blob> {
+    const restProject: RestProject = toRestProject(project);
+    return this.axiosHttp
+      .post<Blob, RestProject>('api/projects/download', restProject)
+      .then(response => new Blob([response.data]))
```
seraphinandrieux, line 33: I prefer here to let [response.data] as promise return.
When you instantiate a blob, you can specify it which type you want it will be (an xlsx, json, ...).
So I think the secondary role is just to say : Ok here you have datas wanted, and you do what you want with that.
Like that into primary, if later you want to let user the type file choice, you only will have to change the type when you will instantiate the Blob

src/main/webapp/app/springboot/secondary/ProjectRepository.ts
```diff
@@ -25,4 +25,12 @@ export default class ProjectRepository implements ProjectService {
...
+      .catch(err => new Blob());
```
seraphinandrieux, line 34: Do you really want an empty blob when something wrong with server ?
I suggest to don't catch promise error here and let primary manages the case, with an error message for exemple.

src/test/javascript/spec/springboot/secondary/ProjectRepository.spec.ts
```diff
@@ -60,4 +60,31 @@ describe('ProjectRepository', () => {
     expect(uri).toBe('api/frontend-maven-plugin');
     expect(payload).toEqual<RestProject>(expectedRestProject);
   });
+
+  it('should download the project', () => {
+    const axiosHttpStub = stubAxiosHttp();
+    axiosHttpStub.post.resolves();
+    const projectRepository = new ProjectRepository(axiosHttpStub);
+    const project: Project = createProject({ folder: 'folder/path' });
```
seraphinandrieux, line 68: Extract it into a new File => Project.fixtures.ts

src/test/javascript/spec/springboot/secondary/ProjectRepository.spec.ts
```diff
@@ -60,4 +60,31 @@ describe('ProjectRepository', () => {
     expect(uri).toBe('api/frontend-maven-plugin');
     expect(payload).toEqual<RestProject>(expectedRestProject);
   });
+
+  it('should download the project', () => {
```
seraphinandrieux, line 64: You can go further in your test here with stubbing the axios response and check if your return is the one expected

src/test/javascript/spec/springboot/secondary/ProjectRepository.spec.ts
```diff
@@ -60,4 +60,31 @@ describe('ProjectRepository', () => {
...
+
+    projectRepository.download(project);
+
+    const expectedRestProject: RestProject = toRestProject(project);
+    const [uri, payload] = axiosHttpStub.post.getCall(0).args;
+    expect(uri).toBe('api/projects/download');
+    expect(payload).toEqual<RestProject>(expectedRestProject);
+  });
+
+  it('should not download the project', () => {
```
seraphinandrieux, line 78: As I told you, i think this case is not wanted here

src/main/webapp/app/springboot/secondary/ProjectRepository.ts
matthieulapatate: THanks for the commentary, here if i don't specify a catch block, i get an error saying 
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TypeError: Cannot read properties of undefined (reading 'data')".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
As you say, i didnt test to catch the error on primary because it is not the main goal of this PR. If i want to try it, i should have to implement the front (because of generator component test need to get the button associated).
That's the reason why i returned an empty blob in order to test the size once get on primary size (meaning there was an error occured)

seraphinandrieux: You totally right about the error.
As the error says to you : This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().

So it means you didn't manage an error => So the your test seems to be in the error case and finally it was throw by the catch did before.

So in this case the solution is to pass your test as async and pass to your stub the correct resolve.

Like that you only test your success case (you can also test your code throws an error if no data but not really usefull)
And you will manage you error case in your primary which will catch the secondary exception

src/main/webapp/app/springboot/domain/ProjectService.ts
```diff
@@ -5,4 +5,5 @@ export interface ProjectService {
   addMaven(project: Project): Promise<void>;
   addFrontendMavenPlugin(project: Project): Promise<void>;
   addJavaBase(project: Project): Promise<void>;
+  download(project: Project): Promise<ArrayBuffer>;
```
seraphinandrieux, line 8: Hum not sure about that, especially if you want a Blob at the end.
I suggest `BlobPart[]`

src/main/webapp/app/springboot/secondary/ProjectRepository.ts
```diff
@@ -25,4 +25,11 @@ export default class ProjectRepository implements ProjectService {
     const restProject: RestProject = toRestProject(project);
     await this.axiosHttp.post('api/servers/java/base', restProject);
   }
+
+  async download(project: Project): Promise<ArrayBuffer> {
+    const restProject: RestProject = toRestProject(project);
+    return this.axiosHttp
+      .post<ArrayBufferLike, RestProject>('api/projects/download', restProject)
```
seraphinandrieux, line 32: I suggest into the headers config to set the content type (`'application/json;charset=UTF-8'`) and the Accept field  type (in this case check which application file type do you expect)

src/main/webapp/app/springboot/secondary/ProjectRepository.ts
```diff
@@ -25,4 +25,11 @@ export default class ProjectRepository implements ProjectService {
...
+      .then(response => new Uint8Array(response.data));
```
seraphinandrieux, line 33: If you use BlobPart[] return type, here you just have to return [res.data]

src/test/javascript/spec/springboot/secondary/ProjectRepository.spec.ts
```diff
@@ -60,4 +60,16 @@ describe('ProjectRepository', () => {
     expect(uri).toBe('api/frontend-maven-plugin');
     expect(payload).toEqual<RestProject>(expectedRestProject);
   });
+
+  it('should download the project', async () => {
+    const [projectRepository, axiosHttpStub] = createStubedProjectRepository({ data: [1, 2, 3] });
+    const project: Project = createProject({ folder: 'folder/path' });
+
+    expect(await projectRepository.download(project)).toEqual(new Uint8Array([1, 2, 3]));
```
seraphinandrieux, line 68: extract the method call just above the expect statement.

In an unit test I suggest to have 3 paragraphs, one which gives the stubs used, another for calling the correct methods, and another which does assertions

src/test/javascript/spec/springboot/domain/Project.fixture.ts
```diff
@@ -1,4 +1,7 @@
 import { Project } from '@/springboot/domain/Project';
+import ProjectRepository from '../../../../../main/webapp/app/springboot/secondary/ProjectRepository';
+import { stubAxiosHttp } from '../../http/AxiosHttpStub';
+import { AxiosHttpResponse } from '../../../../../main/webapp/app/http/AxiosHttp';
```
pascalgrimaud, line 4: you should be able to use alias with `@`

pascalgrimaud:
@seraphinandrieux : don't hesitate to ping me when you think it's ready, so I can merge this

seraphinandrieux:
> @seraphinandrieux : don't hesitate to ping me when you think it's ready, so I can merge this

Ok on my side

PR #1289, opened 2022-04-08
Naming convention for APIs

### LINKED ISSUE #1273: [REST-compliance] Naming convention for APIs
Following https://github.com/jhipster/jhipster-lite/issues/144

Try to have something like `<collection>/<id>/<collection>/<id>/.../<element>` (URL "leaf" can be either a collection, an id or an element):
-  `/api/servers/spring-boot/databases/mysql` is compliant
- `/api/svelte/styled` is not compliant, could be renamed to something like `/api/clients/svelte/style`
- `/api/servers/spring-boot/mvc/web/springdoc/init-with-security-jwt` is not compliant, could be renamed to something like `/api/servers/spring-boot/api-documentations/springdoc/security-systems/jwt`

This is to have a self explaining API and ready to add new options for each "tool" (ie: `oauth2` security method instead of `jwt`).

> before renaming an API, be sure to understand what it generates/does, that's the hard part :)

hdurix:
- Fix #1273


pascalgrimaud:
I think the auto merged is broken as it didn't wait the CI to finish, that's why the current CI failed
I'll have a look once at home

hdurix:
bounty claimed https://opencollective.com/generator-jhipster/expenses/89124

pascalgrimaud:
@hdurix : approved

PR #1295, opened 2022-04-09
Init project without package.json + husky

### LINKED ISSUE #1227: Init project without package.json + husky
Some project don't need `package.json` and husky config

So it would be nice to add a new API for adding only:
- readme
- editor config
- git config

qmonmert:
Fix #1227

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/init/infrastructure/primary/rest/InitResource.java
+++ b/src/main/java/tech/jhipster/lite/generator/init/infrastructure/primary/rest/InitResource.java
@@ -38,6 +38,15 @@ class InitResource {
     initApplicationService.init(project);
   }
 
+  @Operation(summary = "Init minimal project")
+  @ApiResponse(responseCode = "500", description = "An error occurred while initializing minimal project")
+  @PostMapping("/projects-minimal")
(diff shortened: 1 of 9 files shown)


src/main/java/tech/jhipster/lite/generator/init/infrastructure/primary/rest/InitResource.java
```diff
@@ -38,6 +38,15 @@ public void init(@RequestBody ProjectDTO projectDTO) {
     initApplicationService.init(project);
   }
 
+  @Operation(summary = "Init minimal project")
+  @ApiResponse(responseCode = "500", description = "An error occurred while initializing minimal project")
+  @PostMapping("/projects-minimal")
+  @GeneratorStep(id = GeneratorAction.INIT)
```
pascalgrimaud, line 44: Should be a new id here

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/init/infrastructure/primary/rest/InitResource.java
qmonmert: done

pascalgrimaud (approved):

Issue #1296, opened 2022-04-09
Add playwright as additional e2e framework


atomfrede:
I love playwright and I think it would be nice to have it here. There will be a blueprint for the jhipster "large" generator, but it takes more time to rewrite the tests. WDYT?

pascalgrimaud:
Yes we can take care of it, here :-)

pascalgrimaud:
@atomfrede : just gave a try and it didn't work well yet. Can you have a look ?

- start JH Lite
- generate an app : ./tests-ci/generate.sh reactapp
- go into /tmp/jhlite/reactapp
- start the app: ./mvnw
- launch npm run e2e

Some wrong config, maybe:

```diff
diff --git a/src/test/javascript/integration/common/primary/app/Home.spec.ts b/src/test/javascript/integration/common/primary/app/Home.spec.ts
index 64235ef..8f50026 100644
--- a/src/test/javascript/integration/common/primary/app/Home.spec.ts
+++ b/src/test/javascript/integration/common/primary/app/Home.spec.ts
@@ -1,11 +1,11 @@
 import { test, Page } from '@playwright/test';
 
-import { HomePage } from './home-page';
+import { HomePage } from './Home-Page';
 
 test.describe('Home', () => {
   test('display home page', async ({ page }) => {
     const homePage = new HomePage(page);
-    await homePage.visit();
+    await homePage.goto();
     await homePage.appLocator.waitFor();
   });
 });
```

atomfrede:
will Check in the afternoon.

PR #1373, opened 2022-04-16
fix playwright test spec

atomfrede:
Fixed the specification. Just noticed when you have e.g. already cypress added, and add playwright, the package json becomes invalid as mutiple scripts named e2e are created. Do we have already some means to prevent adding multiple e2e frameworks or multiple server side frameworks?

image

udpates #1296


pascalgrimaud:
Nothing to prevent this.
I would say:
- in a real project, there is already Cypress, if the user add another framework like Playwright, I would say it's a bad idea

Another exemple:
- in a real project, there is already Angular, if the user add React in this project, it won't work because it's a bad idea :-D

atomfrede:
👍 totally agree with you just wanted to make sure.

PR #1391, opened 2022-04-18
Use Java Records

qmonmert:


qmonmert:
@pascalgrimaud an idea, why coverage decreases?

pascalgrimaud:
@qmonmert : don't know. I already see the proposition from intellij to use record instead but I found the use very strange, that's why I prefered to use default class instead of record

PR #1449, opened 2022-04-23
Add generation timestamp in history file

### LINKED ISSUE #1445: Add generation timestamp in history file
For each generatorStep executed, a new entry is created in the `.jhipster/history.json` file with the following format

```
{
    "serviceId" : "<generatorStep name>"
}
```

If the same generation is executed many times, the same serviceId will be defined many times with exactly the same data.

This history service could be improved in two different ways
- add existence check on GeneratorHistory to avoid to add twice the same line (idem potency)
- add date field to store the generation date.

What do you think of this?

MSaguer:
To close #1445 

Add new timestamp field in history file.
For back compatibilty's sake, null timestamp values are authorized as timestamp values.

### DIFF
--- a/src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
+++ b/src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
@@ -1,0 +1,21 @@
+package tech.jhipster.lite.common.domain;
+
+import java.time.Clock;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class TimeUtils {
+
+  private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMddHHmmss");
+
+  private TimeUtils() {}
+
+  public static String getNowTimestamp(Clock clock) {
+    return getNowTimestamp(clock, 0);
+  }
+
+  public static String getNowTimestamp(Clock clock, int plusSecond) {
+    LocalDateTime localDateTime = LocalDateTime.now(clock).plusSeconds(plusSecond);
+    return localDateTime.format(DATE_TIME_FORMATTER);
+  }
+}
--- a/src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
+++ b/src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
@@ -10,11 +10,11 @@ public class TimeUtils {
 
   private TimeUtils() {}
 
-  public static String getNowTimestamp(Clock clock) {
-    return getNowTimestamp(clock, 0);
+  public static String getDateString(Clock clock) {
+    return getDateString(clock, 0);
   }
 
-  public static String getNowTimestamp(Clock clock, int plusSecond) {
+  public static String getDateString(Clock clock, int plusSecond) {
     LocalDateTime localDateTime = LocalDateTime.now(clock).plusSeconds(plusSecond);
     return localDateTime.format(DATE_TIME_FORMATTER);
   }
--- a/src/main/java/tech/jhipster/lite/generator/history/infrastructure/primary/GeneratorHistoryInterceptor.java
+++ b/src/main/java/tech/jhipster/lite/generator/history/infrastructure/primary/GeneratorHistoryInterceptor.java
@@ -5,7 +5,6 @@ import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.annotation.AfterReturning;
 import org.aspectj.lang.annotation.Aspect;
 import org.springframework.context.annotation.Configuration;
-import tech.jhipster.lite.common.domain.TimeUtils;
 import tech.jhipster.lite.generator.history.application.GeneratorHistoryApplicationService;
 import tech.jhipster.lite.generator.history.domain.GeneratorHistoryValue;
 import tech.jhipster.lite.generator.project.domain.Project;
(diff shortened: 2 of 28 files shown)


MSaguer:
Miss the HistoryDTO update to end this

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
```diff
@@ -0,0 +1,21 @@
+package tech.jhipster.lite.common.domain;
+
+import java.time.Clock;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class TimeUtils {
+
+  private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMddHHmmss");
```
Bolo89, line 9: Maybe use a standard date time format ? the timezone does not appear here (ISO)

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private TimeUtils() {}
+
+  public static String getNowTimestamp(Clock clock) {
+    return getNowTimestamp(clock, 0);
+  }
+
+  public static String getNowTimestamp(Clock clock, int plusSecond) {
```
Bolo89, line 17: Maybe i'm wrong but for me, timestamp is the time passed between 1970 and now. If I read the method name, I except a long value instead of a string

src/main/java/tech/jhipster/lite/generator/history/infrastructure/primary/GeneratorHistoryInterceptor.java
```diff
@@ -15,17 +17,19 @@
...
+    this.clock = clock;
   }
 
   @After(value = "@annotation(generatorStep)")
   public void addInHistory(JoinPoint joinPoint, GeneratorStep generatorStep) {
     String serviceId = generatorStep.id();
     ProjectDTO projectDTO = (ProjectDTO) joinPoint.getArgs()[0];
     Project project = ProjectDTO.toProject(projectDTO);
-    GeneratorHistoryValue generatorHistoryValue = new GeneratorHistoryValue(serviceId);
+    GeneratorHistoryValue generatorHistoryValue = new GeneratorHistoryValue(serviceId, TimeUtils.getNowTimestamp(clock));
```
Bolo89, line 32: clock.instant() is enough ? I think it's better to keep instant as parameter instead of string. Maybe tomorrow the history will be saved in database, and instant is standard

Bolo89 (changes requested):

pascalgrimaud:
@MSaguer : the review was done by @Bolo89, can you take care of it plz ? And resolve conflict too plz

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
MSaguer: I've changed this to getDateString. Is that ok for you?

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
MSaguer: That's the date time format used for the liquibase xml file naming. As it was used two times in 2 differents classes, I've just refactored existing code into one util class.
I don't think that behaviour shall be changed in this MR.

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
Bolo89: Yes

src/main/java/tech/jhipster/lite/common/domain/TimeUtils.java
Bolo89: OK this format is only used for file name

MSaguer:
@pascalgrimaud first modification batch.
Work in progress  : creation of GeneratorHistoryDTO to solve the ArchUnit test on domain dependency to Jackson annotation.

@Bolo89 I've change the timestamp type from String  to Instant. Is the new serialized format "yyyy-MM-dd HH:mm:ss" Ok for you?

Bolo89:
@MSaguer, in the format  "yyyy-MM-dd HH:mm:ss", the timezone / offset is skipped. Possible to have this format ? `2022-02-22T10:11:12.000Z` ? (used when you convert instant to string)

```java
@JsonFormat(pattern="yyyy-MM-dd'T'HH:mm:ss.SSSZ", timezone = "UTC")
```

or for global configuration of the objectMapper in JSonUtils.java
```java
new ObjectMapper()
                .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```

MSaguer:
thanks @Bolo89. Adding ObjectMapper configuration in JsonUtils was what I was lokking for.

MSaguer:
There is a sonar "code smell" that I can't easily solve
image

I think that the only solution is to add `@SuppressWarnings("java:S110")`

Do you agree?

pascalgrimaud:
Agree @MSaguer as it does not depend on our code

pascalgrimaud:
this part will probably have an impact with the front part, so I'll have a deep look + tests before merging this

Bolo89 (approved):

pascalgrimaud:
@MSaguer : can you solve conflict plz?

Issue #199, opened 2021-11-13
Project: add rules


pascalgrimaud:
- init : can be run without requirement
- maven : can be run without requirement
- spring boot: need build tool, maven or gradle
- spring boot web: need spring boot
- spring boot database : need spring boot
- spring boot liquibase : need sql database and spring boot
- java error : need java

pascalgrimaud:
No need any more, as we'll use module system
Link to https://github.com/jhipster/jhipster-lite/issues/1666

PR #1455, opened 2022-04-23
Init RequiredGeneratorStep

MSaguer:
Here is a quick proposal as a first step to answer #199.

I propose here to create a new RequiredGeneratorStep annotation to list all GeneratorStep that should have been executed before this step.
The GeneratorHistoryInterceptor has been transformed into GeneratorStepInterceptor to check that all required steps have been executed. If some GeneratorStep miss, the execution is stopped and an exception is raised.

What do you think of this?

If the principle is validated, I will improve the code and add tests.


pascalgrimaud:
In fact, it's not so easy, that's why it's not implemented yet.

I give you a use case:
- I already have an existing Spring Boot project, generated from scratch with Spring Initializr
- with my team, we added a lot of specific code to our project
- now, I'd like to add Spring Cloud Config with Consul
- I'd like to use JHipster Lite as I think it's possible and it should be possible

So the problem with this approach will block this use case, as there is no history

MSaguer:
Oh, I see. Far more complex than the use case I understood reading the issue.
So we need to be able to check current project properties just based on files found in project. May be easy for maven/gradle but harder in some other cases.

pascalgrimaud:
Yes I'm still thinking and don't have a clear solution, excepting doing manually the check in each context.

For example, if the user wants to add liquibase, we could simply check if there is already an existing datasource in properties, or a driver in pom.xml/gradle

pascalgrimaud:
I think the idea is good, but it will be tackle by https://github.com/jhipster/jhipster-lite/issues/1666
So don't hesitate to comment there
So let's close this

PR #1459, opened 2022-04-24
Download: only JHipster Lite project

### LINKED ISSUE #1452: Download: only JHipster Lite project
To protect the download, this API should only be allow for JHipster Lite projects.
There is already this method: isJHipsterLiteProject

Just use it

qmonmert:
Fix #1452


pascalgrimaud:
Conflict because of https://github.com/jhipster/jhipster-lite/pull/1458
Can you take care of it plz @qmonmert ?

qmonmert:
New PR https://github.com/jhipster/jhipster-lite/pull/1461

Issue #1288, opened 2022-04-08
Refactoring: Generator.component.ts


pascalgrimaud:
`Generator.component.ts` can become very big.
It could be nice to split into sub component, like:
- SpringBoot.component.ts
- Angular.component.ts
- React.component.ts
- Vue.component.ts
- etc.

So SpringBoot.component.ts will call only SpringBootService for example

Franceq34:
🔧WIP🔨

pascalgrimaud:
Any progress on this, @Franceq34 ? If it can motivate to achieve this, I can put a bounty :)

Then, if I can have one example with 1 component, I can implement the other components.

Franceq34:
$100 bounty claimed => https://opencollective.com/generator-jhipster/expenses/74251

pascalgrimaud:
@Franceq34 : approved

PR #1480, opened 2022-04-25
Refactor Generator component with AngularGenerator

Franceq34:
- related https://github.com/jhipster/jhipster-lite/issues/1288


src/test/javascript/spec/springboot/primary/Generator.spec.ts
```diff
@@ -36,7 +36,7 @@ const wrap = (wrapperOptions?: Partial<WrapperOptions>) => {
     vueService: stubVueService(),
     ...wrapperOptions,
   };
-  wrapper = mount(GeneratorVue, {
+  wrapper = shallowMount(GeneratorVue, {
     global: {
       provide: {
         angularService,
```
pascalgrimaud, line 42: maybe we don't need any more angularService and all other reference ?

pascalgrimaud (approved):

Issue #1478, opened 2022-04-25
Front-end - Improvements


fdelbrayelle:
Following a little test of [front-end generator](https://jhlite.herokuapp.com/generator), here are some improvements that could be achieve:
- Use a CSS "pointer" custor, so that the user knows he can click the whole button + radio button area
- Focus on the input of `Path (required) :` label

fdelbrayelle:
I mean that when you arrive on the page, your input focus is on this field by default whether it's prefilled or not. What do you think about that?

Franceq34:
Oh ok I see ! 
then I agree, we could do that by adding **autofocus** on the input (https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/autofocus)

done here :
https://github.com/jhipster/jhipster-lite/pull/1482/files

Franceq34:
both ideas were done and merged I think the ticket can be closed

PR #1481, opened 2022-04-25
Cursor pointer on inputs label

Franceq34:
- related https://github.com/jhipster/jhipster-lite/issues/1478


pascalgrimaud:
related to https://css-tricks.com/html-inputs-and-labels-a-love-story/

PR #1483, opened 2022-04-26
Add GraalVM support for JHipster Lite

jdubois:
- fix https://github.com/jhipster/jhipster-lite/issues/2819

This adds GraalVM support to JHipster Lite.

To build the native image:

```
mvn -Pnative clean package spring-boot:build-image -DskipTests
```

To run it:

```
docker run -p 7471:7471 -it docker.io/library/jhlite:0.2.1-SNAPSHOT
```

What still needs to be done:

- Eclipse JGit doesn't support GraalVM, so when the image runs, calling any JGit method will result in a stack trace. Quarkus has a JGit extension at https://github.com/quarkiverse/quarkus-jgit so this is possible to fix.


jdubois:
@pascalgrimaud one important change: this removes Logback as it's not supported by Spring Native.

pascalgrimaud:
wow thanks @jdubois ! I'll try it soon

jdubois:
@pascalgrimaud I added a GitHub Action step to automatically build the image, but it will only push it to GitHub packages if it's merged on the main branch.
This is to limit the size this will take, as you need to pay for GitHub packages (the limit is pretty low). Maybe we should look at hosting this with Docker, JFrog, or Azure Container Registry.

.github/workflows/github-actions.yml
```diff
@@ -96,6 +96,31 @@ jobs:
...
+        run: docker login -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }} ghcr.io
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          check-latest: false
+          cache: 'maven'
+      - name: Build the Docker image
+        run: mvn -Pnative spring-boot:build-image -Dspring-boot.build-image.imageName=${{ env.DOCKER_IMAGE_NAME }}:main -DskipTests
```
pascalgrimaud, line 119: can't we use maven wrapper with `./mvnw` here, instead of installing maven?

src/main/resources/logback-spring.xml
```diff
@@ -1,44 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" ?>
```
pascalgrimaud, line 1: I wonder if it's possible: instead of removing this file, could it be possible to exclude it during packaging ?

```xml
        <excludes>
          <exclude>logback-spring.xml</exclude>
        </excludes>
```

pascalgrimaud:
I built the image locally and indeed it started very very fast!

image

Inside the image, we need to install `git`, as some generation part needs to use git.
Otherwise, we'll have this error:

```
org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.Error: java.lang.NoSuchMethodException: org.eclipse.jgit.internal.JGitText.<init>()
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1082) ~[na:na]
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:963) ~[na:na]
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006) ~[na:na]
	at org.springframework.web.servlet.FrameworkServlet.doPost(FrameworkServlet.java:909) ~[na:na]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:681) ~[na:na]
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883) ~[na:na]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:764) ~[na:na]
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:227) ~[na:na]
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:162) ~[na:na]
	at org.springframework.web.filter.CorsFilter.doFilterInternal(CorsFilter.java:91) ~[na:na]
```

src/main/resources/logback-spring.xml
pascalgrimaud: In fact, forget this comment, no need to keep this file

jdubois:
@pascalgrimaud for Git: in fact you're using JGit, which is a Java library and it is not compatible with GraalVM by default. I'm having a look at fixing this, but this will take time and I will need help.

pascalgrimaud:
Ok @jdubois : let me think if we can avoid using git, I'll investigate too

pascalgrimaud:
I'm converting this PR to draft, until we found a solution for this :-)

jdubois:
What do you think about recoding `tech.jhipster.lite.generator.project.infrastructure.secondary.GitUtils` and instead of using JGit (the Java implementation of Git that causes the issue), using an external command for Git like we do in `tech.jhipster.lite.generator.packagemanager.npm.infrastructure.secondary.NpmLocalRepository` (using the locally-installed version of NPM)?
It's a little less portable, but if we consider that people need to have NPM installed, we can also consider they should have Git installed. And for cloud-based environments made with Docker, both shouldn't be an issue (we'll add them to the Docker image).

pascalgrimaud:
Yes it's a valid solution, I'm fine with this :)

pascalgrimaud:
Thanks @alina-yur for trying this and for your suggestions

atomfrede:
The quarkus extensions looks quite small, it replaces the work queue and user home implementation (which seems to be the problematic parts, regarding the jgit issue tracker). I can check it, if we can just use the same substitutes in jhipster lite to make it work with graal without doing heavy refactoring or replacing jgit completly.

pascalgrimaud:
@atomfrede : after discussing with @DamnClin I think we'll remove jgit. In fact we don't need it. It means we'll have this feature soon ;)

atomfrede:
@pascalgrimaud Awesome!

DamnClin:
After https://github.com/jhipster/jhipster-lite/pull/2469 we'll be left wit h2 jgit usages: 

* In dummy feature -> we should totally rework this one (make it a module, generate different files, etc)
* In init (for a git init), let's replace that by an exec (so that mean you'll need to install git on your servers)

DamnClin:
cc @jdubois, if you deploy the current main on a computer with git installed it won't using JGit anymore, hope that helps

jdubois:
Thanks @DamnClin this is probably the easiest way to fix this!

pascalgrimaud:
I gave a quick try this morning. Here different analysis:

I upgrade spring.native version to `0.12.1`
Then, after building and running the image:

I got:

image



```
➜ docker run -it paketobuildpacks/builder:tiny git --version            
git version 2.17.1
```

But I think this image is used only for building, and git is not in the final image

So I added a new class to replace the JGitGitRepository:

```java
class NoGitRepository implements GitRepository {

  private final Logger log = LoggerFactory.getLogger(NoGitRepository.class);

  @Override
  public void init(JHipsterProjectFolder folder) {
    log.info("Can't launch git init");
  }
}
```

After updating Bean configuration, I can start correctly the image. But there is a problem when using the "init" button:

image

This problem is not related to `git` but to something else.
I didn't have time to investigate more

DamnClin:
For the command part this seems relevant https://github.com/oracle/graal/issues/1185

pascalgrimaud:
Following these links:
- https://docs.spring.io/spring-native/docs/current/reference/htmlsingle/
- https://github.com/spring-projects-experimental/spring-native/issues/1152

 I manage to make it works


To fix the reflection, related to jackson :

```java
@org.springframework.nativex.hint.TypeHint(
  types = { RestJHipsterModuleProperties.class, PersistedProjectHistory.class, ProjectHistory.class, ProjectDTO.class, PersistedProjectAction.class},
  access = { TypeAccess.DECLARED_CONSTRUCTORS, TypeAccess.PUBLIC_METHODS }
)
```

To fix the git configuration :

```java
package tech.jhipster.lite.git.infrastructure.secondary;

import org.slf4j.Logger;
import tech.jhipster.lite.git.domain.GitCommitMessage;
import tech.jhipster.lite.git.domain.GitRepository;
import tech.jhipster.lite.module.domain.properties.JHipsterProjectFolder;

class NoGitRepository implements GitRepository {

  private final Logger log = LoggerFactory.getLogger(NoGitRepository.class);

  @Override
  public void init(JHipsterProjectFolder folder) {
    log.info("Can't launch git init");
  }

  @Override
  public void commitAll(JHipsterProjectFolder folder, GitCommitMessage message) {
    log.info("Can't launch git commit");
  }

  @Override
  public void commitAll(JHipsterProjectFolder folder, String message) {
    log.info("Can't launch git commit all");
  }
}
```

And `GitRepositoryConfiguration` :

```java
@Configuration
class GitRepositoryConfiguration {

  private static final Logger log = LoggerFactory.getLogger(GitRepositoryConfiguration.class);

  @Value("${application.git:true}")
  private boolean applicationGit;

  @Bean
  GitRepository gitRepository(GitChecker checker) {
    if (!applicationGit) {
      log.info("Using No git repository");
      return new NoGitRepository();
    }

    if (checker.hasGit()) {
      log.info("Using command line git repository");
      return new CommandLineGitRepository();
    }

    log.info("Command line not available, using JGit git repository");
    return new JGitGitRepository();
  }
}
```

In `application.properties` :

```properties
application.git=true
```

DamnClin:
I don't understand. If we have to remove the git support then let's just delete the command line implementation. The whole goal of this was to be able to use it on graal, if it doesn't work let's just delete it!

BUT, removing git can't be done just like that, this is buggy since users will still have the commit checkbox on the frontend. I think we should either make git works (probably the easiest one to do) or to something to totally remove git options

pascalgrimaud:
@DamnClin : no, I don't want to remove git, I just tried to make it works, that's all
and as I didn't find a solution for Git in GraalVM image, I disabled it

atomfrede:
I can try the Quarkus like jgit replacement if you haven't done it already.

pascalgrimaud:
@atomfrede : you can give a try, I already did it but it doesn't work

DamnClin:
Since the command line implementation for git manipulation wasn't helping (but still creating some logic duplication) I'm deleting it in https://github.com/jhipster/jhipster-lite/pull/3228

pascalgrimaud:
Closing this, in favor of https://github.com/jhipster/jhipster-lite/pull/4314

pascalgrimaud:
Oh maybe there can be some ideas which can be taken, specially for the build part
cc @atomfrede

atomfrede:
Yes maybe we can (re)use the github actions configuration

PR #1505, opened 2022-04-27
Templating forward port for github codespace

### LINKED ISSUE #1444: Codespaces: forwardPorts and serverPort
The current generated Codespaces does not use the server port, and hard coded 8080 instead

As you did the feature, can you fix this bug plz @matthieulapatate ?

matthieulapatate:
- Fix #1444 

I replaced the forward port file with mustache extension to allow import port when importing the file.

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/setup/codespaces/domain/CodespacesDomainService.java
+++ b/src/main/java/tech/jhipster/lite/generator/setup/codespaces/domain/CodespacesDomainService.java
@@ -21,7 +21,8 @@ public class CodespacesDomainService implements CodespacesService {
   }
 
   private void addConfig(Project project) {
-    projectRepository.add(project, SOURCE, "devcontainer.json", DEVCONTAINER_DEST, "devcontainer.json");
+    project.addConfig("serverPort", 8080);
+    projectRepository.template(project, SOURCE, "devcontainer.json", DEVCONTAINER_DEST, "devcontainer.json");
   }
 
   private void addDockerfile(Project project) {
--- a/src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
+++ b/src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
@@ -23,7 +23,7 @@
     "christian-kohler.npm-intellisense"
   ],
 
-  "forwardPorts": [3001, 4200, 8080, 9000, 18080],
+  "forwardPorts": [{{serverPort}}, 9000],
 
   "remoteUser": "vscode",
   "features": {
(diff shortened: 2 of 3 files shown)


src/main/java/tech/jhipster/lite/generator/setup/codespaces/domain/CodespacesDomainService.java
```diff
@@ -21,7 +21,8 @@ public void init(Project project) {
   }
 
   private void addConfig(Project project) {
-    projectRepository.add(project, SOURCE, "devcontainer.json", DEVCONTAINER_DEST, "devcontainer.json");
+    project.addConfig("serverPort", 7471);
```
pascalgrimaud, line 24: the default port of all Spring Boot application is 8080
7471 is only related to JHipster Lite

src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
```diff
@@ -23,7 +23,7 @@
     "christian-kohler.npm-intellisense"
   ],
 
-  "forwardPorts": [3001, 4200, 8080, 9000, 18080],
+  "forwardPorts": {{serverPort}},
```
pascalgrimaud, line 26: you need to keep 9000 too, as 9000 is used when you launch the front with `npm start`

pascalgrimaud (changes requested):

src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
pascalgrimaud: This file needs to be renamed with mustache template too

pascalgrimaud (changes requested):
Waiting your fix after my review

src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
pascalgrimaud: @matthieulapatate : your initial change was good, we should only have here: {{serverPort}} and 9000
3001, 4200, 18080 need to be removed

pascalgrimaud (changes requested):
the Dockerfile doesn't contain any variable, so no need to be mustache

Only devcontainers.json has variable, so this one need to be mustache

src/main/resources/generator/setup/codespaces/devcontainer.json.mustache
```diff
@@ -23,7 +23,7 @@
     "christian-kohler.npm-intellisense"
   ],
 
-  "forwardPorts": [3001, 4200, 8080, 9000, 18080],
+  "forwardPorts": [3001, 4200, {{serverPort}}, 9000, 18080],
```
pascalgrimaud, line 26:
```suggestion
  "forwardPorts": [{{serverPort}}, 9000],
```

pascalgrimaud (changes requested):
plz, squash all your commits at the end too

pascalgrimaud:
I convert it to draft so:
- once the tests are fixed
- once the commits are squash
- change it to ready for review

pascalgrimaud (approved):

Issue #1547, opened 2022-04-29
Frontend: Move Header in dedicated component


Franceq34:
Generator component is still too big and Header can be moved in a dedicated component

PR #1548, opened 2022-04-29
Move header in dedicated component

Franceq34:
- fix : https://github.com/jhipster/jhipster-lite/issues/1547


seraphinandrieux:
Good :)

Issue #1506, opened 2022-04-27
Create new REST api for generating User and Authority entity for MongoDB


beatfreaker:
Related to #805

PR #1552, opened 2022-04-30
Add user and authority for mongodb

beatfreaker:
Fixes - #1506

@pascalgrimaud  have below questions

Didn't added below two method in `UserJpaRepository` when generated for `MongoDB` as I think `@EntityGraph` is not supported by `spring-data-mongodb` and as a result have not set relation between `UserEntity` and `AuthorityEntity` Lazy so that all the child entites also get loaded.
Is it fine ?

```java
@EntityGraph(attributePaths = "authorities")
Optional<UserEntity> findOneWithAuthoritiesByLogin(String login);

@EntityGraph(attributePaths = "authorities")
Optional<UserEntity> findOneWithAuthoritiesByEmailIgnoreCase(String email);
```

Also for enable auditing i.e `@CreatedBy` & `@LastModifiedBy`, we have to provide implementation of `AuditorAware`
should we generate some dummy implementation ?
[Ref - mongo.auditing](https://docs.spring.io/spring-data/mongodb/docs/current/reference/html/#mongo.auditing)


pascalgrimaud:
The reference should be this repo https://github.com/jhipster/jhipster-sample-app-mongodb

About repository, try to do similar methods than https://github.com/jhipster/jhipster-sample-app-mongodb/blob/main/src/main/java/io/github/jhipster/sample/repository/UserRepository.java

About audit:
- https://github.com/jhipster/jhipster-sample-app-mongodb/blob/main/src/main/java/io/github/jhipster/sample/domain/AbstractAuditingEntity.java
- https://github.com/jhipster/jhipster-sample-app-mongodb/blob/main/src/main/java/io/github/jhipster/sample/config/DatabaseConfiguration.java#L26

So yes, we could initialize audit too, but as I'm not a MongoDB user, I don't know if it should be generated by default. At least for consistency with SQL options, I would say yes

beatfreaker:
@pascalgrimaud 
should we generate [SpringSecurityAuditorAware.java](https://github.com/jhipster/jhipster-sample-app-mongodb/blob/15cec0c97badf75182ea8cc9d96b51c7731d23c0/src/main/java/io/github/jhipster/sample/security/SpringSecurityAuditorAware.java) when user execute either of these two api `/jwt` or `/oauth2` ?

also if we use it in `mongodb` users have to add `security` when they want to use `mongodb`, is it fine ?

src/main/java/tech/jhipster/lite/generator/server/springboot/user/infrastructure/primary/rest/SpringBootUserResource.java
```diff
@@ -62,6 +63,18 @@ public void addUserAndAuthorityEntitiesForMariaDB(@RequestBody ProjectDTO projec
     addUserAndAuthorityEntities(project, MARIADB);
   }
 
+  @Operation(summary = "Add UserEntity, AuthorityEntity and JpaRepository for MongoDB")
+  @PostMapping("/mongodb")
+  @ApiResponse(
+    responseCode = "500",
+    description = "An error occurred while adding UserEntity, AuthorityEntity and JpaRepository for MongoDB"
+  )
+  @GeneratorStep(id = GeneratorAction.USER_AND_AUTHORITY_ENTITIES_MARIADB)
```
seraphinandrieux, line 72:
```suggestion
  @GeneratorStep(id = GeneratorAction.USER_AND_AUTHORITY_ENTITIES_MONGODB)
```

beatfreaker:
@pascalgrimaud I have aligned the repository like https://github.com/jhipster/jhipster-sample-app-mongodb/blob/main/src/main/java/io/github/jhipster/sample/repository/UserRepository.java as you mentioned

and for auditing for now I am always returning `system`, as if we want to extract current logged in user we need security libraries as well, and this will create dependency on security endpoint.

So if user wants to generate only `mongodb` app they can't and have to include security too.

what you think about this ?

```java
@Component
public class SpringSecurityAuditorAware implements AuditorAware<String> {

  @Override
  public Optional<String> getCurrentAuditor() {
    return Optional.of(UserConstants.SYSTEM);
  }
}
```

pascalgrimaud:
Sorry, didn't see the previous comment
Is this SpringSecurityAuditorAware mandatory to have?

beatfreaker:
Nope, it is not mandatory, but if we don't have it `@CreatedBy` won't work automatically, the field annotated with that annotation has to be populated manually.

pascalgrimaud:
After thinking, we could do this:
- check if there is `spring-security-test` dependency, it means there is JWT or OAuth2 security, so we can add SpringSecurityAuditorAware
- otherwise, we won't' generate SpringSecurityAuditorAware and CreatedBy will be empty (or with default value)

Is it possible?
So we'll have 2 use cases to test:
- use mongodb + mongodb user/authority
- use security -> mongodb -> mongodb user/authority

beatfreaker:
@pascalgrimaud getting failure in sonar with code duplication.

And it shows duplication between generated `UserDTO` and `UserEntity` class,
could you please suggest what can be done to fix this issue ?

Screenshot 2022-05-10 at 1 11 45 AM

pascalgrimaud:
I'll have a look tomorrow but for me, primary is out of scope of this. The ticket is related only to secondary (entity part)

beatfreaker:
Yes, I have not made any changes in primary, basically I am generating a new project to test `mongodb` with `oauth2`, and for that project `sonar` analysis is failing with code duplication,
as `UserDTO` is coming from `oauth2` code generation and `UserEntity` is coming from `mongodb` code generation.

pascalgrimaud:
Thanks for the input @beatfreaker 
I'll have a look and try to find a solution

pascalgrimaud:
So in fact, UserDTO will changed, as JWT option and OAuth2 option will be refactorized to respect more the hexagonal architecture
So maybe we can wait the refactor before merging this?

beatfreaker:
Yup make sense, I am okay with that.

src/main/java/tech/jhipster/lite/generator/server/springboot/database/mongodb/domain/MongodbDomainService.java
```diff
@@ -92,8 +95,28 @@ public void addJavaFiles(Project project) {
     project.addDefaultConfig(BASE_NAME);
     String packageNamePath = project.getPackageNamePath().orElse(getPath("com/mycompany/myapp"));
     String mongodbPath = "technical/infrastructure/secondary/mongodb";
+    String securityPath = "security";
+
+    Optional<SecurityType> securityTypeOptional = getAddedSecurityType(project);
```
seraphinandrieux, line 100: To be more readeable, could you extract the two statements into methods ( `addMongoDbWithSecurity` and `addMongoDbWithoutSecurity` for example ...), and use the optional method `ifPresentOrElse()`

src/main/java/tech/jhipster/lite/generator/server/springboot/database/mongodb/domain/MongodbDomainService.java
```diff
@@ -148,4 +171,27 @@ public void addLogger(Project project, String packageName, Level level) {
   private void updateIntegrationTestAnnotation(Project project) {
     springBootCommonService.updateIntegrationTestAnnotation(project, "MongodbTestContainerExtension");
   }
+
+  private Optional<SecurityType> getAddedSecurityType(Project project) {
+    if (isOAuth2DependencyAdded(project)) {
+      return Optional.of(SecurityType.OAUTH2);
+    } else if (isJWTDependencyAdded(project)) {
```
seraphinandrieux, line 178: Not need an `else if` statement when you return an object into your first statement, an` if `is enough

pascalgrimaud:
As discussed here https://github.com/jhipster/jhipster-lite/issues/2482

I don't think we should merge this.
But as you spent a lot of time on this, can I propose you to take the bounty anyway, @beatfreaker ?
Tell me if you agree:
- I'll close this ticket
- you can claim it
- then, we can focus on other stuffs

DamnClin:
We'll also use it as an example for the dummy mongo persistence

beatfreaker:
@pascalgrimaud I am okay closing this ticket, thank you 😊

beatfreaker:
@pascalgrimaud as discussed claiming bounty https://opencollective.com/generator-jhipster/expenses/88036

pascalgrimaud:
Approved!
thanks for your work @beatfreaker and sorry for closing this one

PR #1556, opened 2022-05-01
[Angular] Add Angular Material

### LINKED ISSUE #1464: Angular: improve initial generation
To do:
- remove styled api
- keep 1 single API to init Angular part
- at the same time, add a css framework, so it will help to code the other page

Update the front part related to Angular:

image

qmonmert:
Fix #1464 

- add a css framework, so it will help to code the other page ✅

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularDomainService.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularDomainService.java
@@ -81,31 +81,30 @@ public class AngularDomainService implements AngularService {
   }
 
   public void updateAngularFilesForJwt(Project project) {
-    String oldHtml = "<router-outlet></router-outlet>";
+    String oldHtml = "<!-- jhipster-needle-angular-jwt-login-form -->";
     String newHtml =
       """
-        <div [ngSwitch]="account !== null">
-
-          <form role="form" (ngSubmit)="login()" [formGroup]="loginForm" *ngSwitchCase="false">
-            <div>
-              <label class="username-label" for="username">Login</label>
-              <input type="text" name="username" id="username" formControlName="username" #username />
+      <mat-card>
+        <mat-card-content>
+          <div [ngSwitch]="account !== null">
+            <form class="login-form" (ngSubmit)="login()" [formGroup]="loginForm" *ngSwitchCase="false">
+              <mat-form-field class="form-field-full-width" appearance="fill">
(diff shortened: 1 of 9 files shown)


src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularDomainService.java
```diff
@@ -81,31 +81,32 @@ public void addJwtDependencies(Project project) {
   }
 
   public void updateAngularFilesForJwt(Project project) {
-    String oldHtml = "<router-outlet></router-outlet>";
+    String oldHtml = "<div id=\"footer\">";
```
seraphinandrieux, line 84: Hi,
Little question about the old/new html replacement.
Why don't you had a specific comment to replace when you want to add some template code instead ?
I'm asking because the footer is also added at the end of new html ?
Thanks :)

qmonmert: hi, indeed it will be a good idea to have a specific comment but this comment will be strange when we don't replace it!

seraphinandrieux: I think it depends of comment meaning.
If you have something as `-- jhipster-needle-extra-template` which explains it used by generator to add extra info it could be understood. For sure once the project is done the user can delete it.
However, if for now you can add/replace only once from generator indeed it's not maybe necessary.
As you prefer :)

pascalgrimaud:
I'm starting to review this PR @qmonmert 
as usual, thanks for your patience

qmonmert:
@pascalgrimaud ok good, I think the idea of @seraphinandrieux was good, I updated the PR

src/main/resources/generator/client/angular/angular.json
```diff
@@ -30,7 +30,7 @@
                 "output": "content"
               }
             ],
-            "styles": ["src/main/webapp/styles.css"],
+            "styles": ["./node_modules/@angular/material/prebuilt-themes/indigo-pink.css", "src/main/webapp/styles.css"],
```
pascalgrimaud, line 33: just wonder if we can do this import directly inside styles.css ?

pascalgrimaud (dismissed):

src/main/resources/generator/client/angular/angular.json
qmonmert: yes we can, I update

PR #1564, opened 2022-05-02
Rework assert class

DamnClin:
Mutiple goals here:
* Mark some business related assertions as deprecated for deletion;
* Provide first fluent assertions for some types (more to come later);
* Add some javadoc to those APIs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,6 +142,7 @@ Desktop.ini
 ######################
 # Others
 ######################
+infinitest.filters
 *.class
 *.*~
 *~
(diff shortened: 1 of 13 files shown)


pascalgrimaud:
Missing a branch in unit test here: https://app.codecov.io/gh/jhipster/jhipster-lite/compare/1564/tree/src/main/java/tech/jhipster/lite/error/domain/Assert.java

image

And we should be good to go !

pascalgrimaud (approved):

PR #1625, opened 2022-05-07
Add ProjectDTO details API

### LINKED ISSUE #1453: Extract ProjectDTO from a folder
I think it can be a good idea to have an API which will read the information in a folder, and return the ProjectDTO like this:

```json
{
  "folder": "/tmp/beer",
  "generator-jhipster": {
    "baseName": "beer",
    "projectName": "Beer Project",
    "packageName": "tech.jhipster.beer",
    "serverPort": 8080
  }
}
```

baseName : can be extracted from package.json, pom.xml, gradle file
projectName: can be extracted from package.json or pom.xml
packageName: can be extracted from pom.xml or gradle file

So with this API, it could be used in the front page:
- fill the path
- click on load button
- then the other fields are automatically filled

This 2nd part can be done in another PR

swarajsaaj:
Adding API to read the details required for ProjectDTO
Partially fixes #1453


src/main/java/tech/jhipster/lite/common/domain/FileUtils.java
```diff
@@ -89,6 +91,20 @@ public static void write(String filename, String text, String eol) throws IOExce
     Files.write(getPathOf(filename), transformEndOfLine(text, eol).getBytes());
   }
 
+  public static Optional<String> getValueBetween(String filename, String prefix, String suffix) {
+    Assert.notBlank(FILENAME, filename);
+
+    Optional<String> line = readLine(filename, prefix);
+    Pattern pattern = Pattern.compile(prefix + REGEXP_DOT_PLUS_MIN + suffix, Pattern.DOTALL);
+    if (line.isPresent()) {
+      Matcher matcher = pattern.matcher(line.get());
```
seraphinandrieux, line 100: You can maybe extract the first if statement into a method like Function` <String, Optional<String>> getValueFromLine()` and call it into a map from line like `line.map(this::getValueFromLine).orElse(Optional.empty())`.
It avoids to have an if statement in another one

pascalgrimaud:
@swarajsaaj : sorry, can you solve the conflict plz?

swarajsaaj:
> @swarajsaaj : sorry, can you solve the conflict plz?

Done @pascalgrimaud

pascalgrimaud:
Thanks for your work here @swarajsaaj :-)))

Issue #1592, opened 2022-05-05
Frontend: Add an AlertBus


Franceq34:
**Statement :**
We have multiple messages that transit in the application.
When calling "add Angular" for example : 

![Capture d’écran de 2022-05-05 11-00-22](https://user-images.githubusercontent.com/24738688/166892560-2ff37ec5-c867-4442-9744-c4fe98c0f42b.png)

You can notice we are calling the Logger, the toaster service, and we can imagine other options in the future.

The idea of the ticket is to create an AlertBus in the secondary that would **emit** events such as success, error or information events.

This Bus will be subscribed by the Logger and and the ToastService.

Doing so, we will also be removing this primary dependency in secondary : 

![Capture d’écran de 2022-05-05 11-05-28](https://user-images.githubusercontent.com/24738688/166893274-8afe8d51-4921-4884-b286-ceb3d29ec94e.png)

What do you think about it ? @nonomoho @seraphinandrieux

pascalgrimaud:
Ok then, looks good so let's do it

Franceq34:
WIP :)

PR #1690, opened 2022-05-10
Frontend: Add an AlertBus

Franceq34:
- fix https://github.com/jhipster/jhipster-lite/issues/1592


src/main/webapp/app/common/domain/alert/AlertBus.ts
```diff
@@ -0,0 +1,10 @@
+import { Alerted } from '@/common/domain/alert/Alerted';
+import { Unsubscribe } from '@/common/domain/alert/Unsubscribe';
+import { AlertMessage } from '@/common/domain/alert/AlertMessage';
+
+export interface AlertBus {
```
DamnClin, line 5: Defining your bus like this will force you to break the hexagonal architecture since this port will be both primary and secondary. You should check with @sstan001 :)

sstan001: 👋

pascalgrimaud:
Strange no CI, I'm closing and reopening this to try to restart the CI

src/main/webapp/app/common/domain/alert/AlertBus.ts
Franceq34: are my changes ok for both of you ?

DamnClin: Ok for me, an interesting step would be to create a plugin for injections: 

in main.ts
```
const alertsInjections: AlertsInjections = //...

app.use(alertsProvider, alertsInjections)
```

and somewhere:
```
export interface AlertsInjections {
  // ...
}

export const alertsProvider = (app, injects: AlertsInjections) => app.provide(...)
```

This way it may be easier to see matching injections

pascalgrimaud: ok, maybe it can be done in another PR as improvements

PR #1700, opened 2022-05-11
Add remote url to projectDTO schema

matthieuRioual:
Add an optional variable into project DTO and getter for that variable into project class


src/main/java/tech/jhipster/lite/generator/project/domain/Project.java
```diff
@@ -126,6 +125,10 @@ public Optional<String> getPackageNamePath() {
     return getPackageName().map(packageName -> packageName.replace(".", "/"));
   }
 
+  public Optional<String> getRemotePath() {
+    return getStringConfig(REMOTE_URL);
```
pascalgrimaud, line 129: The config map is used by mustache to do some replace in the template.

Here, I think we should add a new optional field, at the same level with path, instead

pascalgrimaud (changes requested):

pascalgrimaud (approved):

PR #1702, opened 2022-05-11
Simple collections utility

DamnClin:
The main goal here is to init a Collection manipulation utility class
for the project. It's always usefull to have this kind of null safe
utility


pascalgrimaud:
nice to have, thanks @DamnClin

PR #1731, opened 2022-05-13
Rework add file repository

DamnClin:
This is a first refactoring step to avoid ambiguities in file
parameters. Next step (if you are fine with this first one)
would be to do the same for the other methods in this repository


pascalgrimaud:
This is really smart and will help a lot our contributors.
Thanks so much @DamnClin 

I have few comments, related to prettier, not to this PR, because auto format with prettier is broken but it can be fixed in another PR

pascalgrimaud (approved):

PR #1803, opened 2022-05-20
Rework java base module

DamnClin:
Rework Java base module to use the new module mechanism


src/main/java/tech/jhipster/lite/generator/module/domain/JHipsterProjectBaseName.java
```diff
@@ -0,0 +1,28 @@
+package tech.jhipster.lite.generator.module.domain;
+
+import org.apache.commons.lang3.StringUtils;
+
+//TODO: format
```
DamnClin, line 5: Here I want to enfore some kind of formatting (avoid spaces, for example) but I'm not sure of the kind of format is suitable for project names, any ideas?

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class JavaBaseModuleFactory {
+
+  public JHipsterModule buildModule(JavaBaseModuleProperties properties) {
+    Assert.notNull("properties", properties);
+
+    String packagePath = properties.basePackage().path();
+    String baseClassName = properties.projectBaseName().capitalized();
+
+    JHipsterDestination testDestination = toJavaTestSource().append(packagePath);
```
DamnClin, line 18: I'm not happy with this `.append(...)` API, any better idea?

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
```diff
@@ -0,0 +1,65 @@
...
+    JHipsterDestination mainDestination = toJavaMainSource().append(packagePath);
+
+    return moduleForProject(properties.project())
```
DamnClin, line 21: Using the default formatting is not that great but I think it's OK, any advice on this?

src/main/java/tech/jhipster/lite/generator/module/domain/JHipsterProjectBaseName.java
pascalgrimaud: here the rules from generator-jhipster: https://github.com/jhipster/generator-jhipster/blob/main/generators/generator-base.js#L2112-L2120

So I think we can use this regexp `^[a-zA-Z0-9]+$` (without underscore -> no need to apply different check for microservice)

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
pascalgrimaud: we can keep like this and improve later if needed
I'm fine with `append` here :)

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
pascalgrimaud: I agree the default formatting makes it a little bit hard to read, it would be better with // prettier-ignore or // @formatter:off

Anyway, what I can say, it's more fluent and easy to write this kind of code, rather than playing with projectRepository.template / projectRepository.add etc
So it's already a big improvement, don't worry :)

The important thing is to migrate 1 or 2 modules first, test the limitation of the API, try to improve it, before migrating all other modules.
Anyway, the other modules can be migrated by their respective lead too

src/main/java/tech/jhipster/lite/generator/module/domain/JHipsterProjectBaseName.java
DamnClin: Ok, I chose to fail for bad name, another option would have been to reformat but this doesn't sound like a good idea

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
DamnClin: I don't really like it because it sounds like it mutating the object but since I can't come with a better name let's stick to this one till we find better

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
DamnClin: Did a manual format here. Looks better!

pascalgrimaud: Indeed! With manual format, it's far better

pascalgrimaud:
The CI failed because the module java base didn't manage to find the template:

image

To reproduce:
- compile to jar
- move the jar to another folder than target folder (so it won't find template)
- launch it with java -jar *.jar
- use: init, maven, java base

DamnClin:
> The CI failed because the module java base didn't manage to find the template:
> 
> image
> 
> To reproduce:
> 
>     * compile to jar
> 
>     * move the jar to another folder than target folder (so it won't find template)
> 
>     * launch it with java -jar *.jar
> 
>     * use: init, maven, java base

Oh great, thx! Just a path handling thing so :D

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
```diff
@@ -0,0 +1,67 @@
...
+  private JHipsterDestination collectionsDestination(String className, JHipsterDestination mainDestination) {
+    return mainDestination.append("common/domain").append(className + "Collections.java");
+  }
+
+  private JHipsterDestination collectionsTestDestination(String className, JHipsterDestination testDestination) {
+    return testDestination.append("common/domain").append(className + "CollectionsTest.java");
+  }
+
+  private JHipsterSource source() {
+    return from("server/javatool/base");
```
pascalgrimaud, line 65: I'd extract this to a local constant, so for all other context, there will be a SOURCE

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
```diff
@@ -0,0 +1,67 @@
...
+    JHipsterDestination mainDestination = toSrcMainJava().append(packagePath);
+
+    //@formatter:off
+    return moduleForProject(properties.project())
+      .context()
+        .packageName(properties.basePackage())
+        .put("collectionClass", baseClassName)
+        .and()
+      .files()
+        .batch(source(), mainDestination.append("error/domain"))
```
pascalgrimaud, line 28: Extract error/domain to a constant, so it will be easier to change it

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
```diff
@@ -0,0 +1,67 @@
...
+          .add("NotAfterTimeException.java")
+          .add("NotBeforeTimeException.java")
+          .add("NullElementInCollectionException.java")
+          .add("NumberValueTooHighException.java")
+          .add("NumberValueTooLowException.java")
+          .add("StringTooLongException.java")
+          .add("StringTooShortException.java")
+          .add("TooManyElementsException.java")
+          .and()
+        .batch(source(), testDestination.append("error/domain"))
```
pascalgrimaud, line 41: Use const for error/domain

src/test/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+      .project(FileUtils.tmpDirForTest())
+      .basePackage("com.jhipster.test")
+      .projectBaseName("myapp")
+      .build();
+
+    JHipsterModule module = factory.buildModule(properties);
+
+    assertThatModule(module)
+      .hasFiles(
+        "src/main/java/com/jhipster/test/error/domain/Assert.java",
```
pascalgrimaud, line 28: We can reduce the long size src main java ... using a local variable I think

pascalgrimaud (approved):
Some minor comments, after that, it can be merged

src/main/java/tech/jhipster/lite/generator/server/javatool/base/domain/JavaBaseModuleFactory.java
DamnClin: Why a constant instead of a method?

pascalgrimaud: Sorry, I'm not clear enough. My idea is to extract only 'server/javatool/base' to a constant and keep the method, and do similar thing to all module.
So for a new contributor, he/she will know quickly where the templates are, by looking at the beginning of the class, with SOURCE or SOURCE_TEMPLATE for example

It's just an idea, tell me what you think?

DamnClin: it's a first step, I think a JHipsterModuleFactory will appear in some modules refactoring and this probably will be enforced here, we'll see

pascalgrimaud (approved):

PR #1805, opened 2022-05-21
Webflux / Add actuator

Bolo89:
- fix https://github.com/jhipster/jhipster-lite/issues/1395


pascalgrimaud (approved):

pascalgrimaud:
For Spring Boot MVC, it's a separated feature: https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/infrastructure/primary/rest/SpringBootMvcResource.java#L51

We should keep consistency here:
- MVC should add actuator by default, like here
- change this PR to add a new API

What do you prefer?

Bolo89:
> For Spring Boot MVC, it's a separated feature: https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/infrastructure/primary/rest/SpringBootMvcResource.java#L51
> 
> We should keep consistency here:
> 
> * MVC should add actuator by default, like here
> * change this PR to add a new API
> 
> What do you prefer?

:/ I will separate like mvc, 1 feature = 1 endpoint.

Issue #1669, opened 2022-05-09
Allow user to create Merge request on github repository


matthieuRioual:
The idea would be to allow authenticated user to be able to create merge request on a repository.

The authentification could be done like on jhipster-online with Oauth2  and store token ever in a DB or in the storage.
For creating merge request, we could use Github API for java.

matthieulapatate:
Just wondering if i should create a new context or simply use init one for instance.

pascalgrimaud:
It's for managing project, so it's in Project context:
- project/domain -> something like Cloud
- project/infra/sec -> CloudRepository

I don't have better name, need to think more, but it could be a start

PR #1820, opened 2022-05-23
Cloud context: clone and pull request

matthieuRioual:
Related to #1669 

Following #1719

Create a pull request function that takes code in the user given folder and create a pull request on the repository given in the form. Have to think about how to test this function since every test are supposed to create a PR which is not optimal. Currently looking to mock github server that's why i let this PR as draft for instance.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "java.configuration.updateBuildConfiguration": "automatic"
-}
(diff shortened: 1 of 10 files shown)


matthieuRioual:
@pascalgrimaud When you will have time can you check the GitUtilsTest on this branch and give me some clues about shouldCreatePR, shouldPush, shouldConnect tests. These are related to CloudRepositoryTest used in shouldCreatePullRequest test

pascalgrimaud:
as this one was not finished in time (before the migration to modules), it would be better to start from scratch, using the module system, as currently, it can be usable
so closing this one

PR #1821, opened 2022-05-23
Cucumber integration in JHLite

DamnClin:
Two goals:
* Provide a cucumber integration to use in JHLite;
* Prepare a dedicated module creation.


pascalgrimaud:
So nice to have this.
Now I can start to learn Cucumber :-)

Issue #1787, opened 2022-05-19
Handle java dependencies in the new module mechanism


DamnClin:
We started the modules mechanism (https://github.com/jhipster/jhipster-lite/issues/1768) with just context and files, we now need to handle java dependencies, here's an API idea: 

```
JHipsterDependency optionalTestPom = dependency()
    .groupId("group.id")
    .artifactId("artifact-id")
    .versionSlug("version-slug")
    .optional()
    .scope(DependencyScope.TEST)
    .type(DependencyType.POM)
    .build()

moduleForProject(new JHipsterProjectFolder(FileUtils.tmpDirForTest()))
.javaDependencies()
  .add(optionalTestPom)
  .add(groupId("org.springframework.boot"), artifactId("spring-boot"))
  .add(groupId("org.springframework.cloud"), artifactId("spring-cloud-dependencies"), versionSlug("spring-cloud"))
  .and()
.build()
```

This give us 3 ways of adding a dependency:

* A way with all available options (and a builder);
* A way to add dependencies without versions (just group id and artifact id);
* A way to add group id, artifact id and version slug.

The version slug is used to get the version from `src/main/resources/generator/dependencies/pom.xml` to keep the current update behavior but reduce the complexity for module creators

pascalgrimaud:
adding a bounty on this ticket:
- it's a lot of work
- it will manage add / update dependencies, to this part is awesome

DamnClin:
Bounty claim: https://opencollective.com/generator-jhipster/expenses/79245

pascalgrimaud:
@DamnClin : approved

PR #1841, opened 2022-05-24
Init: use cucumber for tests

pascalgrimaud:


DamnClin:
What do you think, the usage is easy enough? In this case you can delete your Rest controller IT and your domain UnitTest I think, you'll get the same coverage but more value since it's actually testing stuff

pascalgrimaud:
I already deleted the InitResourceIT

About the Unit test in Domain and Application Service, I'm working on another PR to use the new system with modules, so I'll delete these files in another PR. Then, we'll have a clean "init" context :-)

DamnClin:
Careful you don't need https://github.com/jhipster/jhipster-lite/issues/1787 to make this happen (the work on this is at step 0 :D: I have a ticket)

pascalgrimaud:
ahaha yes, that's why I choose a context which doesn't need Java dependencies :)

Issue #1873, opened 2022-05-28
Add ComponentTest annotation in JavaBase


DamnClin:
We have UnitTest annotation for... unit test but we need a ComponentTest annotation to mark... component tests

PR #1874, opened 2022-05-28
Add component test annotation

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/1873


pascalgrimaud:
Can you update the test at https://github.com/jhipster/jhipster-lite/blob/main/src/test/features/java-base.feature#L23 too plz?

DamnClin:
> Can you update the test at https://github.com/jhipster/jhipster-lite/blob/main/src/test/features/java-base.feature#L23 too plz?

Yes but I think this is a bad idea to test all files in component test D. I was thinking to remove most of them and just to check one file since checking all files is unit test responsability

pascalgrimaud:
Indeed, it's an acceptable point of view :-)

Issue #1699, opened 2022-05-11
Feature: Cucumber


pascalgrimaud:
The goal is to add Cucumber support: https://cucumber.io/

Some links which can help:

- https://blog.ippon.fr/2020/08/31/cucumber-et-resttemplate/
- https://gitlab.com/nerp-ippon/nerp/-/tree/main/webservices/src/test/java/com/ippon/nerp/cucumber

DamnClin:
I have some improved version of this, I'll take it :)

DamnClin:
Just a little update: we have done the Cucumber integration with fluent assertion API on a dedicated project. We'll take some time to battle test it on the project and we'll then work on the templating

pascalgrimaud:
Good news! Thanks a lot @DamnClin

DamnClin:
Normally https://github.com/jhipster/jhipster-lite/pull/1859 should be he last prerequisite for that, will try to do it when merged

DamnClin:
Hum, I was wrong... Need to add a new annotation in the hexagonal architecture this have other prerequisites... we'll be back on this one after

pascalgrimaud:
Don't forget to claim the bounty @DamnClin and put the link of opencollective here, in comments :)

DamnClin:
https://opencollective.com/generator-jhipster/expenses/79244

pascalgrimaud:
@DamnClin : approved

PR #1875, opened 2022-05-28
Cucumber module

DamnClin:
Module to add cucumber to a project

Close https://github.com/jhipster/jhipster-lite/issues/1699

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/domain/CucumberModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/domain/CucumberModuleFactory.java
@@ -1,0 +1,108 @@
+package tech.jhipster.lite.generator.server.springboot.cucumber.domain;
+
+import static tech.jhipster.lite.generator.module.domain.JHipsterModule.*;
+
+import tech.jhipster.lite.error.domain.Assert;
+import tech.jhipster.lite.generator.module.domain.JHipsterModule;
+import tech.jhipster.lite.generator.module.domain.JHipsterSource;
+import tech.jhipster.lite.generator.module.domain.javadependency.JavaDependency;
+import tech.jhipster.lite.generator.module.domain.javadependency.JavaDependencyScope;
+
+public class CucumberModuleFactory {
+
+  private static final String CUCUMBER_GROUP_ID = "io.cucumber";
+  private static final String CUCUMBER_VERSION = "cucumber.version";
+
+  public JHipsterModule buildModule(CucumberModuleProperties properties) {
+    Assert.notNull("properties", properties);
+
+    String packagePath = properties.basePackage().path();
+    String applicationName = properties.projectBaseName().capitalized();
+    JHipsterSource source = from("server/springboot/cucumber");
+
+    //@formatter:off
+    return moduleForProject(properties.project())
+      .context()
+        .packageName(properties.basePackage())
+        .put("applicationName", applicationName)
+        .and()
+      .files()
+        .batch(source, toSrcTestJava().append(packagePath).append("cucumber"))
+          .add("AsyncElementAsserter.java")
+          .add("AsyncHeaderAsserter.java")
+          .add("AsyncResponseAsserter.java")
+          .add("Awaiter.java")
+          .add("CucumberAssertions.java")
+          .add("CucumberConfiguration.java")
+          .add("CucumberJson.java")
+          .add("CucumberTest.java")
+          .add("CucumberTestContext.java")
+          .add("CucumberTestContextUnitTest.java")
+          .add("ElementAsserter.java")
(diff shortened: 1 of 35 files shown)


DamnClin:
@pascalgrimaud where do you think the project generation test should be added for this one?

DamnClin:
BTW this needs https://github.com/jhipster/jhipster-lite/pull/1874 :D

pascalgrimaud:
> @pascalgrimaud where do you think the project generation test should be added for this one?

Don't understand well the question. Shouldn't we keep the same structure of folder than in JH Lite ?

DamnClin:
> > @pascalgrimaud where do you think the project generation test should be added for this one?
> 
> Don't understand well the question. Shouldn't we keep the same structure of folder than in JH Lite ?

No, I mean I added it in `fullapp`, is this correct?

pascalgrimaud:
@DamnClin : oh yes, looks good to add this to fullapp

src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/infrastructure/primary/CucumberResource.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public CucumberResource(CucumberApplicationService cucumber) {
+    this.cucumber = cucumber;
+  }
+
+  @PostMapping
+  @GeneratorStep(id = GeneratorAction.SPRINGBOOT_CUCUMBER)
+  @Operation(summary = "Add cucumber integration to project")
+  @ApiResponse(responseCode = "500", description = "An error occurred while adding cucumber elements")
+  public void addJavaBase(@RequestBody ProjectDTO projectDTO) {
```
pascalgrimaud, line 32: Wrong method name :-p

src/test/java/tech/jhipster/lite/generator/ProjectsSteps.java
```diff
@@ -20,6 +21,11 @@ public class ProjectsSteps {
 
   private static String lastProjectFolder;
 
+  @Given("I have a default project with a maven file")
+  public void createDefaultProjectWithMavenFile() {
+    // TODO
```
pascalgrimaud, line 26: Still todo?

pascalgrimaud (approved):

src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/domain/CucumberModuleFactory.java
```diff
@@ -0,0 +1,108 @@
...
+        .add(source.template("cucumber.md"), to("documentation/cucumber.md"))
+        .add(source.file("gitkeep"), to("src/test/features/.gitkeep"))
+        .and()
+      .javaDependencies()
+        .add(cucumberJunitDependency())
+        .add(cucumberJavaDependency())
+        .add(cucumberSpringDependency())
+        .add(junitVintageDependency())
+        .add(testNgDependency())
+        .add(awaitilityDepencency())
```
pascalgrimaud, line 59: it looks like it's adding 3 times the property:

image

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/infrastructure/primary/CucumberResource.java
DamnClin: Copy pasta spotted :D

src/test/java/tech/jhipster/lite/generator/ProjectsSteps.java
DamnClin: No, changed my mind on this but forgot to remove...

src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/domain/CucumberModuleFactory.java
DamnClin: Oh, good catch, forgot about that case, will fix, thanks

pascalgrimaud: it's minor because prettier can do the job, but there are some additional useless spaces, which can be removed too. Can be done in a separated PR as improvement :)

DamnClin: I think a dedicated for those useless spaces + useless line break in some cases after a removal can be a good idea since it may be more challenging than it looks and value is not that great (I gave it a 3mn try but failed...). The other points are fixed :D

pascalgrimaud (approved):

Issue #1858, opened 2022-05-26
Autofill of project details from path


swarajsaaj:
Following #1453

- On entering the path in project details
- Prefill the other details like projectName, packageName, serverPort, baseName if project already exists on the path

Will provide a PR soon.

pascalgrimaud:
@swarajsaaj : adding a bounty as it's a very nice feature :)

PR #1876, opened 2022-05-28
Feature/autofill projectdetails

swarajsaaj:
Fix https://github.com/jhipster/jhipster-lite/issues/1858


src/main/webapp/app/springboot/domain/ProjectService.ts
```diff
@@ -1,5 +1,6 @@
 import { Project } from '@/springboot/domain/Project';
 import { DocumentFile } from '@/common/domain/DocumentFile';
+import { RestProject } from '../secondary/RestProject';
```
pascalgrimaud, line 3: as we are in domain, we should not import a secondary

src/main/webapp/app/springboot/domain/ProjectService.ts
```diff
@@ -12,4 +13,5 @@ export interface ProjectService {
   download(project: Project): Promise<DocumentFile>;
   addCodespacesSetup(project: Project): Promise<void>;
   addGitpodSetup(project: Project): Promise<void>;
+  getProjectDetails(folder: string): Promise<RestProject>;
```
pascalgrimaud, line 16: maybe return a Promise<Project> instead ?

pascalgrimaud (changes requested):

src/main/webapp/app/springboot/domain/ProjectService.ts
swarajsaaj: Do you mean `Promise<void>` instead of `Promise<RestProject>?`

pascalgrimaud: Yes, or Promise<Project>

swarajsaaj: Sure, as return type generic is not required for the feature, I have made it `Promise<void>` for generic required in promise

src/main/webapp/app/springboot/domain/ProjectService.ts
swarajsaaj: Fixed in [727b697](https://github.com/jhipster/jhipster-lite/pull/1876/commits/727b6976e67a31641da8a03e8c794aa1831c755b)

pascalgrimaud (approved):

swarajsaaj:
bounty claimed at https://opencollective.com/generator-jhipster/expenses/81769

pascalgrimaud:
@swarajsaaj : approved :)

PR #1894, opened 2022-05-30
Update ci to filter tests to execute

### LINKED ISSUE #1649: CI: start and test the generated applications
If it's only backend:
- test health check

If there is a front:
- maybe cypress ?

seraphinandrieux:
- fix #1649


.github/workflows/github-actions.yml
```diff
@@ -101,6 +101,10 @@ jobs:
   cypress:
     needs: [tests-linux]
     runs-on: ubuntu-latest
+    on:
+      push:
+        paths:
+          - 'src/main/webapp/**'
```
pascalgrimaud, line 107: Maybe package.json too, in case of some dependencies are updated

seraphinandrieux: okidoki (didn't expect to see you so early)

PR #1896, opened 2022-05-30
Common springdoc mvc webflux

Bolo89:
- fix: https://github.com/jhipster/jhipster-lite/issues/1806


src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/MavenDomainService.java
```diff
@@ -285,4 +286,11 @@ public Optional<String> getName(String folder) {
 
     return FileUtils.getValueBetween(getPath(folder, POM_XML), NAME_BEGIN, NAME_END);
   }
+
+  @Override
+  public boolean containsDependency(Project project, String dependencyName) throws IOException {
```
DamnClin, line 291: You can rely on https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/module/domain/javadependency/JavaDependenciesCurrentVersionsRepository.java for this I think

Bolo89: This is what I wanted to use at first, but spring starter dependencies don't have version in the pom. And get the version of spring boot does not mean that webflux is in the pom (my case).
```xml
<dependency>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-webflux</artifactId>
</dependency>
```

DamnClin: Oh yes, I totally gave you the wrong link.... I mean `ProjectJavaDependenciesRepository` and you can then use `dependency(id)` to get an optional dependency that should do what you want.

The main point of this is you ill get the gradle support with gradle implementation in this

Bolo89: ok thanks, done

src/test/java/tech/jhipster/lite/generator/module/domain/JHipsterModulesDomainServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+import tech.jhipster.lite.generator.module.domain.javadependency.GroupId;
+import tech.jhipster.lite.generator.module.domain.javadependency.JavaDependencies;
+import tech.jhipster.lite.generator.module.domain.javadependency.JavaDependenciesVersions;
+import tech.jhipster.lite.generator.module.domain.javadependency.JavaDependency;
+import tech.jhipster.lite.generator.module.domain.javadependency.ProjectJavaDependencies;
+import tech.jhipster.lite.generator.module.domain.javadependency.ProjectJavaDependenciesRepository;
+
+@UnitTest
+@ExtendWith(MockitoExtension.class)
+class JHipsterModulesDomainServiceTest {
```
DamnClin, line 29: I may be wrong but this looks a lot like a test after as it's really related to implementation details (sorry if I'm wrong here). I don't like this kind of test since: 

* They are complicated to do;
* They will break as soon as you change an implementation detail.

I think this involve some flow in the design:
* I don't think `getDependency` belong to `JHipsterModules`;
* I think you probably need a `SpringDocDependenciesRepository` in generator/server/springboot/apidocumentation/springdoc/domain with `springDocType()` method that will get an enum telling the dependency type

This way you will cut the link between those two packages and will be able to easily manage the dependencies creation in the enum.

Not sure if all that make sens, I'm available if you want to talk about that further on

Bolo89: I will remove the added test in JhipsterModules, but, for me (my opinion only), if the method to test has if / else, try / catch... ( this can evolve in the time...) it's better to test directly the method (and to be honest, I can't leave the class that I write, before testing it, that eliminate my main development errors (mapping, wrong params,... ). 

The method getDependency does not belong to JhipsterModules, but Jhipster module use it, and the "project" is a module. For the moment, only sprindoc need this, so I will remove it for JhipsterModules, I can understand that it's not the better place.

And for SpringDocDependenciesRepository, it's for me a little bit " lourd " for only one method, but it's maybe better, I will see.

PR #1662, opened 2022-05-08
Add Gitpod generation

MSaguer:
Add gitpod configuration for generated project
New setup tool card in front part

Close https://github.com/jhipster/jhipster-lite/issues/1628

PR #1922, opened 2022-06-01
Add gitpod Readme file

MSaguer:
Following #1662, add missing README file


src/main/java/tech/jhipster/lite/generator/setup/gitpod/README.md
```diff
@@ -0,0 +1,12 @@
...
+## Description
+
+This context is used to add [Gitpod](https://gitpod.io) to an existing project:
+
+- `.gitpod.yml` to configure the web IDE
+- `.gitpod.Dockerfile` to define the custom Docker container used to run the web IDE
+
+## Maintainers
+
+- [Pascal Grimaud](https://github.com/pascalgrimaud)
```
pascalgrimaud, line 12: Can you put your name instead of mine? :-D
As I don't use gitpod, I won't be able to maintain it

MSaguer: I'm not an experimented Gitpod user but I would be happy to work on this 👍

pascalgrimaud (approved):

Issue #1868, opened 2022-05-28
Add Spring properties management in modules


DamnClin:
Add the possibility to manage spring properties in modules. API ideas:

```
moduleForProject(...)
    .springProperties()
      .set("springdoc.swagger-ui.operationsSorter", "alpha")
      .and()
    .context()
    // ...
```

Totally handling this one is looks challenging! We need to handle: 

* .properties and .yml files
* Existing properties update
* New properties
* Decent new key position in files

pascalgrimaud:
At the beginning, I chose properties as it's simpler to manage, it's easy to find and replace the needle

About yml, we could use SnakeYAML, but maybe it's a lot of effort. And adding support for both type of file: properties and yml will be overkill.

So for now, I'd follow https://start.spring.io/ as it generates application.properties

DamnClin:
Ok, I'll try to make a PR for that in the next days

PR #1953, opened 2022-06-05
Handle spring properties in modules

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/1868

I chose to simply append new properties at the end of the file (so this works without needle), tell me if that ok (ofc it will replace existing properties)


antarus:
For a first version, it looks good.

PR #2026, opened 2022-06-10
Front: add dummy feature button

### LINKED ISSUE #1637: Front: add dummy feature button

marioneyraud:
fix #1637


pascalgrimaud:
Thanks so much for your contribution @marioneyraud !
I'll review it soon ;)

pascalgrimaud:
There is a problem:
- the value used in front part doesn't match with the value in GeneratorAction, so when using Dummy feature, the button doesn't change

src/main/java/tech/jhipster/lite/generator/project/domain/GeneratorAction.java
```diff
@@ -81,7 +81,7 @@ private GeneratorAction() {}
 
   public static final String AOP_LOGGING = "aop-logging";
 
-  public static final String DUMMY_CONTEXT = "dummy-context";
+  public static final String DUMMY_CONTEXT = "dummy-feature";
```
pascalgrimaud, line 84: I think the correct value is the one in RestServiceId.ts so, it should be `springboot-dummy-feature` instead
At the same time, can you rename DUMMY_CONTEXT to SPRINGBOOT_DUMMY_FEATURE plz ?
- in this file
- in DummyResource too

pascalgrimaud (changes requested):

pascalgrimaud (approved):

Issue #2003, opened 2022-06-07
Expose new APIs for modules


DamnClin:
For each module we should expose: 

* `GET /api/modules/{module-slug}` to get a `RestJHipsterModulePropertiesDefinition` for this module
* `POST /api/modules/{module-slug}/apply` to apply module updates

For all those endpoints we'll have to build the OpenApi DOC manually since they are dynamic.

Issue #1565, opened 2022-05-03
Frontend : UI/UX improvements


Franceq34:
The goal of this issue is to track the work done on UI/UX and have a better idea of what could look like the final Jhipster generator app. Feel free to **submit any idea in this thread.** 📝 If you feel creative don't hesitate to design some mock-ups with pencil/paper or any tools you are comfortable with. ([Marvel](https://marvelapp.com/features/prototyping), [Miro](https://miro.com/fr/), [Adobe XD](https://www.adobe.com/fr/products/xd.html), [Figma](https://www.figma.com/) etc ...)

### Functionalities needed 🛠

**SpringBoot generator :**

- Project configuration (path, basename, project name, package and port)
- Generator commands (cf « [Generate your project](https://github.com/jhipster/jhipster-lite#generate-your-project) » → or refer to [generator swagger](https://jhlite.herokuapp.com/swagger-ui/index.html)) (Here is the biggest challenge of the UX, as there are tens of APIs, we need to determine how to display them and filter smartly)
- Download the generated project
- Swagger access
- "About" section (JHipster, Github, Twitter, Opencollective ...) 

**Account management (cloud version) :**
- Registration
- Forgot password 

We won’t focus now on **Quarkus** generator or **Micronaut** generator, as this is a completely different context and there’s not any work began on it, but we can keep in mind that this could be added someday.

### Users habits 👤

We can imagine 3 types of user needs using JHLite.

- **The designer** : he **generates his project with the JHLite** small bricks. He initialized his project and added his favorite Springboot options. His project has its lifecycle and the user c**an come back at any moment** to add bricks (ex: adding a new database for example ...)
- **The builder** : this user **has already his project generated with or without JHLite** (maybe JHipster). He wants to add a few bricks.
- **The newbie** : this user could be **afraid with the modular aspect of JHlite**, he is a beginner or could be used to JHipster legacy. He will use an interface "Quickstart" with some API bundles -> he wants to generate his application easily with basic options.

pascalgrimaud:
cc @assimbly as you have probably some suggestion

pascalgrimaud:
Here the current version (v0.3.0):

![image](https://user-images.githubusercontent.com/9156882/166646768-e8167302-d9b9-411c-ab20-c92eaf5583dd.png)

skin27:
Looks good, I will give it a try later this week to give feedback in detail.

Franceq34:
Here is what I got using Figma.

[...]

Issue #1666, opened 2022-05-09
Ease new module creation by structuring modules as a tree


DamnClin:
Most of current modules have a parent -> child relationship (with some modules being root). The idea would be to enforce that with an interface to define a new module, here's a very incomplete idea:

```java
public interface JHipsterModule {
  ModuleKey key();
  ModuleKey parent();
}

public interface ModuleKey {
  String get();
}

public class RootKey implements ModuleKey {
  public String get() {
    return "root";
  }
}
```

This enforced parent -> child relationship will allow us to define a dependencies tree between modules those allowing: 
* Automatic HATEOAS API generation;
* Generic (no new code for a new module) JHipster Lite front end part using the HATEOAS API (we'll need some form definition mechanism for that).

I think this can be a great facilitator to create new modules!

pascalgrimaud:
It should not be mandatory @antarus 

Supposing I created a new project, using my personal starter, I'd like to add PostgreSQL, without the requirement to use init, maven, java base.

That's why I think building a tree can help but should not be mandatory. Each developer should know what he/she adds to the project when using each feature of JHLite

antarus:
@pascalgrimaud I completely agree that it shouldn't be mandatory. 
But jhipster-lite needs to be clearer on what a module needs and maybe what it will add/modify.
For example, I had to switch to debug mode to see the error and determine the files that were missing to add the oauth2. Then I had to look for the modules which added its files.

To be studied, but I think that with the new module system, we can maintain a list of files that each module adds / modifies. This could provide solutions to errors related to missing files for example.

pascalgrimaud:
I'm adding a bounty on this as it's important to ease the use of module

DamnClin:
I think this is covered by https://github.com/jhipster/jhipster-lite/pull/3022

hdurix:
I also think so. We can close this issue. Don't forget the bounty @DamnClin.

DamnClin:
Claimed https://opencollective.com/generator-jhipster/expenses/91032

PR #2036, opened 2022-06-12
Automatic module URLs

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2003
Close https://github.com/jhipster/jhipster-lite/issues/2038


antarus:
Awesome job.

Test and seems OK.

I have a question about the `api/modules` response:
Why not bring up the notion of tree(https://github.com/jhipster/jhipster-lite/issues/1666) and put cucumber "in" springboot with sub category as `tests`?

DamnClin:
> Awesome job.
> 
> Test and seems OK.
> 
> I have a question about the `api/modules` response: Why not bring up the notion of tree(#1666) and put cucumber "in" springboot with sub category as `tests`?

Because I think we need two modes: 
* 1 where you just apply the module you need and that will be used to patch existing projects;
* 1 where you are guided in module applications and that will come later.

For now I'm focusing on the simple modules application, we'll see later for the guided mode

antarus:
Ok that seems like a good approach.

pascalgrimaud:
Merging this as it was tested by @antarus 
Thanks so much

pascalgrimaud:
Adding a bounty here as it's huge work to achieve this
Plz claim it @DamnClin

DamnClin:
Claimed https://opencollective.com/generator-jhipster/expenses/81952

PR #2055, opened 2022-06-13
Simple modules screen

DamnClin:
A simple component to list module. The goal is to use this to apply module depending on there properties but I still don't know what the UX should be on this part.


src/main/webapp/app/loader/primary/Loader.ts
```diff
@@ -0,0 +1,35 @@
+export class Loader<T> {
+  private assignedValue: T | undefined;
```
Gnuk, line 2: Maybe

```typescript
private assignedValue?: T;
```

pascalgrimaud:
Will give a try soon :)

PR #2059, opened 2022-06-13
Set max length for property description

DamnClin:


pascalgrimaud (approved):

src/main/java/tech/jhipster/lite/generator/module/domain/properties/JHipsterPropertyDescription.java
```diff
@@ -6,7 +6,7 @@
 
 public record JHipsterPropertyDescription(String description) {
   public JHipsterPropertyDescription {
-    Assert.notBlank("description", description);
+    Assert.field("description", description).notBlank().maxLength(100);
```
pascalgrimaud, line 9: No unit test for that?

DamnClin: No, I stopped testing assertion some time ago and, so far, so good

Issue #1689, opened 2022-05-10
Refactoring: Spring Boot OAuth2


pascalgrimaud:
Following discussion with @DamnClin 

The current Spring Boot + OAuth2 option is simply a c/c of existing code from generator-jhipster
This part can be improved to respect better hexagonal architecture concept

Cc @hdurix so you can follow the work here too

DamnClin:
I can take this since we've done this refactoring. We'll take some time to battle test and contribute the templates here

PR #2087, opened 2022-06-14
Rework OAuth2 module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/1689


DamnClin:
My current idea is to create an `Account` domain aggregate with: 

* Username
* Name
  * Firstname
  * Lastname
* Email
* Roles

With all those fields mandatory (with roles that can be empty) and provide tooling to read that from token and to expose an endpoint for the authenticated user.

What do you think?

antarus:
I think that's enough at first. we can always upgrade later

pascalgrimaud:
@DamnClin : yes it's exactly what should be done, a specific account context

DamnClin:
Ok, I should be able to do that tomorrow

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/oauth2/domain/OAuth2ModuleFactory.java
```diff
@@ -0,0 +1,192 @@
...
+        .add("WithUnauthenticatedMockUser.java");
+    //@formatter:on
+  }
+
+  private void appendDependencies(JHipsterModuleBuilder builder) {
+    builder
+      .javaDependencies()
+      .add(SRPING_GROUP, artifactId("spring-boot-starter-security"))
+      .add(SRPING_GROUP, artifactId("spring-boot-starter-oauth2-client"))
+      .add(SRPING_GROUP, artifactId("spring-boot-starter-oauth2-resource-server"))
```
DanielFran, line 130:
```suggestion
      .add(SPRING_GROUP, artifactId("spring-boot-starter-security"))
      .add(SPRING_GROUP, artifactId("spring-boot-starter-oauth2-client"))
      .add(SPRING_GROUP, artifactId("spring-boot-starter-oauth2-resource-server"))
```

DanielFran (changes requested):

pascalgrimaud (approved):

PR #2099, opened 2022-06-15
Bump sonarqube 9.5.0

pascalgrimaud:


pascalgrimaud:
Sonarqube 9.5 needs commits, otherwise it will fail like this

Issue #1652, opened 2022-05-07
Angular: health page


pascalgrimaud:
After generating Angular as client, it could be nice to have the possibility to generate health page:

- entry in menu
- health html
- health component
- health service
- health tests

SImilar to:
![image](https://user-images.githubusercontent.com/9156882/167257792-7b6a661a-1b21-4ce9-a1c9-d4634385f357.png)

qmonmert:
@pascalgrimaud I don't understand why on JHLite, I have:
<img width="483" alt="image" src="https://user-images.githubusercontent.com/9989211/168442058-48dfa59f-75cc-4e21-a943-e2cc1f1930ab.png">
<img width="351" alt="image" src="https://user-images.githubusercontent.com/9989211/168442069-57e0241e-2dc9-4a9f-b2a9-6341600819e7.png">

whereas on Generator:
<img width="464" alt="image" src="https://user-images.githubusercontent.com/9989211/168442086-a81f3b74-e5bb-48a3-9533-0329eb944056.png">
<img width="637" alt="image" src="https://user-images.githubusercontent.com/9989211/168442100-828b3d91-196c-497a-b142-0a554c5fcba3.png">
(I have property `components`). An idea?

pascalgrimaud:
@qmonmert : I edit your pull request to keep this opened.

Can you add the front part to generate the health check plz? Then, we can close this one after that :)

![image](https://user-images.githubusercontent.com/9156882/173880005-056fd0fe-a753-441d-ac90-a13259b396f7.png)

qmonmert:
@pascalgrimaud https://opencollective.com/generator-jhipster/expenses/82559 :)

pascalgrimaud:
@qmonmert : approved

PR #2120, opened 2022-06-15
Front part to generate Angular Health

qmonmert:
Related to #1652


src/test/javascript/spec/springboot/secondary/client/AngularRepository.spec.ts
```diff
@@ -58,4 +58,21 @@ describe('AngularRepository', () => {
     const [projectFolder] = projectHistoryService.get.getCall(0).args;
     expect(projectFolder).toBe(PROJECT_FOLDER);
   });
+
+  it('should add Oauth2', async () => {
```
pascalgrimaud, line 62: Bad c/c: should not be oauth2 :-p

pascalgrimaud (changes requested):
- missing cypress tests
- missing projection to handle the button when it is already used at least 1 time

src/test/javascript/spec/springboot/secondary/client/AngularRepository.spec.ts
qmonmert: oups

pascalgrimaud (approved):

PR #2131, opened 2022-06-16
Front : add Ehcache with XML config button

### LINKED ISSUE #1634: Front: add ehcache (xml) button

marioneyraud:
fix #1634


src/test/javascript/spec/springboot/primary/generator/spring-boot-generator/SpringBootGenerator.spec.ts
```diff
@@ -1077,6 +1077,48 @@ describe('SpringBootGenerator', () => {
...
+      await component.addEhcacheWithXml();
+
+      const args = springBootService.addEhcacheWithXML.getCall(0).args[0];
+      expect(args).toEqual({
+        baseName: 'beer',
+        folder: 'project/path',
+        projectName: 'Beer Project',
+        packageName: 'tech.jhipster.beer',
+        serverPort: 8080,
+      });
```
pascalgrimaud, line 1106: I think you should update your local main with upstream, as you can use projectJson here

Issue #2111, opened 2022-06-15
Convert MariaDB to module


DamnClin:
Convert MariaDB (src/main/java/tech/jhipster/lite/generator/server/springboot/database/mariadb) to module (this is a good first issue)

Issue #2106, opened 2022-06-15
Add the possibility to comment spring properties in modules


DamnClin:
Sometimes it's usefull to add some comments in spring properties, the API can be:

```
.springMainProperties()
  .comment(propertyKey("springdoc.swagger-ui.operationsSorter"), comment("This is a comment"))
  .set(propertyKey("springdoc.swagger-ui.operationsSorter"), propertyValue("alpha"))
```

This will add a comment before the given property key

DamnClin:
Hum, I'm really not sure this is a good first issue, designing this may be challenging! (we have to stay yaml friendly)

pascalgrimaud:
ahaha ok, 'good first issue' removed

pascalgrimaud:
Adding a bounty as I'd like to have this

pascalgrimaud:
Increasing the bounty, following the amount of work and time spent on this

renanfranca:
> Increasing the bounty, following the amount of work and time spent on this

Thanks @pascalgrimaud 💪❤️😃!

Issue #1954, opened 2022-06-05
Refactor FileSystemJHipsterModulesRepository


DamnClin:
Once the current PR related to modules are merged a refactoring on `FileSystemJHipsterModulesRepository` is needed. It's responsibilities are not good (so does some changes related responsibilities like `JHipsterModuleReplacements` who should not read file content directly).

This partly needs to be moved to a domain service. Nothing really tricky here but I want to avoid rebase nightmare so I'll wait for some merge :)

DamnClin:
Done in https://github.com/jhipster/jhipster-lite/issues/2468

PR #2154, opened 2022-06-19
Migrate mariadb domain service to new module system

DanielFran:
Fix https://github.com/jhipster/jhipster-lite/issues/2111


DanielFran:
@DamnClin invoking [mariaDBApplicationService.build(properties)](https://github.com/jhipster/jhipster-lite/pull/2154/files#diff-b1efa59419a4d21736d7ae6259334ccf3152ae06957fdd8c09922aa4288750a4R229) did not execute the module, so the test fails because the source files were not created.
What is the easiest way to execute a module in an integration test?

DamnClin:
Hello,

You can create a component test with cucumber, ex: https://github.com/jhipster/jhipster-lite/blob/main/src/test/features/cucumber.feature or https://github.com/jhipster/jhipster-lite/blob/main/src/test/features/infinitest.feature

The goal of those tests is not to ensure that all files are set but just to check module application. The steps are written so you can stick to them

DanielFran:
@DamnClin The current issue happens with liquibase integration tests that use the MariaDB service and require the module to create the source files. Otherwise, the tests fail...

DamnClin:
Oh sorry, haven't checked the logs...

For now you can do:

```
JHipsterModulesAssertions.assertThatModule(mariaDBApplicationService.build(properties))
```

That's not really pretty but that will do the job since those tests are meant to disappear soon. Duplicating the logic in `assertThatModule` is probably a bad idea since I'll totally rework that in https://github.com/jhipster/jhipster-lite/issues/1954

I'll try to refactor this to have a better (understandable) way of applying modules but that depend on our PRs timming :D

DamnClin:
Following https://github.com/jhipster/jhipster-lite/pull/2174 you can use: 

```
TestJHipsterModules.applyer()
  .module(...)
  .properties(...)
  .slug(...)
  .apply()
```

Hope that helps

DamnClin:
Now that you have played with the modules mechanism do you have any improvement ideas?

DamnClin:
For the comment in property files I was thinking about that kind of API: https://github.com/jhipster/jhipster-lite/issues/2106

DanielFran:
Seems good, but actually the order of the properties is not maintained, so this needs also to be fixed.

DamnClin:
I'm not sure we want to keep properties order, this may be strange when replacing an existing property or for yaml support

DanielFran:
OK, but how will you ensure the comments are located before the good properties?

DamnClin:
> No idea for now, I never think about code before doing it, I'll figure that out while doing it :)

This probably needs some explanations :P. In fact that's one of TDD greatest strength, the one that allows you to be really faster. You only consider problems when you are really faced to them with all information at sight. Ian Cooper makes a really good explanation of that in [TDD, Where Did It All Go Wrong](https://www.youtube.com/watch?v=EZ05e7EMOLM)

DanielFran:
@DamnClin DId you see any other changes required for this PR, or it is OK for you to merge?

Issue #1430, opened 2022-04-22
Liquibase: Version defined in changelogs should match liquibase version


DanielFran:
All changelogs should define correct liquibase version. 

Example: dbchangelog-4.5.xsd should be updated to use current liquibase "_major.minor_" version
https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/springboot/dbmigration/liquibase/resources/master.xml#L5
````xml
  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.5.xsd"

````

PR #2155, opened 2022-06-19
Liquibase changelog version

DanielFran:
Fix https://github.com/jhipster/jhipster-lite/issues/1430

Starting on liquibase version [4.12.0](https://github.com/liquibase/liquibase/releases/tag/v4.12.0), _it has been introduced "latest" xsd file path. By using "latest", Liquibase will use the bundled XSD for it's version. The uploaded "latest" XSD will always be the most recent Liquibase release, so note how they can be different and locally cached IDE versions may be different yet._


pascalgrimaud:
Thanks a lot @DanielFran

Issue #1947, opened 2022-06-04
Module: handle package.json


pascalgrimaud:
It should be possible, in package.json, to manage:
- dependencies
- devDependencies
- scripts

DamnClin:
I think I'll take this one but I don't know if it is interesting to do it with somebody else (pairing or just sharing ideas)... Any though on that? Any volunteer?

PR #2176, opened 2022-06-20
Modules packages json

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/1947

This is a first version for packageJson handling in modules. Not really elegant but moving in the way of https://github.com/jhipster/jhipster-lite/issues/1954


pascalgrimaud:
important to have, if someone would like to migrate front modules
so let's merge this

PR #2194, opened 2022-06-21
Add default maven goal

DamnClin:
Just adding a default goal to avoid error when running ./mvnw

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,42 +70,42 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.18.6",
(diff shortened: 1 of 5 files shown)


pascalgrimaud:
The only problem with this, is the user who wants to build a simple lib (init, maven, java base) without spring boot

The best solution would be:
- when adding spring boot, add this default goal if it doesn't already exist

But I'm fine to merge this

DamnClin:
Yep, right, we can wait for spring boot migration to module and do that in a postAction then (as replacement won't check if the value is there already)

Issue #2210, opened 2022-06-22
Add option to wipe data in cucumber module


DamnClin:
Add a boolean module option to wipe all data before each test when using cucumber

PR #2211, opened 2022-06-22
Add JPA reset option for cucumber

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2210


DamnClin:
To be honest I think this should be a dedicated module in the future BUT I need a boolean option in a module to advance on integration and this is the first one that came to my mind...

pascalgrimaud:
It looks good. Anyway, it can't work for NoSQL database
How could we manage this?

PR #2213, opened 2022-06-22
Clean code

qmonmert:


qmonmert:
@DamnClin I saw that
https://www.developer.com/design/how-to-use-enum-effectively-in-java/#:~:text=Enum%20cannot%20have%20a%20default,redundant%20to%20specify%20them%20explicitly.&text=Note%20that%20the%20enum%20constructor,do%20not%20require%20new%20keyword.

DamnClin:
Considering behavior described here https://stackoverflow.com/questions/30454635/when-using-for-a-primitive-and-a-boxed-value-is-autoboxing-done-or-is-unbox I think it's better to keep the explicit "unboxings"

qmonmert:
@DamnClin ok I will remove the commit "unboxing"

pascalgrimaud (approved):

PR #2231, opened 2022-06-26
Modules screen

DamnClin:

So this is a very very simple (and ugly) implementation of the module application screen, lots of UX element are probably missing but this is usable


pascalgrimaud:
Very impressive! Thanks a lot!

DamnClin:
> Very impressive! Thanks a lot!

NP :D I want to add some component test using cypress on that because the vue test utils test are not really testing anything...

pascalgrimaud:
I'm adding a specific bounty for this, as it's well deserved (you can claim it when it's merged)

Yes, about Cypress, indeed, we didn't add real tests so this part should be improved

pascalgrimaud:
you claim it a little bit too early but it's fine, I'm merging this :)

DamnClin:
> you claim it a little bit too early but it's fine, I'm merging this :)

Oh sorry, read too fast... (was drinking coffee)

PR #2239, opened 2022-06-27
Add cypress tests to modules page

DamnClin:


src/test/javascript/cypress/support/Interceptor.ts
```diff
@@ -0,0 +1,21 @@
+import { HttpResponseInterceptor, RouteMatcher, StaticResponse } from '../../../../../node_modules/cypress/types/net-stubbing';
```
pascalgrimaud, line 1: just wonder if we can improve this import without using ../../../etc

pascalgrimaud (approved):

src/test/javascript/cypress/support/Interceptor.ts
DamnClin: No idea :D the front build is totally broken from me, I have to to all thinks by hand using VSCode it's a nightmare to contribute JHLite frontend :D (perhaps I'll take some time to fix that...)

Issue #2253, opened 2022-06-27
Add generated annotation to java base


DamnClin:
The Generated annotation is used to mark generated / not relevant code that jacoco will ignore, we need that in java base :)

PR #2256, opened 2022-06-27
Add generated annotation to java base

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2253

### DIFF
--- a/src/main/resources/generator/server/javatool/base/Generated.java.mustache
+++ b/src/main/resources/generator/server/javatool/base/Generated.java.mustache
@@ -1,4 +1,4 @@
-package tech.jhipster.lite.common.domain;
+package {{packageName}}.common.domain;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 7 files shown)


src/main/resources/generator/server/javatool/base/Generated.java.mustache
```diff
@@ -0,0 +1,11 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE, ElementType.CONSTRUCTOR })
+public @interface Generated {
+}
```
pascalgrimaud, line 11: What do you think about adding something like:

```
  String description() default "";
```

So the user can comment when he/she wants, to explain the reason of adding this annotation in his/her project

DamnClin: what to you think of reason instead of description?

pascalgrimaud: `reason` instead of `description` is fine too

pascalgrimaud (approved):

Issue #2276, opened 2022-06-28
Migrate init module


DamnClin:
It should be possible to migrate the init module now, I'll give it a try

PR #2280, opened 2022-06-28
Init module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2276


src/main/java/tech/jhipster/lite/generator/init/domain/InitModuleFactory.java
```diff
@@ -0,0 +1,86 @@
...
+      .context()
+        .put("prettierEndOfLine", endOfLine(properties))
+        .put("dasherizedBaseName", WordUtils.kebabCase(properties.projectName().get()))
+        .put("nodeVersion", NODE_VERSION)
+        .and()
+      .files()
+        .batch(SOURCE, DESTINATION)
+          .file(".lintstagedrc.js")
+          .file(".prettierignore")
+          .template(".prettierrc")
```
pascalgrimaud, line 39: here the current result of this file, which is not correct, because of `{{^prettierDefaultIndent}}2{{ /prettierDefaultIndent }}`

We could simply delete this default value from mustache, if it's easier

image

src/main/java/tech/jhipster/lite/generator/init/domain/InitModuleFactory.java
```diff
@@ -0,0 +1,86 @@
...
+          .template("package.json")
```
pascalgrimaud, line 40: the package.json is not correct too:

image

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/init/domain/InitModuleFactory.java
DamnClin: Removal was the plan but looks like I missed that one :D fixing it

src/main/java/tech/jhipster/lite/generator/init/domain/InitModuleFactory.java
DamnClin: For the project name I'm relying on the same method from WordUtils so I'm confused here, what's the expected behavior?

DamnClin: I think the name was JHipster Project and the current method result in that no?

DamnClin: Ok, that was project base name and not project name... My bad

Issue #2113, opened 2022-06-15
Convert MySQL to module


DamnClin:
Convert MySQL (src/main/java/tech/jhipster/lite/generator/server/springboot/database/mysql) to module (this is a good first issue)

PR #2299, opened 2022-06-29
Migrate mysql domain service to new module system

DanielFran:
Fix https://github.com/jhipster/jhipster-lite/issues/2113


src/test/java/tech/jhipster/lite/generator/history/infrastructure/primary/GeneratorHistoryInterceptorIT.java
```diff
@@ -28,7 +28,7 @@ void shouldNotAddHistoryOnUnknownFolder() throws Exception {
 
     mockMvc
       .perform(
-        post("/api/servers/spring-boot/databases/mysql")
+        post("/api/servers/spring-boot/databases/mssql")
```
pascalgrimaud, line 31: Just wonder why this change? Does it work only with MSSQL now?

DanielFran: @pascalgrimaud need to check, with modules, when calling an endpoint that is not available, the rest API returns not found instead of the expected bad request

But it is something to review in another issue...

Issue #2311, opened 2022-06-30
Migrate codespace to module


DamnClin:
Migrate src/main/java/tech/jhipster/lite/generator/setup/codespaces to module

PR #2317, opened 2022-06-30
Migrate codespace to module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2311


src/main/java/tech/jhipster/lite/generator/setup/codespaces/domain/CodespaceModuleFactory.java
```diff
@@ -0,0 +1,28 @@
+package tech.jhipster.lite.generator.setup.codespaces.domain;
+
+import static tech.jhipster.lite.module.domain.JHipsterModule.*;
+
+import tech.jhipster.lite.error.domain.Assert;
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.JHipsterSource;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+
+public class CodespaceModuleFactory {
```
pascalgrimaud, line 10: it should be `CodespacesModuleFactory`

src/main/resources/generator/setup/codespaces/devcontainer.json
```diff
@@ -22,7 +22,7 @@
     "esbenp.prettier-vscode",
     "christian-kohler.npm-intellisense"
   ],
-  "forwardPorts": [{{serverPort}}, 9000],
+  "forwardPorts": [8080, 9000],
```
pascalgrimaud, line 25: we should keep this file as mustache, because the default port can be changed when using Spring Boot MVC, specially when the user generated different microservices, so there are different ports

pascalgrimaud (changes requested):

src/main/resources/generator/setup/codespaces/devcontainer.json
DamnClin: I added the property to the module with a default value to 8080

pascalgrimaud: Yes, it means the previous implem was not correct so. Thanks for changing this

DamnClin: To be honest I had a doubt...

pascalgrimaud:
Can you rebase and fixing the conflict plz?

pascalgrimaud (approved):

PR #2324, opened 2022-06-30
Ignore the whole .vscode folder

DamnClin:


DamnClin:
Ah... in fact I'm not confident on this change, I want to try it first here :)

pascalgrimaud:
Oh I see, then let's merge this for now, we'll see later if we should report this change to the generated one

Issue #2179, opened 2022-06-20
Add java dependency management in module system


antarus:
Add the possibility to manage dependency management in the module system.
manage `<dependencyManagement>` in pom.xml

(need for migrate springcloud sub modules)

DamnClin:
For this one I was thinking to add POM in `JavaDependencyScope` and put them in dependency management but I need to RTFM to ensure that it is a good way to do that :D

DamnClin:
I started something for that. Can't end it right now but hopefully soon enough :D. Can you affect me on this one please?

pascalgrimaud:
Thanks @DamnClin

PR #2349, opened 2022-07-03
Dependency management in modules

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2179


pascalgrimaud:
As you renamed a method, there are lot of impacts, so I'll merge it as soon as the CI is green

DamnClin:
In fact there was a lot of "real" changes made here... was way harder than expected (due to the import type specificity)

Issue #2116, opened 2022-06-15
Convert eureka to module


DamnClin:
Convert Eureka (src/main/java/tech/jhipster/lite/generator/server/springboot/springcloud/eureka) to module (this is a good first issue)

DamnClin:
This needs https://github.com/jhipster/jhipster-lite/issues/2179

PR #2355, opened 2022-07-03
Migrate eureka to module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2116

The issue was a mistake: this wasn't a good to start issue... Sorry about that


src/main/java/tech/jhipster/lite/generator/server/springboot/springcloud/eureka/domain/EurekaModuleFactory.java
```diff
@@ -0,0 +1,99 @@
...
+        .and()
+      .build();
+    //@formatter:on
+  }
+
+  private String instanceId(String lowerCaseBaseName) {
+    return lowerCaseBaseName + ":${spring.application.instance-id:${random.value}}";
+  }
+
+  private JavaDependency springCloundDependenciesManagement() {
```
pascalgrimaud, line 89:
```suggestion
  private JavaDependency springCloudDependenciesManagement() {
```

Issue #2353, opened 2022-07-03
Add section to readme using modules


DamnClin:
We can add something to add a readme section in modules using optional replacements with needles

PR #2375, opened 2022-07-04
Add readme section from module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2353


src/main/java/tech/jhipster/lite/module/domain/JHipsterModule.java
```diff
@@ -277,8 +279,16 @@ public JHipsterModuleBuilder documentation(DocumentationTitle title, JHipsterSou
       String target = "documentation/" + title.filename() + source.extension();
       files().add(source, to(target));
 
-      String markdownLink = "- [" + title.get() + "](" + target + ") " + LINE_BREAK + JHIPSTER_DOCUMENTATION_NEEDLE;
-      optionalReplacements().in("README.md").add(text(JHIPSTER_DOCUMENTATION_NEEDLE), markdownLink);
+      String markdownLink = "- [" + title.get() + "](" + target + ") ";
```
pascalgrimaud, line 282: do we need this additional space at the end ?

```suggestion
      String markdownLink = "- [" + title.get() + "](" + target + ")";
```

DamnClin: I don't think so :)

Issue #2117, opened 2022-06-15
Convert gateway to module


DamnClin:
Convert gateway (src/main/java/tech/jhipster/lite/generator/server/springboot/springcloud/gateway) to module (this is a good first issue)

DamnClin:
This needs https://github.com/jhipster/jhipster-lite/issues/2179

PR #2382, opened 2022-07-04
Gateway module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2117


antarus:
seems good. Unfortunately I can't test it today.

DamnClin:
> seems good. Unfortunately I can't test it today.

We spot a very probable mistake in the existing module with @pascalgrimaud in the bootsrap properties management (that were on yaml style in a properties file). This version is supposed to be better BUT I have no idea of what those properties were doing (and don't really want to take a deep look at it). @pascalgrimaud can test that easily on is current project I think :)

pascalgrimaud:
Yes, I'll review and test it very soon

pascalgrimaud:
Just tested it at and it works well:

image

The API too http://localhost:8080/api/gateway/routes

But we need to ignore these logs to logback, otherwise, we are spammed in the console :D

```xml
  <logger name="org.apache" level="ERROR" />
  <logger name="reactor.netty" level="ERROR" />
  <logger name="io.netty" level="ERROR" />
```

pascalgrimaud:
Can you resolve the conflict plz @DamnClin so it can be merged

Issue #2078, opened 2022-06-14
Secure path generation


pascalgrimaud:
By default, path folder should be something like `/tmp/jhlite/xxxx`, `xxxx` can be a random uuid
So the cloud version is usable
The value `/tmp/jhlite` should be in `application.properties`

Then, we can add a "local" profile, which will disable this feature and allow the user to choose the path folder he/she wants

pascalgrimaud:
Adding a bounty as we need this, specially @jdubois :)

hdurix:
as discussed with @pascalgrimaud, we'll start with 2 things to make this work:
- front-end part: instead of a placeholder for project path, init the input box with `/tmp/jhlite/{uuid}`
- back-end part: add a spring boot property `forcedFolderPath` and check at generation the user project path starts with the value of `forcedFolderPath`

jdubois:
Instead of "/tmp/jhlite" can we have "/jhlite"?
"/tmp" will be deleted if there’s a reboot, and also won’t work as a shared drive (I want to mount an Azure File storage so it’s persistent and can be shared if I scale out with multiple servers).

hdurix:
ok so let me change the spec then:
- back-end part: add a spring boot property `folderPathPrefix` and prefix the user project path with the value of `folderPathPrefix`

In the front-end part, the value will be initialized with `/tmp/jhlite/{uuid}` and the app will be generated in `/jhlite/tmp/jhlite/{uuid}`

DamnClin:
I think a spring profile may be better than a property for that since it will be easier to replace adapters implementations if needed

pascalgrimaud:
Updated the bounty as it was more work than expected

hdurix:
@jdubois you should now be able to use an env var named `APPLICATION_FORCED_PROJECT_FOLDER` with the prefix you want :)
This will be used for 2 things:
- initialize the project folder input with forced-folder+UUID in the front-end
- check that the chosen project folder is in the forced folder in the back-end

Eventually UX could be better, but for now at least this is secured

pascalgrimaud:
@hdurix @jdubois : another solution is to use cloud profile, it will have default config /tmp/jhlite

PR #2449, opened 2022-07-08
Forced project folder

hdurix:
Be able to force a specific folder (to protect machine when using in cloud environment).

Only for the legacy modules for now.

related to https://github.com/jhipster/jhipster-lite/issues/2078


pascalgrimaud:
Good job @hdurix !!

PR #2457, opened 2022-07-09
Make jest-sonar optional in Vue

wmarques:
The jest-sonar replacement in `package.json` shouldn't be mandatory IMO, as it won't stop an app from starting.
Btw maybe we can put the `jestSonar` method in a "Common" package as I think it will be the same for each client...
But I don't know where to put it, if you have any advice :) 

Ping @DamnClin


DamnClin:
I'm in the "let's duplicate for now and see later" team when it comes to that type of stuff. I usually start thinking during the 3rd one

PR #2471, opened 2022-07-10
Add manual format to typescript module factory

DamnClin:


DamnClin:
@SoniaLahi I made a manual format here because we kinda agreed that it was more readable for this kind of module. Tell me if you disagree and I close this :)

src/main/java/tech/jhipster/lite/generator/typescript/domain/TypescriptModuleFactory.java
```diff
@@ -20,29 +20,31 @@ public class TypescriptModuleFactory {
...
+        .addDevDependency(packageName("eslint-plugin-prettier"), COMMON)
+        .addDevDependency(packageName("jest"), COMMON)
+        .addDevDependency(packageName("@types/jest"), COMMON)
+        .addDevDependency(packageName("ts-jest"), COMMON)
+        .addScript(scriptKey("test"), scriptCommand("jest"))
+        .addScript(scriptKey("test:watch"), scriptCommand("jest --watch"))
+        .addScript(scriptKey("test:watch:all"), scriptCommand("jest --watchAll"))
+        .addScript(scriptKey("eslint:ci"), scriptCommand("eslint './**/*.{ts,js}'"))
+        .addScript(scriptKey("eslint"), scriptCommand("eslint './**/*.{ts,js}' --fix"))
+        .and()
```
SoniaLahi, line 41: I think we should remove the space before `.and()` here so we have this kind of format :
```java
moduleBuilder(properties)
  .packageJson()
    .addDevDependency(packageName("typescript"), COMMON)
    .addScript(scriptKey("eslint"), scriptCommand("eslint './**/*.{ts,js}' --fix"))
  .and()
  .files()
    .add(SOURCE.file(".eslintrc.js"), to(".eslintrc.js"))
  .build();
```

pascalgrimaud: I'd prefer keeping space for 'and', as when reading this code, the 'and' doesn't provide interesting information, but packageJson and files do

Issue #2473, opened 2022-07-10
Migrate maven frontend


DamnClin:
Migrate src/main/java/tech/jhipster/lite/generator/server/javatool/frontendmaven to module

DamnClin:
Looking at that I see that we are reading `node.version` and `npm.version` from `src/main/resources/generator/dependencies/pom.xml`but I'm confused, is this really handled by dependabot? After a quick look I really don't think those 2 versions are really handled by dependabot (there are no dependency using them) but i really don't know this tool.

@pascalgrimaud can you confirm that these 2 versions won't ever be automatically managed that way and that we need to find a way to handle them?

pascalgrimaud:
Yes I confirm the current system cant upgrade automatically node and npm, in pom.xml

PR #2491, opened 2022-07-10
Frontend maven module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2473


src/main/java/tech/jhipster/lite/generator/server/javatool/frontendmaven/domain/FrontendMavenModuleFactory.java
```diff
@@ -0,0 +1,192 @@
...
+
+public class FrontendMavenModuleFactory {
+
+  private static final JHipsterSource SOURCE = from("server/springboot/mvc/web");
+  private static final JHipsterSource MAIN_SOURCE = SOURCE.append("src");
+  private static final JHipsterSource TEST_SOURCE = SOURCE.append("test");
+
+  private static final String REDIRECTION_PRIMARY = "technical/infrastructure/primary/redirection";
+
+  private static final String NODE_VERSION = "v16.15.0";
```
DamnClin, line 19: hardcoded for now because I think the version update don't work on that

src/main/java/tech/jhipster/lite/generator/server/javatool/frontendmaven/domain/FrontendMavenModuleFactory.java
```diff
@@ -0,0 +1,192 @@
...
+  private static final String NPM_VERSION = "8.5.0";
```
DamnClin, line 20: hardcoded for now because I think the version update don't work on that

src/main/java/tech/jhipster/lite/generator/server/javatool/frontendmaven/domain/FrontendMavenModuleFactory.java
pascalgrimaud, line 19: We could already use 16.16.0

DamnClin: I think what we really need is that https://github.com/jhipster/jhipster-lite/issues/2499

src/test/java/tech/jhipster/lite/generator/server/javatool/frontendmaven/domain/FrontendMavenModuleFactoryTest.java
```diff
@@ -0,0 +1,167 @@
...
+      .propertiesBuilder(TestFileUtils.tmpDirForTest())
+      .basePackage("com.jhipster.test")
+      .projectBaseName("myapp")
+      .build();
+
+    JHipsterModule module = factory.buildModule(properties);
+
+    assertThatModuleWithFiles(module, pomFile())
+      .createFile("pom.xml")
+      .containing("    <node.version>v16.15.0</node.version>")
```
pascalgrimaud, line 29: Do we need to update the test, each time we upgrade the node version?
Instead, I'd suggest to use the constant here

src/test/java/tech/jhipster/lite/generator/server/javatool/frontendmaven/domain/FrontendMavenModuleFactoryTest.java
```diff
@@ -0,0 +1,167 @@
...
+      .containing("    <npm.version>8.5.0</npm.version>")
```
pascalgrimaud, line 30: Same comment for npm here, maybe use the constant

DamnClin: This a mistake in the test, I'll remove the version (just keep the key)

pascalgrimaud (changes requested):
Few comments

pascalgrimaud (approved):

Issue #2460, opened 2022-07-10
Add ajva dependencies removal in module mechanism


DamnClin:
Add the possibility to ask for a dependency removal in the module mechanism (create an explicit `RemoveDirectJavaDependency`). Use case: you want to remove mockito and junit direct dependencies when adding spring boot starter test

pascalgrimaud:
Perfect use case

DamnClin:
Should be straightforward to do, I'll take it but after https://github.com/jhipster/jhipster-lite/pull/2463

PR #2500, opened 2022-07-11
Allow explicit java dependency removal in module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2460


src/test/java/tech/jhipster/lite/generator/server/springboot/core/domain/SpringBootCoreModuleFactoryTest.java
```diff
@@ -28,75 +28,115 @@ void shouldBuildModule() {
...
         """
-              <dependency>
-                <groupId>org.springframework.boot</groupId>
-                <artifactId>spring-boot-dependencies</artifactId>
-                <version>${spring-boot.version}</version>
-                <scope>import</scope>
-                <type>pom</type>
-              </dependency>
-        """
+                  <dependency>
```
pascalgrimaud, line 31: auto reformat from your IDE I think

pascalgrimaud (approved):

PR #2505, opened 2022-07-11
Delete unused build tools methods

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2504


DamnClin:
Yep, saw it :D. Happens when my finger are not doing what my brain want when amending the commit in vim :D

Issue #2503, opened 2022-07-11
Migrate or delete gradle modules


DamnClin:
Gradle modules `src/main/java/tech/jhipster/lite/generator/buildtool/gradle` allow initialization of a gradle project but after that everything will fail.

I think we should delete that for now to avoid spending time on that. Any though @pascalgrimaud ?

pascalgrimaud:
Can we just remove the gradle part in the front, for now? Or just hide the button

DamnClin:
A plan can be to create the module factory (not the resources) and delete all other parts so we don't lose the business logic BUT we don't need to take more time on that. What do you think?

pascalgrimaud:
I'm fine so, let's move to module system
It won't impact our users as gradle doesn't work currently

What I'd like is to keep the template in resources, qo it can be reused

atomfrede:
I vote for migration :) But it would be totally okay to remove it for now, not sure I'll have enough time in the coming days/weeks to do significant contributions to this.

DamnClin:
Did something, don't hesitate to tell if that wasn't your idea on that :)

atomfrede:
Looks okay so far. I'll see to make it work again, maybe after the moving parts have settled a little

PR #2528, opened 2022-07-12
Gradle migration

DamnClin:
This is not really a migration since gradle support is currently not
working. We agreed to keep the templating and remove the endpoints.

I also deleted some unused methods in project repository in the process

Close https://github.com/jhipster/jhipster-lite/issues/2503


pascalgrimaud:
minor conflict to resolve, maybe because of some merges, few mins ago

PR #2533, opened 2022-07-12
Remove gitAddCommit unused method

DamnClin:


pascalgrimaud:
The original idea was to add and git, like it was done in generator-jhipster
But finally, I'd prefer to let the user do this job, it's better :)
So let's remove it!

Issue #2351, opened 2022-07-03
Cucumber OAuth2 authentication steps module


DamnClin:
Add a module with cucumber authentication steps for OAuth2

PR #2585, opened 2022-07-14
Add cucumber authentication module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2351

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/cucumberauthentication/application/CucumberAuthenticationApplicationService.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/cucumberauthentication/application/CucumberAuthenticationApplicationService.java
@@ -1,0 +1,20 @@
+package tech.jhipster.lite.generator.server.springboot.cucumberauthentication.application;
+
+import org.springframework.stereotype.Service;
+import tech.jhipster.lite.generator.server.springboot.cucumberauthentication.domain.CucumberAuthenticationModuleFactory;
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+
+@Service
+public class CucumberAuthenticationApplicationService {
+
+  private final CucumberAuthenticationModuleFactory factory;
+
+  public CucumberAuthenticationApplicationService() {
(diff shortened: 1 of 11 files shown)


src/main/java/tech/jhipster/lite/generator/server/springboot/cuucmberauthentication/application/CucumberAuthenticationApplicationService.java
```diff
@@ -0,0 +1,20 @@
+package tech.jhipster.lite.generator.server.springboot.cuucmberauthentication.application;
```
pascalgrimaud, line 1: Typo in package

src/main/java/tech/jhipster/lite/generator/server/springboot/cuucmberauthentication/infrastructure/primary/CucumberAuthenticationModuleConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+import tech.jhipster.lite.generator.server.springboot.cuucmberauthentication.application.CucumberAuthenticationApplicationService;
+import tech.jhipster.lite.module.domain.properties.JHipsterModulePropertiesDefinition;
+import tech.jhipster.lite.module.infrastructure.primary.JHipsterModuleApiDoc;
+import tech.jhipster.lite.module.infrastructure.primary.JHipsterModuleResource;
+
+@Configuration
+class CucumberAuthenticationModuleConfiguration {
+
+  @Bean
+  JHipsterModuleResource cucumberAuthenticationModule(CucumberAuthenticationApplicationService cucumberAuthentications) {
```
pascalgrimaud, line 14: I think we should try to distinct jwt and oauth2

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/cuucmberauthentication/infrastructure/primary/CucumberAuthenticationModuleConfiguration.java
DamnClin: I don't know for now, I think this can work for both considering the way it's done. We'll see during JWT migration I think

Issue #2240, opened 2022-06-27
Migrate angular to modules


DamnClin:
Migrate src/main/java/tech/jhipster/lite/generator/client/angular to modules

PR #2634, opened 2022-07-17
Angular health module

DamnClin:
Related to https://github.com/jhipster/jhipster-lite/issues/2240


src/main/resources/generator/client/angular/admin/src/main/webapp/app/admin/health/health.component.html.mustache
```diff
@@ -10,12 +10,12 @@
 
     <ng-container matColumnDef="key">
       <th mat-header-cell *matHeaderCellDef>Service name</th>
-      <td mat-cell *matCellDef="let element">{{=<% %>=}}{{ element.key }}<%={{ }}=%></td>
+      <td mat-cell *matCellDef="let element">{{ element.key }}</td>
```
qmonmert, line 13: are you sure about that? mustaches are ok in the HTML?

DamnClin: Yep, in modules we only replace the mustache keys we know, unknown keys are ignored and left without any updates

DamnClin: (I generated the app with the old and new module to see that I had to update the templates :))

qmonmert: ok thanks 🙂

PR #2635, opened 2022-07-17
Fix stack overflow errors

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2637

PR #2636, opened 2022-07-17
Switch to add template in files batch

DamnClin:


pascalgrimaud:
@DamnClin @antarus : may I merge this without waiting the CI?

DamnClin:
No :D, it's not the good one, I need to exclude a sonar rule on https://github.com/jhipster/jhipster-lite/pull/2635 which is the real fix

Issue #2543, opened 2022-07-13
Add documentation in sonar module


DamnClin:
Sonar module don't add documentation to start sonar and run analysis in the readme

PR #2639, opened 2022-07-17
Add sonar documentation

antarus:
close https://github.com/jhipster/jhipster-lite/issues/2543

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/sonar/domain/SonarModulesFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/sonar/domain/SonarModulesFactory.java
@@ -4,6 +4,7 @@ import static tech.jhipster.lite.module.domain.JHipsterModule.*;
 
 import tech.jhipster.lite.docker.domain.DockerImages;
 import tech.jhipster.lite.error.domain.Assert;
+import tech.jhipster.lite.module.domain.DocumentationTitle;
 import tech.jhipster.lite.module.domain.JHipsterDestination;
 import tech.jhipster.lite.module.domain.JHipsterModule;
 import tech.jhipster.lite.module.domain.JHipsterSource;
@@ -43,7 +44,9 @@ public class SonarModulesFactory {
     return moduleBuilder(properties)
       .context()
       .put("sonarqubeDockerImage", dockerImages.get("sonarqube").fullName())
+      .put("srcMainDocker", "src/main/docker")
       .and()
+      .documentation(new DocumentationTitle("sonar"), SOURCE.template("sonar.md"))
       .javaBuildPlugins()
       .plugin(propertiesPlugin())
       .pluginManagement(sonarPlugin())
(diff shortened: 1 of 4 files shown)


src/main/java/tech/jhipster/lite/generator/server/sonar/domain/SonarModulesFactory.java
```diff
@@ -43,7 +43,9 @@ private JHipsterModuleBuilder commonModuleFiles(JHipsterModuleProperties propert
     return moduleBuilder(properties)
       .context()
       .put("sonarqubeDockerImage", dockerImages.get("sonarqube").fullName())
+      .put("srcMainDocker", "src/main/docker")
       .and()
+      .documentation(documentationTitle("sonar"), SOURCE.template("sonar.md"))
```
pascalgrimaud, line 48: you could add this file to unit test, to be sure it's generated

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/server/sonar/domain/SonarModulesFactory.java
antarus: yes

pascalgrimaud (approved):

antarus:
@pascalgrimaud test add

DamnClin:
I think we should add short readmesection for the docker image (I made a test in the mongo DB module)

antarus:
ok I'll look it up

antarus:
@DamnClin 
Isn't it a little redundant to have the documentation and after the docker-compose to launch?

maybe add a specific needle to the docker to start?

DamnClin:
I don't know, most of the time I have many docker to start (and I make a makefile but that's not the subject...), I'dd like to have all the command in the frontpage

antarus:
I agree with you, I regularly make a script to start everything but if you want to put all the useful commands at launch, they would be better under `npm install` in the readme, wouldn't they?

DamnClin:
Totally, good idea

antarus:
which name would be the most appropriate for this needle ? i can do a MR

DamnClin:
sounds good. I think we should also add a shortcut in JHipsterModule where you give the command and it's doing the other part (formatting, replacement, etc)

pascalgrimaud:
> I don't know, most of the time I have many docker to start (and I make a makefile but that's not the subject...), I'dd like to have all the command in the frontpage

For your information, in generator-jhipster: there are some npm commands to start all docker-compose services. See https://github.com/jhipster/jhipster-sample-app/blob/main/package.json#L32-L38

Some users said me they don't like this, as it looks strange to have these kinds of commands inside `package.json`
So here some suggestions, in JHipster Lite:
- a module which add to `package.json` these kind of commands, using docker-compose
- a module which will init a makefile and add the different commands

If you think it's a good idea, we can open a ticket for that

Issue #2672, opened 2022-07-18
Make cucumber authentication OAuth2 specific


DamnClin:
Sadly there are differences between OAuth2 and JWT authentication that will require specific module for cucumber test for each of those authentication types

PR #2674, opened 2022-07-18
Cucumber oauth2 specific authentication

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2672


src/main/java/tech/jhipster/lite/generator/server/springboot/cucumberauthentication/domain/CucumberAuthenticationModuleFactory.java
```diff
@@ -42,11 +43,11 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
         .and()
       .files()
         .add(
-          SOURCE.template("AuthenticationSteps.java"),
+            OAUTH2_SOURCE.template("AuthenticationSteps.java"),
```
pascalgrimaud, line 46: Maybe extra spaces but I'm not sure

pascalgrimaud (approved):

Issue #2549, opened 2022-07-13
Multiple implementations for GitRepository


DamnClin:
`GitRepository` should have multiple implementations:

* One using JGit (the current one);
* One doing a git command relying on the git installation on the host machine. This should be triggered by a configuration in the app. This will be used for GraalVM builds

PR #2700, opened 2022-07-19
Multiple git repository depending on environment

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2549

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/init/infrastructure/secondary/GitChecker.java
+++ b/src/main/java/tech/jhipster/lite/generator/init/infrastructure/secondary/GitChecker.java
@@ -1,0 +1,33 @@
+package tech.jhipster.lite.generator.init.infrastructure.secondary;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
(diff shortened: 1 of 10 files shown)


src/main/java/tech/jhipster/lite/generator/init/infrastructure/secondary/GitChecker.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+import tech.jhipster.lite.common.domain.Generated;
+
+@Component
+@Generated(reason = "Higly dependant of execution environment")
```
pascalgrimaud, line 11: Typo: Highly

pascalgrimaud (changes requested):

pascalgrimaud:
About disabled tests, don't you think we can let them? For users who launch unit tests on this project, they probably cloned it, using git, so for sure, they have git :-D

DamnClin:
> About disabled tests, don't you think we can let them? For users who launch unit tests on this project, they probably cloned it, using git, so for sure, they have git :-D

Didn't know if you have it on all CI machines...

pascalgrimaud (approved):

Issue #1862, opened 2022-05-28
Handle Gradle dependencies in modules


DamnClin:
Following https://github.com/jhipster/jhipster-lite/issues/1787 we can add gradle dependencies handling.

TODO: 

- [x] Update `FileSystemCurrentJavaDependenciesVersionsRepository` to have a strategy to handle either maven or gradle
- [x] Create a `JavaDependenciesCommandHandler` interface based on `MavenCommandHandler`
- [x] Create a `GradleCommandHandler` (which is a `JavaDependenciesCommandHandler`)
- [x] Update `FileSystemJavaDependenciesCommandsHandler` to create the most suited handler (depending on project version management)

atomfrede:
@DamnClin I have created a first draft PR, maybe you can have a short look if thats heading in the right direction https://github.com/jhipster/jhipster-lite/pull/2748

murdos:
@atomfrede : do you still intend to work on this feature? otherwise I would be happy to give it a try :-)

atomfrede:
Feel to go on. I did not much so far. I think I would have tried rewrite to handle e.g. the dependency part.

pascalgrimaud:
@murdos : if possible, try to split into small PRs or small tickets, as it will be easier to review and to add small bounties on each ;)

murdos:
> @murdos : if possible, try to split into small PRs or small tickets, as it will be easier to review and to add small bounties on each ;)

My current roadmap is:
- [ ] introduce version catalogs feature with TOML in gradle build tool module: it seems to be the good practice, and will ease simplify manipulation of gradle build files that are hard to parse (I was thinking about using needles for that)
- [ ] Introduce GradleCommandHandler and JavaDependenciesCommandHandler interface, and implements the handling of the `SetVersion` command
- [ ] Handle other dependencies related commands (`(Add|Remove)DirectJavaDependency` and `(Add|Remove)JavaDependencyManagement`)
- [ ] Adapt CI tests and scripts to allow testing gradle projects
- [ ] Handle plugins related commands, but probably in another issue, since it will probably requires changing API related to build plugins in module builder ([related to this comment](https://github.com/jhipster/jhipster-lite/issues/4153#issuecomment-1366408863) and to the fact that plugin in gradle only have an id and a version - no groupId and artifactId)
- [ ] Stop hidding gradle build tool resources slug! :-)

atomfrede:
[...]

PR #2748, opened 2022-07-24
Gradle dependencies command handling

atomfrede:
Just for early feedback to know if thats the right direction. Next step would be to add a direct dependency. For the dependency management equivalent I am thinking of using the version catalogs feature, but thats second or third step. 
 

udpates #1862

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/GradleCommandHandler.java
+++ b/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/GradleCommandHandler.java
@@ -11,11 +11,9 @@ import tech.jhipster.lite.module.domain.javabuild.command.RemoveDirectJavaDepend
 import tech.jhipster.lite.module.domain.javabuild.command.RemoveJavaDependencyManagement;
 import tech.jhipster.lite.module.domain.javabuild.command.SetVersion;
 
-public class GradleCommandHandler extends JavaDependenciesCommandHandler {
+class GradleCommandHandler implements JavaDependenciesCommandHandler {
 
-  GradleCommandHandler(Indentation indentation, Path buildGradlePath) {
-    super(indentation, buildGradlePath);
-  }
+  GradleCommandHandler(Indentation indentation, Path buildGradlePath) {}
 
   @Override
   public void handle(SetVersion command) {}
--- a/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/GradleCommandHandler.java
+++ b/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/GradleCommandHandler.java
@@ -1,0 +1,52 @@
+package tech.jhipster.lite.module.infrastructure.secondary;
+
+import java.nio.file.Path;
+import org.apache.commons.lang3.NotImplementedException;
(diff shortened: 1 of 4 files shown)


src/main/java/tech/jhipster/lite/module/infrastructure/secondary/GradleCommandHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import tech.jhipster.lite.module.domain.Indentation;
+import tech.jhipster.lite.module.domain.javabuild.command.AddBuildPluginManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.AddDirectJavaBuildPlugin;
+import tech.jhipster.lite.module.domain.javabuild.command.AddDirectJavaDependency;
+import tech.jhipster.lite.module.domain.javabuild.command.AddJavaDependencyManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.RemoveDirectJavaDependency;
+import tech.jhipster.lite.module.domain.javabuild.command.RemoveJavaDependencyManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.SetVersion;
+
+public class GradleCommandHandler extends JavaDependenciesCommandHandler {
```
DamnClin, line 14: Should be package private

DamnClin (commented):
Looks a lot like what I got in mind expect I wasn't expecting that you'll need xml manipulation on one file for gradle

atomfrede:
Yes, xml manipulation should not be needed. Still thinking of a nice way to do it. The more I tempting using rewrite looks.

DamnClin:
> Yes, xml manipulation should not be needed. Still thinking of a nice way to do it. The more I tempting using rewrite looks.

So you should extract an interface instead of an abstract class

atomfrede:
Ah! 🤦 No I got it, because of the match inside the super class. Yes thats totally nonsense.

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/JavaDependenciesCommandHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import tech.jhipster.lite.module.domain.Indentation;
+import tech.jhipster.lite.module.domain.javabuild.command.AddBuildPluginManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.AddDirectJavaBuildPlugin;
+import tech.jhipster.lite.module.domain.javabuild.command.AddDirectJavaDependency;
+import tech.jhipster.lite.module.domain.javabuild.command.AddJavaDependencyManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.RemoveDirectJavaDependency;
+import tech.jhipster.lite.module.domain.javabuild.command.RemoveJavaDependencyManagement;
+import tech.jhipster.lite.module.domain.javabuild.command.SetVersion;
+
+public interface JavaDependenciesCommandHandler {
```
DamnClin, line 22: The can be package private

murdos:
@atomfrede : I'm closing this draft since a few PR for gradle support have already been merged, and https://github.com/jhipster/jhipster-lite/issues/1862 has been fixed

PR #2768, opened 2022-07-25
Move module form position

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2766


DamnClin:

PR #2802, opened 2022-07-27
Rename cypress module

DamnClin:


nonomoho:
Change in front too?

DamnClin:
> Change in front too?

What do you mean?

nonomoho:
You have `REACT_CYPRESS = 'REACT_CYPRESS',` in Service.ts for instance

DamnClin:
I think this is useless and should be deleted :)

DamnClin:
https://github.com/jhipster/jhipster-lite/pull/2803 :)

nonomoho (approved):

Issue #2733, opened 2022-07-21
Commit each module


DamnClin:
Add an option to create a commit for each applied module

PR #2804, opened 2022-07-27
Commit module application

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2733


pascalgrimaud:
Following this, the SonarCloud analysis failed:
Screenshot_20220728-062811

It wasn't detect in the CI because we still use Sonar 9.4 and Sonarcloud uses probably 9.5

Can you have a look plz @DamnClin ?
Otherwise, I should find time at the end of this morning

DamnClin:
I can check but knowing the issues would help

pascalgrimaud:
Yes, the analysis is public: https://sonarcloud.io/project/overview?id=jhipster_jhipster-lite

DamnClin:
I think I don't know how to read the report because for me this was not caused by this PR and this code was there for a very long time

pascalgrimaud:
No worry and you're right, maybe it was not because of this PR as Sonarcloud can upgrade and change some rules, so it can happen :-D

DamnClin:
The rule is basivcally saying we can't write to user manage path. So I'll just disable it since it is JHlite job do to that

PR #2808, opened 2022-07-28
Allow file manipulation ot user controlled paths

DamnClin:


pascalgrimaud:
Thanks so much for the quick fix

DamnClin:
I'm doing the sonar upgrade (testing it right now)

Issue #2813, opened 2022-07-28
Bug: OAuth2 and authenticated-user-account api


pascalgrimaud:
When generating OAuth2 with account, the API doesn't work:

![image](https://user-images.githubusercontent.com/9156882/181457772-d6958bd8-0b66-4601-aa6d-99b1d77971fc.png)

Here the stacktrace:

```
2022-07-28 10:20:41.760 ERROR 53723 --- [nio-8080-exec-8] o.z.problem.spring.common.AdviceTraits   : Internal Server Error

tech.jhipster.beer.account.infrastructure.secondary.UnknownAuthenticationSchemeException: Tried to read authentication from an unknown shceme
	at tech.jhipster.beer.account.infrastructure.secondary.OAuth2AuthenticationReader.readAttributes(OAuth2AuthenticationReader.java:56)
	at tech.jhipster.beer.account.infrastructure.secondary.OAuth2AuthenticationReader.toAccount(OAuth2AuthenticationReader.java:27)
	at java.base/java.util.Optional.map(Optional.java:260)
	at tech.jhipster.beer.account.infrastructure.secondary.OAuth2AuthenticationReader.authenticatedUserAccount(OAuth2AuthenticationReader.java:19)
	at tech.jhipster.beer.account.infrastructure.secondary.OAuth2AccountsRepository.authenticatedUserAccount(OAuth2AccountsRepository.java:19)
	at tech.jhipster.beer.account.infrastructure.secondary.OAuth2AccountsRepository$$FastClassBySpringCGLIB$$b91ad814.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:793)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:137)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
[...]

PR #2816, opened 2022-07-28
Fix OAuth2 authenticated user account api

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/2813


src/main/resources/generator/server/springboot/mvc/security/oauth2/account/main/infrastructure/secondary/OAuth2AuthenticationReader.java.mustache
```diff
@@ -54,7 +53,10 @@ class OAuth2AuthenticationReader {
     if (authentication instanceof OAuth2AuthenticationToken oauthToken) {
       return oauthToken.getPrincipal().getAttributes();
     }
+    if (authentication instanceof JwtAuthenticationToken jwtToken) {
```
DamnClin, line 56: Adding a line break before it can be a good idea

pascalgrimaud: thanks, done

PR #2818, opened 2022-07-28
Clean the readme

### LINKED ISSUE #2817: Clean the readme
The current README lists some features, it's not needed anymore as the module system will list them.
So the README needs to be cleaned

qmonmert:
Fix #2817


pascalgrimaud:
Thanks @qmonmert :)

PR #2835, opened 2022-07-29
Add documentation link

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2833

### DIFF
--- a/src/main/webapp/app/common/primary/header/Header.vue
+++ b/src/main/webapp/app/common/primary/header/Header.vue
@@ -29,8 +29,19 @@
 
         <ul class="navbar-nav flex-row flex-wrap ms-md-auto">
           <li class="nav-item col-6 col-md-auto">
+            <a
+              class="nav-link p-2"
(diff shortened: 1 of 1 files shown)


DamnClin:
cc @jdubois with this you have an updated usage guide and link to it in the app (if you want to deploy). Far from being perfect but we still have a long way to go to the modules hierarchy :) (see https://github.com/jhipster/jhipster-lite/discussions/2671).

BTW is there an easy way to have CD for this app?

jdubois:
@DamnClin I'm not sure what is the best option, as I do this on my fork with my own Azure account -> how about having a GitHub Action that syncs this daily?

DamnClin:
Can do if it's easier to build that a github action here deploying after main build (which would be the first option for me).

jdubois:
My issue is that I can't share my secret token on an OSS repo (not that I don't trust you, but it's for security compliance)

DamnClin:
Totally understandable! We can start like that and change the trigger later on to a trigger at the end of main build here

PR #2911, opened 2022-08-03
Angular OAuth2 module

DamnClin:
I'm not really happy with this one since it is really fragile (the generated app is not in an hexagonal architecture so adding something require lots of modification to existing files) but I can't do better easily and i don't want to spent to much time on that since I'm not using Angular.

So, a migration "as is" of the current module


pascalgrimaud:
Same, I don't use Angular for my projects
@qmonmert : do you want to review or can I merge as it?

qmonmert (approved):
LGTM
I know Angular but I'm not a specialist of hexagonal architecture, sorry for that! Maybe when Vue OAuth2 will be developed, we could adapt this.

PR #2951, opened 2022-08-05
Access ts files from tests in generated Vue.js app

### LINKED ISSUE #1732: Add front modules acces from tests
In the current version when you generate a vue.js front end (an perhaps any other front end) VisualStudio Code don't see main sources from test source (which is very bad for the Dev experience).

I ended up updating `tsconfig.json` to have: 

```
  "include": [
    "src/test/**/*.ts",
    "src/main/webapp/**/*.ts",
    "src/main/webapp/**/*.d.ts",
    "src/main/webapp/**/*.tsx",
    "src/main/webapp/**/*.vue"
  ],
```

I also created a `tsconfig-build.json` with:

```
{
    "extends": "./tsconfig.json",
    "exclude": [
        "src/test/**/*.ts"
    ]
}
```

And updated the build target (in `package.json`) to: 

```
"build": "vue-tsc -p tsconfig-build.json --noEmit && vite build --emptyOutDir",
```

This looks OK but I don't want to contribute that directly since I'm really bad on those part and not sure that's a good way to handle this!

hdurix:
Fix #1732

I checked for every generated front-ends:
- Vue.js: fixed in this PR with adding a `tsconfig.build.json` as we do in JHipster Lite generator
- React & Svelte: this is already working. Note: tsc is run on test files but I think that's not a bad thing to check that we don't have any TypeScript issue/typo in tests when building the app. IMO nothing to do.
- Angular: This is already working. A bit different as tests are collocated with implementation code. Nothing to do.


package.json
```diff
@@ -14,7 +14,7 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "build": "vue-tsc -p tsconfig-build.json --noEmit && vite build --emptyOutDir",
+    "build": "vue-tsc -p tsconfig.build.json --noEmit && vite build --emptyOutDir",
```
hdurix, line 17: I used a more conventional name as I saw on the Internet

pascalgrimaud:
thanks a lot @hdurix

Issue #2973, opened 2022-08-08
Use spring managed FlyWay version


DamnClin:
The last version of Flyway is not compatible with spring (the app just fails), we should stick to the spring managed version for this dependency

Issue #2859, opened 2022-07-31
Upgrade to Flyway v9


pascalgrimaud:
See https://github.com/jhipster/jhipster-lite/pull/2796#issuecomment-1196314446

From comment's @DamnClin :

Some breaking change were introduce in V9, they are described here: https://flywaydb.org/blog/version-9-is-coming-what-developers-need-to-know

Easy to reproduce locally with:

init
maven-java
java-base
springboot
springboot-tomcat
postgresql
flyway

PR #2974, opened 2022-08-08
Stick to spring managed FlyWay version

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2973
Close https://github.com/jhipster/jhipster-lite/issues/2859


DamnClin:
Merging this to fix unwanted upgrades

Issue #2499, opened 2022-07-11
Automatically update node and npm


DamnClin:
Node and NPM version are not automatically managed (they are defined as constants in `InitModuleFactory` and `FrontendMavenModuleFactory` not even to the same value). Let's find a way to manage those versions to automatically upgrade to LTS

PR #2979, opened 2022-08-08
Upgrade node and npm

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2499

To be honest I don't know (at all) if this is a good idea / a good way to do that (not even sure that work, can't really test automatic dependabot updates on that)

@qmonmert any advice on that?


qmonmert:
First time that I will see dependabot working on npm and node versions. The idea is good but I am afraid that tests are ok but the reality will be different 😅 but we can try

qmonmert (approved):

DamnClin:
Yep, I really don't think that'll do but let's try...

qmonmert:
🎸

DamnClin:
Looks like it work: https://github.com/jhipster/jhipster-lite/pull/2984

BUT I don't know if we want to update to the latest versions (cc @pascalgrimaud)

qmonmert:
Good but 18.0.7 is not a LTS ...

DamnClin:
Yep, this is what bothering me...

PR #3006, opened 2022-08-10
Init SpringPropertyComment

### LINKED ISSUE #2106: Add the possibility to comment spring properties in modules

MSaguer:
To close #2106

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/javaproperties/SpringProperty.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/javaproperties/SpringProperty.java
@@ -7,6 +7,7 @@ public class SpringProperty {
   private final SpringPropertyType type;
   private final PropertyKey key;
   private final PropertyValue value;
+  private final PropertyComment comment;
   private final SpringProfile profile;
 
   private SpringProperty(SpringPropertyBuilder builder) {
@@ -16,6 +17,7 @@ public class SpringProperty {
     type = builder.type;
     key = builder.key;
     value = builder.value;
+    comment = builder.comment;
     profile = buildProfile(builder.profile);
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/tech/jhipster/lite/module/domain/javaproperties/SpringProperty.java
```diff
@@ -7,6 +7,7 @@ public class SpringProperty {
   private final SpringPropertyType type;
   private final PropertyKey key;
   private final PropertyValue value;
+  private final PropertyComment comment;
```
DamnClin, line 10: I don't think this belong here since most properties won't come with a comment and that we can totally imagine setting comment on properties not set by the module. I think this should be in a dedicated object (but hard to tell so early, I usually start with an higher level test case for module features and I'm reviewing from phone...)

MSaguer: I've seen comments as an optional value added to a spring property but it can be also splitted in a different class to be able to handle case when we want to add a comment on an existing property.

I think I could use `FileSystemJHipsterModulesRepository.apply(JHipsterModuleChanges changes)` as a start point for my modification.
I that case I would add a new line `springPropertyComments.handle(changes.projectFolder(), changes.springPropertyComments());`

Is that OK for you?

DamnClin: To be honest I can't tell, I start by a test and see where this goes, I can't tell any design like that

pascalgrimaud:
closing this one in favor of https://github.com/jhipster/jhipster-lite/pull/4814

Issue #2996, opened 2022-08-09
Display module landscape


DamnClin:
The first step for the module landscape page is to be able to see the module dependencies (with a minimum cost for contributors). This page will then be used to chose multiple modules and apply them in one go.

I started a display on "fake" data, that looks like this
![landscapeV0](https://user-images.githubusercontent.com/59929679/183684208-fc09ebee-2f0f-4bd3-811c-d859fa67ca47.png)
 
After looking for graph libraries I can't find anything I as happy with (either not the good features or bad licenses or bad packaging...) I ended up drawing this wit ha svg (generated in the JHlite vue app).

Now I have to find a way to generate the needed data on the backend...

DamnClin:
This is related to https://github.com/jhipster/jhipster-lite/issues/1666 as this is the first feature creating a hierarchy between modules.

So I'll have to build the whole hierarchy in this PR BUT it's not a good idea since we also need automated test to ensure that those hierarchy are good (but we have to start that somewhere...).

Issue #2997, opened 2022-08-09
Split multi option modules


DamnClin:
Following https://github.com/jhipster/jhipster-lite/issues/2996 we'll need to split module with multiple options to keep each module doing one thing. Multi option modules are: 

- [x]  springdoc-openapi (checking mvc or webflux)
- [x]  springdoc-openapi-with-security-jwt (checking mvc or webflux)
- [x]  springdoc-openapi-with-security-oauth2 (checking mvc or webflux)
- [x]  flyway (option for mysql / maria types) 
- [x]  dummy-flyway-changelog (option for PostGresSQL types)
- [x]  springboot-cucumber (option for JPA reset)

PR #3022, opened 2022-08-11
Modules landscape

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/2996
Related to https://github.com/jhipster/jhipster-lite/issues/1666

This is a first version, still not really usable as is, here is the beginning of the tree: 
level1-2

This first part is OK but then comes a problem when we have modules depending on multiple levels, here is one of the worst part (bottom of levels 4 to 6): 
level4-6

Easy problems to tackle next: 
* Descriptions are pretty bad, we must do a full review of them;
* Some modules must be split (see https://github.com/jhipster/jhipster-lite/issues/2997).

Harder problems:
* Some dependencies are probably not good. The messing lines may indicate that, we probably need to rethink some of the dependencies;
* Reaching something really readable...
* Were do we put the form with all properties for application of multiple modules?
* And probably lot of other but I think this is an important first step!


DamnClin:
An idea may be to merge all dependencies from a feature to something so we can have less lines

DamnClin:
Back to draft: just saw that resizes send the lines very far, need to fix that :)

DamnClin:
Overnight though: 

* The application form can be a fixed block top left;
* We should give lots of hints during selection: 
  * Cleary mark non selectable modules
  * Change line shape (full vs dotted) for links to selected modules VS not selected modules
  * Change line and box color for selected module (use primary)
  * Animate linked elements when going over a not selected module (something like a color change)

But all this will be in dedicated PR since this one is already huge and we first need to reorganize some modules

hdurix:
Good idea to split PRs :smile: 

I've just read this PR, seems good. Nice work!!

DamnClin:
> Good idea to split PRs
Yep... sorry for this one...

> I've just read this PR, seems good. Nice work!!
Still not really usable but I think this is an interesting step to have a really usable JHlite

hdurix:
Do you think it could be relevant to have only 1 line from a module to a feature? Instead of having 1 line for each module of the feature?

image

Here a line from `springboot` to `jpa-persistence`.

Do you think that in a feature, modules could have different dependencies?

DamnClin:
Not for all cases (because sometimes all modules in the same feature don't share the same dependencies) but this is this comment: 

> An idea may be to merge all dependencies from a feature to something so we can have less lines
> 
> In the current landscape this removes 17 connections, it require some effort but it's probably worth it

Issue #3157, opened 2022-08-19
Split git init and startup


DamnClin:
Split git init from startup and make all modules depends on that to avoid errors when committing modules

PR #3159, opened 2022-08-20
Add git-init module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/3157


DamnClin:
After discussion it's probably better to make a git init (if needed) before each module

PR #3202, opened 2022-08-23
Add move file feature in module

DamnClin:


hdurix (approved):

DamnClin:
cc @SoniaLahi you can use that if you want to move svelte file

Issue #3229, opened 2022-08-24
Rework apply-patches payload


DamnClin:
The apply-patches payload is really misleading since it has properties in properties

PR #3230, opened 2022-08-24
Rename missleading properties

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/3229


src/main/java/tech/jhipster/lite/module/infrastructure/primary/RestJHipsterModuleProperties.java
```diff
@@ -10,29 +11,44 @@ class RestJHipsterModuleProperties {
...
+    this.parameters = parameters;
   }
 
   public JHipsterModuleProperties toDomain(ProjectFolder jHipsterProjectFolderFactory) {
     Assert.notNull("jHipsterProjectFolderFactory", jHipsterProjectFolderFactory);
 
     assertValidProjectFolder(jHipsterProjectFolderFactory);
 
-    return new JHipsterModuleProperties(projectFolder, commit, properties);
+    return new JHipsterModuleProperties(getProjectFolder(), isCommit(), getParameters());
```
hdurix, line 31: is this change useful?

hdurix (approved):

src/main/java/tech/jhipster/lite/module/infrastructure/primary/RestJHipsterModuleProperties.java
DamnClin: Somehow, it's a "trick" :D. I add the getter so springdoc is happy and find the parameters BUT, if I do that I have some missing tests cases so I update the mapping to use the getter to trick the coverrage

hdurix: :see_no_evil:

PR #3254, opened 2022-08-25
Split init & init-minimal to have a prettier module

### LINKED ISSUE #3224: Split init & init-minimal to have a prettier module
Init & init-minimal are not so meaningful. And with Landscape, we force user to choose between both for any project as they are root.

Is there any objection to create a `prettier` module (which could depend on `init`)?

hdurix:
Fix #3224


src/main/java/tech/jhipster/lite/generator/init/infrastructure/primary/InitModuleConfiguration.java
```diff
@@ -16,45 +15,16 @@ class InitModuleConfiguration {
   JHipsterModuleResource initModule(InitApplicationService inits) {
     return JHipsterModuleResource
       .builder()
-      .legacyUrl("/api/inits/full")
+      .legacyUrl("/api/inits")
       .slug("init")
       .propertiesDefinition(initPropertiesDefinition())
       .apiDoc(new JHipsterModuleApiDoc("Init", "Init project"))
-      .organization(organization())
+      .organization(JHipsterModuleOrganization.builder().feature("startup").build())
```
hdurix, line 22: not sure about the feature, maybe this should have none

DamnClin: Yep, I think this is now a standalone module

hdurix: removed

DamnClin (dismissed):

PR #3392, opened 2022-09-04
Allow to customize Tikui cache folder

### LINKED ISSUE #3389: Tikui: rename cache to .cache
Is it possible to rename this folder to something else than `cache` ?
Here some suggestion:
- `.cache/`
- `.tikui/`

Or maybe directly inside target ?
- `target/cache/`
- `target/tikui/`

cc @Gnuk

Gnuk:
Resolves #3389

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ npm-debug.log.*
 ######################
 # Tikui cache
 ######################
-/cache
+/.tikui-cache/
 
 ######################
 # Fontello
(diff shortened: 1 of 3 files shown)


Gnuk:
The latest tikui core version cache is `.tikui-cache` by default, but it's also possible to change it using the `tikuiconfig.json` file, here is an example based on the JHipster one:

```json
{
  "documentation": "tikui",
  "src": "src/main/style",
  "dist": "target/classes/public/style",
  "cache": ".tikui-cache",
  "expose": {
    "src/main/glyph": "glyph"
  }
}
```

pascalgrimaud (approved):

PR #3434, opened 2022-09-07
Add navigation on Tikui

Gnuk:


DamnClin (approved):

Gnuk:
Seems to not be about the code because there is a 404 from codecov in the error. Maybe is it possible to launch the pipeline again ?

Issue #3531, opened 2022-09-14
Spring Boot core: missing spring-boot-starter-validation


pascalgrimaud:
When generating spring-boot application, there is a missing lib:

![image](https://user-images.githubusercontent.com/9156882/190096916-6dbba184-17e6-4d6d-8e83-62906d7e89ee.png)

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.365 s
[INFO] Finished at: 2022-09-14T10:02:36+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:testCompile (default-testCompile) on project beer: Compilation failure: Compilation failure: 
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[8,24] package javax.validation does not exist
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[9,24] package javax.validation does not exist
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[10,24] package javax.validation does not exist
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[20,26] cannot find symbol
[ERROR]   symbol:   class Validator
[ERROR]   location: class tech.jhipster.beer.BeanValidationAssertions.BeanAsserter<T>
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[21,23] cannot find symbol
[ERROR]   symbol:   class ConstraintViolation
[ERROR]   location: class tech.jhipster.beer.BeanValidationAssertions.BeanAsserter<T>
[ERROR] /tmp/35e03add-5f8c-4212-be40-13c7f72965db/src/test/java/tech/jhipster/beer/BeanValidationAssertions.java:[42,23] cannot find symbol
[ERROR]   symbol:   class ConstraintViolation
[...]

PR #3535, opened 2022-09-14
Move Bean Validation Test to Spring MVC

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/3531


src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
```diff
@@ -56,6 +57,15 @@ private JavaDependency springBootWebWithoutTomcatDependency() {
       .build();
   }
 
+  private JavaDependency reflectionsDependency() {
```
DamnClin, line 60: To move after `springMvcBuilder` to keep a reading logic order

DamnClin (approved):

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
pascalgrimaud: Thanks will do that

PR #3548, opened 2022-09-14
Use default property value for blank value

### LINKED ISSUE #3547: Use default property value for blank value
Following https://github.com/jhipster/jhipster-lite/issues/3424#issuecomment-1238406494, use default value when the property value is blank/empty.

hdurix:
- Fix #3547


hdurix:
Example of errors:
image


https://github.com/jhipster/jhipster-lite/actions/runs/5354941671/jobs/9712609209?pr=6647

Issue #3376, opened 2022-08-31
README: display sponsors


pascalgrimaud:
Add sponsors in README, like it's done in generator-jhipster

SoniaLahi:
I can take this issue

PR #3550, opened 2022-09-14
Display sponsors and contributing sections in README.md [skip ci]

SoniaLahi:
Fix https://github.com/jhipster/jhipster-lite/issues/3376

* I added the contributing section as well
* I omitted this line from the project `generator-jhipster`, since I couldn't see any difference but I may be wrong :smile: 
```md
<object data="https://opencollective.com/generator-jhipster/tiers/backer.svg?avatarHeight=40&width=890&button=false" type="image/svg+xml"></object>
```


pascalgrimaud:
thanks @SoniaLahi

Issue #2921, opened 2022-08-04
Hexagonal Arch: add new test - secondary can call primary from another context


pascalgrimaud:
I'm trying to improve the `HexagonalArchTest` and I saw this issue.

In `FileSystemJHipsterModulesRepository` which is in `secondary`, we import a class from `primary`. We should use a port instead I think. Can you have a look and suggest a fix plz @DamnClin ?

```java
package tech.jhipster.lite.module.infrastructure.secondary;

import org.springframework.stereotype.Repository;
import tech.jhipster.lite.error.domain.Assert;
import tech.jhipster.lite.module.domain.JHipsterModuleApplied;
import tech.jhipster.lite.module.domain.JHipsterModuleChanges;
import tech.jhipster.lite.module.domain.JHipsterModulesRepository;
import tech.jhipster.lite.module.domain.postaction.JHipsterModuleExecutionContext;
import tech.jhipster.lite.npm.domain.NpmVersions;
import tech.jhipster.lite.project.infrastructure.primary.JavaProjects; // <--- should not import a primary here
import tech.jhipster.lite.projectfile.domain.ProjectFilesReader;
```

DamnClin:
This is not a bug, it's totally fine! A secondary from one context can call a primary from another context. Here we are doing a java adaptation.

pascalgrimaud:
Oh I see, tell me if I'm correct:
- a secondary can't call a primary from the same context
- a secondary can call a primary from another context

DamnClin:
Yep

pascalgrimaud:
Then, I'm changing the ticket, because I wonder if we could write an arch unit test for that, I think it should not be so easy

DamnClin:
I think it's not really hard to do, loop other the context packages and ensure that the secondary package have no dependency to the primary one

hdurix:
maybe the test could ensure that only primary classes containing **`Java`** could be called from a secondary package?

pascalgrimaud:
Adding a bounty as it's more work than expected, and need to think deeply to provide the good solution here

Issue #3540, opened 2022-09-14
Suggestion: rename BusinessContext to BoundedContext


pascalgrimaud:
suggested by @onepunchmagne 

cc @DamnClin : what do you think?

DamnClin:
I don't like it since it is not a BoundedContext so I don't really get the idea here

pascalgrimaud:
The discussion with @onepunchmagne was related to this : https://github.com/jhipster/jhipster-lite/tree/main/src/main/java/tech/jhipster/lite/technical

As it can look strange to have a BusinessContext package-info in a Technical package

pascalgrimaud:
No other comment since months
So I'm fine to keep BusinessContext. Anyway, it can be renamed easily in the generated project, so no problem for that

Let's close this, as we have more important ticket

PR #3552, opened 2022-09-14
Redefining ArchUnit tests

onepunchmagne:
Fix [#2921](https://github.com/jhipster/jhipster-lite/issues/2921)

* Shared kernels cannot depend on bounded contexts
* Bounded context can calls each others only using secondary to primary Java adapters communications
* Domain cannot use anything but its self-domain, Java vanilla features or some common utils libraries
* Application cannot call infrastructure
* Primary cannot directly call secondary
* Secondary cannot call any application service

### DIFF
--- a/src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
+++ b/src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
@@ -5,16 +5,19 @@ import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.*;
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.importer.ClassFileImporter;
 import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.library.Architectures;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Function;
 import java.util.stream.Stream;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RestController;
 
@@ -27,24 +30,28 @@ class HexagonalArchTest {
     .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
     .importPackages(ROOT_PACKAGE);
 
-  private static final Collection<String> businessContexts = buildBusinessContexts();
+  private static final Collection<String> businessContexts = packagesWithAnnotation(BusinessContext.class);
+  private static final Collection<String> businessContextsPackages = buildPackagesPatterns(businessContexts);
 
-  private static final Collection<String> commonPackages = Arrays.asList("java..", "org.slf4j..", "org.apache.commons.lang3..");
+  private static final Collection<String> sharedKernels = packagesWithAnnotation(SharedKernel.class);
+  private static final Collection<String> sharedKernelsPackages = buildPackagesPatterns(sharedKernels);
 
-  private static final Collection<String> sharedKernelsPackages = buildSharedKernelsPackages();
+  // the empty package is related to: https://github.com/TNG/ArchUnit/issues/191#issuecomment-507964792
+  private static final Collection<String> vanillaPackages = List.of("java..", "");
+  private static final Collection<String> commonToolsAndUtilsPackages = List.of(
+    "org.slf4j..",
+    "org.apache.commons..",
+    "com.google.guava.."
+  );
 
-  private static Collection<String> buildBusinessContexts() {
-    return getPackageAnnotatedBy(BusinessContext.class);
+  private static Collection<String> buildPackagesPatterns(Collection<String> packages) {
+    return packages.stream().map(path -> path + "..").toList();
   }
 
-  private static Collection<String> buildSharedKernelsPackages() {
-    return getPackageAnnotatedBy(SharedKernel.class).stream().map(path -> path + "..").toList();
-  }
-
-  private static Collection<String> getPackageAnnotatedBy(Class<? extends Annotation> annotationClass) throws AssertionError {
+  private static Collection<String> packagesWithAnnotation(Class<? extends Annotation> annotationClass) throws AssertionError {
     try {
       return Files
-        .walk(Paths.get("src", "main", "java", {{ packageWalkPath }}))
+        .walk(Paths.get("src", "main", "java", {{packageWalkPath}}))
         .filter(path -> path.toString().endsWith("package-info.java"))
         .map(toPackageName())
         .map(path -> path.replaceAll("[\\/]", "."))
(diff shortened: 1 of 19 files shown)


pascalgrimaud:
Does it fix #2921 @onepunchmagne ?

onepunchmagne:
Work left:
* a bug fixed in ArchUnit 1.0.0-rc1 seems to be still an issue for check primary Java adapters are called only by secondaries to fix #2921
* @DamnClin @pascalgrimaud I have a problem understanding why `module` is a shared kernel and `project` a bounded context -> we have an architecture issue on `FileSystemJHipsterModulesRepository` calling `JavaProjects`

DamnClin:
> I have a problem understanding why module is a shared kernel and project a bounded context

Not sure to understand the question so I can only answer by saying that module is a shared kernel and project is a business context (I don't think we have a bounded context annotation as this would be really wrong and missleading !)

> we have an architecture issue on FileSystemJHipsterModulesRepository calling JavaProjects

This is not a bug, JavaProject is a primary adapter, it's totally normal

onepunchmagne:
> I don't think we have a bounded context annotation as this would be really wrong and missleading !

Here is the thing that should be discussed, the annotation is called `@BusinessContext` but the tests say otherwise, here is the behavior implemented in the existing tests: `"Shared kernels should not have dependencies to bounded contexts"`  
Moreover in the `hexagonal-architecture.md` we explain in the same paragraph:  

> So, first things first: **an application is made of multiple "hexagons"**, one for each [Bounded Context](https://martinfowler.com/bliki/BoundedContext.html). (Yes, sometimes you can have only one but this is an exception). We usually have each Bounded Context as root packages in the application.
> - `my_business_context`: root package for the context (naming depends on your technology naming conventions)  

I think @pascalgrimaud and @hdurix aggried that JHLite generates `bounded contexts` from DDD, but this is very confusing in JHLite code itself. As a user of JHLite our project team also understood that `@BusinessContext` annotation was from "bounded context" (without understanding why it was not named exactly after the DDD strategic patterns). Also the other annotation coming in pair is the `@SharedKernel` so it made perfectly sense for JHLite **users** (not contributors). But for contributors point of view, it is indeed misleading, I can't understand our `package-info.java` descriptors.

If I may, I would suggest we rename to `@BoundedContext` explicitely to stick to a well-documented notion that would wear a conviction along with `@SharedKernel`. But that we remove/redefine the existing `package-info.java` that seems to randomly wear the annotations. (ticket is here https://github.com/jhipster/jhipster-lite/issues/3540)

DamnClin:
Ok, if I'm the only one finding that misleading let's rename it

src/main/java/tech/jhipster/lite/project/package-info.java
```diff
@@ -1,2 +1,2 @@
-@tech.jhipster.lite.BusinessContext
+@tech.jhipster.lite.SharedKernel
```
DamnClin, line 1: Project is not a shared kernel, I don't understand this change

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -27,24 +32,28 @@ class HexagonalArchTest {
     .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
     .importPackages(ROOT_PACKAGE);
 
-  private static final Collection<String> businessContexts = buildBusinessContexts();
+  private static final Collection<String> businessContexts = getPackageAnnotatedBy(BusinessContext.class);
```
DamnClin, line 35: getPackageAnnotatedBy -> packagesWithAnnotation ?

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -27,24 +32,28 @@ class HexagonalArchTest {
...
+  private static final Collection<String> businessContextsPackages = toPackages(businessContexts);
```
DamnClin, line 36: toPackages -> buildPackagesPatterns?

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -27,24 +32,28 @@ class HexagonalArchTest {
...
 
-  private static final Collection<String> commonPackages = Arrays.asList("java..", "org.slf4j..", "org.apache.commons.lang3..");
+  private static final Collection<String> sharedKernels = getPackageAnnotatedBy(SharedKernel.class);
+  private static final Collection<String> sharedKernelsPackages = toPackages(sharedKernels);
 
-  private static final Collection<String> sharedKernelsPackages = buildSharedKernelsPackages();
+  // the empty package is related to: https://github.com/TNG/ArchUnit/issues/191#issuecomment-507964792
+  private static final Collection<String> vanillaPackages = Arrays.asList("java..", "");
```
DamnClin, line 42: Arrays.asList -> List.of (not allowing null values)

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -27,24 +32,28 @@ class HexagonalArchTest {
...
+  private static final Collection<String> commonToolsAndUtilsPackages = Arrays.asList(
```
DamnClin, line 43: Arrays.asList -> List.of

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -76,147 +85,170 @@ class HexagonalArchTest {
...
+        String[] otherContexts = otherContextsDomains(context);
 
-      noClasses()
+        noClasses()
+          .that()
+          .resideInAnyPackage(context + "..")
+          .should()
+          .dependOnClassesThat()
+          .resideInAnyPackage(otherContexts)
+          .because("Contexts can only depend on classes in the same context or in common")
```
DamnClin, line 102: or in common -> or in shared kernels?

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -76,147 +85,170 @@ class HexagonalArchTest {
...
+          .check(classes);
+      });
+    }
+
+    @Test
+    void shouldBeAnOnionArchitecture() {
```
DamnClin, line 108: Hum no, this is an hexagonal architecture test, not an onion architecture test

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -76,147 +85,170 @@ class HexagonalArchTest {
...
+      businessContexts.forEach(context -> {
+        Architectures
+          .onionArchitecture()
+          .withOptionalLayers(true)
+          .domainModels(context + ".domain..")
+          .domainServices(context + ".domain..")
+          .applicationServices(context + ".application..")
+          .adapter("infrastructure", context + ".infrastructure..", context + ".infrastructure.secondary..")
+          .because("Each bounded context should implement an onion architecture")
```
DamnClin, line 117: Totally disagree with that. We should make a dedicated module if we want to enforce onion architecture

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -76,147 +85,170 @@ class HexagonalArchTest {
...
-    businessContexts.forEach(context ->
+  @Nested
+  class Domain {
+
+    @Test
+    void shouldNotDependOnOutside() {
       classes()
         .that()
-        .resideInAnyPackage(context + ".domain..")
+        .resideInAPackage(".domain..")
```
DamnClin, line 170: I don't get it, here we are allowing domain to depend on any domain no? No more check for domain dependencies to other domains?

DamnClin (changes requested):

DamnClin:
This can give some god hints: https://github.com/whiskeysierra/archunit-hexagonal/blob/trunk/src/main/kotlin/io/github/whiskeysierra/archunit/hexagonal/HexagonalArchitecture.kt

onepunchmagne:
> This can give some god hints: https://github.com/whiskeysierra/archunit-hexagonal/blob/trunk/src/main/kotlin/io/github/whiskeysierra/archunit/hexagonal/HexagonalArchitecture.kt

Thanks for the correction you're right, I used the hint trying to define some basics rules, I get that it's hard to define "the" hexagonal architecture test but it should probably serve as a good base for people to update according to their definition.

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
onepunchmagne: Yes but there were a lot of overlaying tests so I tried defining different separate responsibility tests such as:  
* `BoundedContexts.shouldNotDependOnOtherBoundedContextDomains` which will verify our hexagons co-existence and boundaries
* `Domain.shouldNotDependOnOutside` which does not even know it co-exists with other hexagons, it simply check its dependencies to "outside" of a domain package

src/main/java/tech/jhipster/lite/project/package-info.java
onepunchmagne: If possible I'd like to take also this job but in issue https://github.com/jhipster/jhipster-lite/issues/3540  
because sticking to the architecture tests, we have a failure because `project` is called directly by `module` which is tagged as a shared kernel.

To be honest I have absolutely no clue currently of which package is a shared kernel and which is a "business" context, but there is something to do in this separate issue I think to realign the dependencies.

DamnClin: This is a mistake :D project is a business context and module is a SharedKernel, what is the link?

onepunchmagne: Ok, so the whole thing about `SharedKernels.shouldNotDependOnBoundedContexts` is wrong I suppose, I delete it. A shared kernel can actually call bounded contexts following the same rule as passing through its secondary adapter to call the bounded context primary adapter, is that right?

Thus the `shouldBeAnHexagonalArchitecture` test should also be applied to shared kernel in the exact same way. Please check new change where I applied this logic

src/main/resources/generator/server/javatool/arch/HexagonalArchTest.java.mustache
```diff
@@ -1,20 +1,24 @@
 package {{packageName}};
 
-import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.*;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
```
DamnClin, line 3: Can you change your imports to import * starting at first static import?

DamnClin (dismissed):

onepunchmagne:
Claimed https://opencollective.com/generator-jhipster/expenses/114216

DanielFran:
@onepunchmagne approved

PR #3600, opened 2022-09-19
Build normal jar for deployment

DamnClin:
cc @pascalgrimaud this is building the a "classic" jar without the profile BUT I can't tell if there are other impacts to this, leaving this to you...


pascalgrimaud:
I'll test and do a release to check how it works :)

Issue #3484, opened 2022-09-11
Allow modules deactivation


DamnClin:
Related to https://github.com/jhipster/jhipster-lite/issues/3482 adding a configuration to disable / enable given modules can be a good idea so users can remove some modules if needed

pascalgrimaud:
this ticket would be the perfect use for disabling the different banner, so no need to remove the existing banner modules, just hide them

PR #3591, opened 2022-09-18
Custom JHLite instance module

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/3482

Issue #3049, opened 2022-08-12
Project banner module


DamnClin:
Add a module to create a banner using the project short name (in ASCII art ofc)

MrMattam:
This has been done in pair with @hdurix : thank you for the onboarding :)

PR #3602, opened 2022-09-20
Modules deactivation

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/3484


DamnClin:
Waiting for https://github.com/jhipster/jhipster-lite/pull/3591 to add documentation in custom JHLite

pascalgrimaud (approved):
could you disable all the banner modules too plz?

DamnClin:
> could you disable all the banner modules too plz?

Why??? If we don't want a module in the main JHLite instance we should delete it! For me we need https://github.com/jhipster/jhipster-lite/issues/3049

pascalgrimaud:
Oh last point: I was suggesting to disable it because it will reduce the size of the tree, that's all :-D

DamnClin:
Ok, I'll disable it with a tag so

PR #3635, opened 2022-09-22
Upgrade Angular 14.2.3

qmonmert:
Fix https://github.com/jhipster/jhipster-lite/pull/3628


DamnClin (approved):

pascalgrimaud:
@qmonmert @DamnClin : I wonder if we can find a way to use one single lib in package.json and use the same version for all the others, so dependabot will upgrade 1 single lib, instead of several

PR #3682, opened 2022-09-27
Move technical exception package

DamnClin:


pascalgrimaud:
maybe could be done for the generated app toO?

DamnClin:
> maybe could be done for the generated app toO?

Yep but in dedicated ticket, I'll see to delete the technical package from JHLite and do the same in the generated app after. My main concern here is that this error management should be totally reworked but no mojo to do it right now :D

pascalgrimaud (approved):

PR #3814, opened 2022-10-04
frontend-maven-plugin depends on client-core

### LINKED ISSUE #3810: frontend-maven-plugin should depend on client-core
as if you use frontend-maven-plugin alone (init -> maven-java -> frontend-maven-plugin), it will be broken

AnneJacquet:
closes #3810

❓I am surprised about 2 things : 
- haven't found any test covering that part
- all the magic string

I am not sure this is the correct place for those questions and if they have already been answered 🐣


DamnClin:
There are no test for module dependencies since this is more a configuration than anything else, testing that will be the same as doing it (but if you have a bad organization the app just don't start and you'll have a bunch of failing tests).

We chose to stick to strings for modules and features name to ease custom jhlite instances management but we may go something else in the future (like an interface with only enum implementations, dunno)

pascalgrimaud (approved):

pascalgrimaud:
Thanks @AnneJacquet :-)

PR #3816, opened 2022-10-04
Feature/refactor example default

### LINKED ISSUE #3480: Default value for module properties
We should switch "example" to "default" in module properties and use that to valuate fields if no user input can be found

swarajsaaj:
Fill default values for properties on application of patch/modules

Fix #3480


src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulePropertyDefinition.java
```diff
@@ -117,8 +113,8 @@ public Optional<JHipsterPropertyDescription> description() {
     return description;
   }
 
-  public Optional<JHipsterPropertyExample> example() {
```
DamnClin, line 120: Need to keep it deprecated at least for 1 version. See JHipsterModuleReplacements.JHipsterModuleReplacementsBuilder.in for deprecation example.

Doing that since it can be used in custom JHLite instances

src/main/webapp/app/module/primary/landscape/Landscape.component.ts
```diff
@@ -335,6 +337,21 @@ export default defineComponent({
...
+      selectedModulesProperties().forEach(property => {
+        if (unknownProperty(property.key) && property.defaultValue) {
+          switch (property.type) {
+            case 'INTEGER':
+              updateProperty(createProperty(property.key, Number.parseInt(property.defaultValue)));
+              break;
+            case 'BOOLEAN':
+              updateProperty(createProperty(property.key, Boolean(property.defaultValue)));
+              break;
+            default:
```
DamnClin, line 349: You can avoid default and have code not compiling without all values by putting that in a dedicated method. see `modeClass` in Landscape.component.ts for an example

src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
```diff
@@ -205,6 +205,21 @@ export default defineComponent({
     const applyModule = (module: string): void => {
       operationInProgress.value = true;
 
+      selectedModuleProperties().forEach(property => {
```
DamnClin, line 208: This looks a lot like a duplication :)

src/main/webapp/app/module/primary/landscape/Landscape.component.ts
swarajsaaj: Sure, I would rather do a common util method in PropertyValue.ts for use in both Landscape and PathModule

src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
swarajsaaj: :), would reuse the common method from PropertyValue.ts for casting logic.

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulePropertyDefinition.java
swarajsaaj: @DamnClin Sorry I am not really getting this one, what do you mean by "custom JHLite instances" here and how would they be affected by this change? 
Is it for JHLite being used as a dependency in other project? Not sure how it affects generated apps.

DamnClin: Yep. You have a module called custom-jhlite that generate instances allowing organization to build there own module. In this we have dependencies to JHLite so we need to give them some time to adapt to new apis

swarajsaaj: Thanks, i have kept the example in here and builder as deprecated.

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulePropertyDefinition.java
```diff
@@ -117,8 +114,20 @@ public Optional<JHipsterPropertyDescription> description() {
...
   }
 
+  /**
+   *  @deprecated use {@link #defaultValue()} instead
+   */
+  @Deprecated(forRemoval = true)
+  @Generated(reason = "Candidate for deletion")
   public Optional<JHipsterPropertyExample> example() {
-    return example;
+    if (defaultValue.isPresent()) {
```
DamnClin, line 123: defaultValue.map(value -> JHipsterPropertyExample.of(value.get()));

swarajsaaj: Thanks, I see it has to be `flatMap` , else it would return `Optional<Optional<JhipsterPropertyExample>>`, so:-

`
defaultValue.flatMap(value -> JHipsterPropertyExample.of(value.get()));
`

DamnClin (approved):

swarajsaaj:
Had to add a test case for deprecated method, my assumption was its ignored from Codecov.

DamnClin:
Good job here. cc @pascalgrimaud I think this deserve a Bounty :)

Issue #3779, opened 2022-10-03
Allow CSS refresh without having to do mvn clean


DamnClin:
With the css management update you need to clean and restart the server to see css updates (same using vite), this is really annoying. cc @Gnuk

Gnuk:
I think there is two features behind this:

- Launch tikui + vite in dev mode and add proxy to vite to have the tikui served style instead of JHipster served style
- Change maven configuration to accept build even if it's not clean

About the second point:  I don't really know and have an idea on how to do it and I also don't know if it's the same behavior with vite build for the webapp (need to check).

I may do the first one but I'll be glad to let someone else to do the other one.

PR #3862, opened 2022-10-07
Launch both vite and tikui in dev mode

Gnuk:
* Linked to #3779

### DIFF
--- a/tikuiconfig.json
+++ b/tikuiconfig.json
@@ -2,6 +2,7 @@
   "documentation": "tikui",
   "src": "src/main/style",
   "dist": "target/classes/public/style",
+  "port": 9005,
   "expose": {
     "src/main/glyph": "glyph"
   }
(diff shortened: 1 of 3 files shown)


Gnuk:
Maybe this need to check if it's working on Windows before merging this :D

tikuiconfig.json
```diff
@@ -2,6 +2,7 @@
   "documentation": "tikui",
   "src": "src/main/style",
   "dist": "target/classes/public/style",
+  "port": 9001,
```
pascalgrimaud, line 5: is it possible to use another port? as this one is used for the sonar instance
I don't know which port we can use, maybe use one which is not listed here https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

pascalgrimaud (approved):

tikuiconfig.json
Gnuk: Ok, I just put 9005, please don't hesitate to challenge me.

Issue #3005, opened 2022-08-10
Rework page loading


DamnClin:
Loading pages now just display a "Loading" text which is not good and comes with a blinking effect when changing page, adding placeholder during loading may be a good idea

swarajsaaj:
The blinking effect is mainly due to anchor `href` which is causing the page to reload on each header click.
https://github.com/jhipster/jhipster-lite/blob/6d2317c1572001783f225ce004edcc38b8b6f30b/src/main/webapp/app/common/primary/header/Header.vue#L14

Will provide a PR for this small fix of blinking effect.

I have added a plain css loader as in https://cssloaders.github.io/, but the issue can be kept open till actual placeholders are implemented, something like https://www.npmjs.com/package/placeholder-loading

swarajsaaj:
@pascalgrimaud this issue should be reopened as #3868 only partially fixes the issue

Issue #2894, opened 2022-08-02
UX/UI: improve the CSS


pascalgrimaud:
Nothing again you, @DamnClin :-D
The CSS can be improved

DamnClin:
What???? This lovely handmade with heart shitty page can be improved? In fact it's more a must be than a can be (but I can't do it, totally impossible for me)...

Gnuk:
I may be interested to contribute to this but with something like Tikui integration before. What do you think about it ?

pascalgrimaud:
I don't have any objection to use Tikui, and it would be very good use case :)

But, what to keep in mind:
- the application JHLite should work locally, without the need of external server (nginx or something else)
- the application JHLite should work offline
- it means, the CSS should be directly included in the project

I don't know what would be better:
- create a new repository ? (I'd like to avoid if possible)
- include it directly in the project, but where in the structure ?

Plz, don't start something before we all agree, I don't wanna waste your time

Gnuk:
The idea I have is to includes Tikui as a module and have the style in the project structure and serve the style like vue/angular/react are served. The only thing I need to achieve this is to contribute to Tikui to let use another directory for sources and generation (will be set by default to src and dust like today but could be overridden by a configuration to src/main/... )

pascalgrimaud:
Link for reference:
- Design: https://github.com/jhipster/jhipster-lite/issues/1565
- Landscape: https://github.com/jhipster/jhipster-lite/pull/3022

Gnuk:
Maybe you're interested about this issue @ascelineboullen

pascalgrimaud:
Closing in favor of https://github.com/jhipster/jhipster-lite/issues/10576

PR #3868, opened 2022-10-08
Remove href in header section and add loader css

swarajsaaj:
The header has href which causes full page reloads, this is refactored to router-link and a css loader added instead of text.

Related to #3005 (which is larger issue for adding placeholder loading)

### DIFF
--- a/src/main/webapp/app/assets/loader.scss
+++ b/src/main/webapp/app/assets/loader.scss
@@ -1,0 +1,25 @@
+.jhipster-loader-container {
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+.jhipster-loader {
+  width: 48px;
+  height: 48px;
+  border: 5px solid #fff;
+  border-bottom-color: transparent;
+  border-radius: 50%;
+  display: inline-block;
+  box-sizing: border-box;
+  animation: rotation 1s linear infinite;
+}
+
+@keyframes rotation {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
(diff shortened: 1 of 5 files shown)


pascalgrimaud:
Cc @Gnuk : if you can have a look plz

Gnuk:
I suggest to move styles in Tikui. Maybe it would also be interesting to not forget the examples parts like other components from the pattern library.

To watch it in dev mode, please go to http://localhost:9000/style/index.html or if you launch just tikui : http://localhost:9005/.

Gnuk:
I agree to not close or create another issue with a "placeholder" to indicate the user what will look like the page after a loading.

swarajsaaj:
Thanks @Gnuk ,
moved the styles to tikui along with example documentation in [3b4e5e2](https://github.com/jhipster/jhipster-lite/pull/3868/commits/3b4e5e25f278e4feb85a5861918c658c50c721a7) (This is first time I tried tikui so please let me know if something needs to be done differently :) )

src/main/style/molecule/loader-container/loader-container.mixin.pug
```diff
@@ -0,0 +1,5 @@
+include /atom/loader/loader.mixin.pug
+
+mixin loader-container
+  .jhlite-loader-container
+    .jhlite-loader
```
Gnuk, line 5: Maybe use a `+jhlite-loader` here and include your mixin before line 1.

Gnuk (dismissed):
I may just need to remember to change the color later because the base theme will be light ;) from the #2894

pascalgrimaud:
As soon as last comment is fixed, I'll merge this one

src/main/style/molecule/loader-container/loader-container.mixin.pug
swarajsaaj: Thanks. done

PR #3754, opened 2022-10-01
Add triage label

swarajsaaj:
Add 'area: triage' label to unlabelled PRs and issues

Uses https://github.com/marketplace/actions/simple-issue-labeler for simple issue labeling for unlabelled issues and PRs

Fix #3281

Issue #3281, opened 2022-08-28
CI: add auto label


pascalgrimaud:
When there is no label in issue and in pull request, add automatically :
- area: triage

PR #3870, opened 2022-10-08
Fix failing PR labelling

swarajsaaj:
Currently PRs are not being labelled for triage (due to missing write permission in actions)

pull_request_target as event type gives permissions required to label PR

Following issue #3281 and PR #3754


swarajsaaj:
Reference:- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

pascalgrimaud:
Thanks @swarajsaaj

Issue #3893, opened 2022-10-10
Vitest: clean reference to jest


pascalgrimaud:
In vitest config, there are still some reference to jest. Maybe it can impact the sonar configuration too

pascalgrimaud:
Can be closed

PR #3933, opened 2022-10-12
Vitest: clean reference to jest

qmonmert:
Related to #3893 

I think we can't clean others references without adding `if` in modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "node_modules"
   ],
   "jestSonar": {
-    "reportPath": "target/test-results/jest",
+    "reportPath": "target/test-results",
     "reportFile": "TESTS-results-sonar.xml"
   }
 }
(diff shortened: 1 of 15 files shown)


pascalgrimaud:
I think we should remove jest in these lines and use a common one or simply remove jest:
- https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/client/svelte/vitest.config.ts#L18
- https://github.com/jhipster/jhipster-lite/blob/main/sonar-project.properties#L14

So it won't impact angular as it is still using jest

So are you talking about these lines?

qmonmert:
yes and all references like https://github.com/jhipster/jhipster-lite/blob/67575e9bcb93ffb76d9ed96778f96b13ada306b1/src/main/resources/generator/init/.eslintignore#L5

pascalgrimaud:
for config file like ignore, we can keep both (jest + vitest) I think

qmonmert:
@pascalgrimaud I updated PR

Last thing https://github.com/jhipster/jhipster-lite/blob/2da59fac79239a25bb7a0a565ca41177217b69dd/src/main/java/tech/jhipster/lite/generator/client/common/domain/ClientsModulesFactory.java#L21 `jestSonar` property is only necessary for Angular but generated in all front apps

pascalgrimaud (approved):

Issue #3981, opened 2022-10-15
Format AnnotationArchTest


DamnClin:
`AnnotationArchTest` (in JHLite and module) formatting can be improved for better readability

PR #3983, opened 2022-10-15
Improved formatting of AnnotationArchTest for better readability

murdos:
Draft for fixing #3981

If we agree on the formatting, I can report the changes on the template of module


DamnClin:
I think it's better yes, thanks

PR #3992, opened 2022-10-15
Angular: standalone components

qmonmert:
@wmarques I started to migrate to standalone components (not all components again). Do you think that we must wait Angular 15 or we can merge it? JHipster Lite is not in version 1.0.0 so 🤷‍♂️

https://angular.io/guide/standalone-components

cc @pascalgrimaud


wmarques:
I think we can go ahead the API looks stable enough

qmonmert:
@wmarques 👍 I agree

pascalgrimaud:
As I'm not an Angular user, I'm trusting you both, William and Quentin :-)

src/main/resources/generator/client/angular/admin/src/main/webapp/app/admin/health/modal/health-modal.component.ts.mustache
```diff
@@ -1,10 +1,17 @@
...
 
+import { MatIconModule } from '@angular/material/icon';
+import { MatButtonModule } from '@angular/material/button';
+import { MatTableModule } from '@angular/material/table';
+import { MAT_DIALOG_DATA, MatDialogModule } from '@angular/material/dialog';
+
 @Component({
   selector: 'jhi-health-modal',
   templateUrl: './health-modal.component.html',
+  imports: [CommonModule, HealthModalComponent, MatDialogModule, MatIconModule, MatButtonModule, MatTableModule],
```
wmarques, line 13: I don't think you have to reimport the current component

wmarques (commented):
Done, very good job :) only a few comments but LGTM

qmonmert:
@wmarques thanks, yes you're right I updated 🙂

PR #4092, opened 2022-10-24
Add a default error page for tomcat and undertow

### LINKED ISSUE #3803: Default error page with server app
When using tomcat, undertow or netty, we should add a default error page similar to what generator-jhipster provides

Yachef:
Closes #3803


Yachef:
Not sure what to do about Netty, as the 404 error is already caught :
image


Screenshot for Undertow (similar on Tomcat, just the "Message" changes) : 
image

pascalgrimaud:
Thanks and congrats for your 1st contribution @Yachef 
No worry for Netty, it can be done in another PR

I'll try yo have a look :-)

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
```diff
@@ -66,15 +67,18 @@ private JHipsterModuleBuilder springMvcBuilder(JHipsterModuleProperties properti
       .localEnvironment(localEnvironment("- [Local server](http://localhost:"+properties.serverPort().get()+")"))
       .javaDependencies()
         .addDependency(SPRING_BOOT_GROUP, artifactId("spring-boot-starter-validation"))
+        .addDependency(SPRING_BOOT_GROUP, artifactId("spring-boot-starter-thymeleaf"))
```
DamnClin, line 70: I'm not a fan of adding thymeleaf just for that... is it a good idea to make a dedicated module for this 404? @jdubois I know you want this, what's your advice on that?

pascalgrimaud: Your point of view is good @DamnClin 
Some people don't care about the default 404 error page, so adding thymeleaf dependency for having this page is maybe overkill.
As JHLite is modular, it can be done in a dedicated module as you suggested

Yachef: Yes, I also think that it's overkill.

If we only care about the 404 error, and if a static HTML file is enough, we can also do it using Spring boot static resources feature : 

https://docs.spring.io/spring-boot/docs/2.7.5/reference/html/web.html#web.servlet.spring-mvc.error-handling.error-pages

We don't need any dependency to do that.

DamnClin: Looks better indeed

Yachef: Updated the PR with few UI changes : 
image

DamnClin:
I like this version! Less information bu far less noise for people who just want to expose WebServices

DamnClin (approved):

PR #4131, opened 2022-10-26
add landscape scroll on click

### LINKED ISSUE #4132: Add landscape scroll by mouse grabbing
The idea is to allow the user to scroll vertically and horizontally the landscape view with left mouse click

Franceq34:
Closes #4132

### DIFF
--- a/src/main/resources/generator/dependencies/Dockerfile
+++ b/src/main/resources/generator/dependencies/Dockerfile
@@ -1,5 +1,5 @@
 FROM sonarqube:9.7.1-community
-FROM consul:1.14.1
+FROM consul:1.14.2
 FROM jhipster/consul-config-loader:v0.4.1
 FROM jhipster/jhipster-registry:v7.4.0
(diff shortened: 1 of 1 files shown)


Franceq34:
I've got something **working** here, I would like some advices on the choices made, maybe opinions on my Bus/Listener design.

Also, I fought against unit tests because I don't have access to the data `landscapeContainer` so I didn't figure out how to cover this part in tests:

`const grabbing = (mouseEvent: MouseEvent): void => {
        if (!isMoving.value) {
          return;
        }
        if (!landscapeContainer.value) {
          return;
        }
        landscapeContainer.value.scroll(
          currentScrollX.value + (startX.value - mouseEvent.clientX),
          currentScrollY.value + (startY.value - mouseEvent.clientY)
        );
      };`
 
in `src/main/webapp/app/module/primary/landscape/Landscape.component.ts`

Maybe @Gnuk or @nonomoho you could bring some help ?

PR #4161, opened 2022-10-28
make annotation attribute override explicit

biergit:
This fixes the warning
> AnnotationTypeMapping   : Support for convention-based annotation attribute overrides is deprecated and will be removed in Spring Framework 6.1. Please annotate the following attributes in @tech.jhipster.lite.IntegrationTest with appropriate @AliasFor declarations: [properties]


pascalgrimaud:
Thanks for your contribution @biergit :)

PR #4169, opened 2022-10-29
Front: label + field not well displayed

### LINKED ISSUE #4151: Front: label + field not well displayed
When editing a field, the label is not well displayed.
It's minor but I think it can be improved

image

qmonmert:
Fix #4151 

image

@pascalgrimaud what do you think? Simple but efficient


pascalgrimaud:
Yes looks good to me

PR #4175, opened 2022-10-30
zalando problems with spring-boot-webflux-netty does not compile

### LINKED ISSUE #4111: zalando problems with spring-boot-webflux-netty does not compile
`ExceptionHandler` requires `HttpServletRequest` which is not available.

Yachef:
- Remove zalando from webflux module

- Remove zalando in the whole project

Closes #4111


Yachef:
I was not able to remove zalando only on webflux, otherwise security (JWT & OAuth2) would not compile, nor work.

So I took the initiative to remove it from the whole project.

However I'm not really sure about my ControllerAdvice implementations (web & webflux). They are pretty basic, and the status code for the web version is not as accurate as the one provided by Spring by default (I could not find the implementation Spring uses).

Also : I think webflux + (JWT or OAuth2) does not compile on main branch right now

DamnClin:
I think we should make and actual implantation for that after spring boot 3 migration! This don't really looks on for now, can you put the PR in draft ?

pascalgrimaud:
no need any more as zalando module has been deleted

Issue #4149, opened 2022-10-27
Issues after using custom-jhlite module


pascalgrimaud:
After using the `custom-jhlite` module, we can't apply new modules.

Steps to reproduce the issue:
- `git clone https://github.com/avdev4j/entando-jhipster-lite`
- start JHLite project
- put the correct folder (where the `entando-jhipster-lite` project was cloned)
- try to apply `prettier` module or `jacoco-check-min-coverage` for example

![image](https://user-images.githubusercontent.com/9156882/198223027-c3cf275f-862e-4f92-b2b9-8cd54c39e8dd.png)

I think it's something related to:

```xml
    <dependency>
      <groupId>tech.jhipster.lite</groupId>
      <artifactId>jhlite</artifactId>
      <version>${jhlite.version}</version>
      <classifier>tests</classifier>
      <scope>test</scope>
      <type>test-jar</type>
    </dependency>
```
Here the stacktrace:

```
2022-10-27 09:46:04.592 ERROR 69736 --- [nio-7471-exec-1] o.z.problem.spring.common.AdviceTraits   : Internal Server Error

java.lang.IllegalStateException: Duplicate key DependencyId[groupId=GroupId[groupId=tech.jhipster.lite], artifactId=ArtifactId[artifactId=jhlite]] (attempted merging values tech.jhipster.lite.module.domain.javadependency.JavaDependency@e38b2385 and tech.jhipster.lite.module.domain.javadependency.JavaDependency@b8fb1e31)
	at java.base/java.util.stream.Collectors.duplicateKeyException(Collectors.java:135)
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:182)
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base/java.util.AbstractList$RandomAccessSpliterator.forEachRemaining(AbstractList.java:720)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
[...]

PR #4225, opened 2022-11-01
Add classifier in dependency ID

DamnClin:
Fix https://github.com/jhipster/jhipster-lite/issues/4149

Tried different things for this but... I really think the classifier should be part of the DependencyID, that totally make sense with my current understanding of maven dependencies (but I may be totally wrong here!)

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/javadependency/DependencyId.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/javadependency/DependencyId.java
@@ -1,12 +1,14 @@
 package tech.jhipster.lite.module.domain.javadependency;
 
+import java.util.Optional;
 import tech.jhipster.lite.error.domain.Assert;
 import tech.jhipster.lite.module.domain.javabuild.ArtifactId;
 import tech.jhipster.lite.module.domain.javabuild.GroupId;
 
-public record DependencyId(GroupId groupId, ArtifactId artifactId) {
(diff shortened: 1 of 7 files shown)


src/main/java/tech/jhipster/lite/module/domain/javadependency/DependencyId.java
```diff
@@ -1,12 +1,14 @@
 package tech.jhipster.lite.module.domain.javadependency;
 
+import java.util.Optional;
 import tech.jhipster.lite.error.domain.Assert;
 import tech.jhipster.lite.module.domain.javabuild.ArtifactId;
 import tech.jhipster.lite.module.domain.javabuild.GroupId;
 
-public record DependencyId(GroupId groupId, ArtifactId artifactId) {
+public record DependencyId(GroupId groupId, ArtifactId artifactId, Optional<JavaDependencyClassifier> classifier) {
```
hdurix, line 8: maybe add a default constructor with empty classifier as you call it a lot with empty parameter.

hdurix (approved):

src/main/java/tech/jhipster/lite/module/domain/javadependency/DependencyId.java
DamnClin: Did that a the beginning but I don't think this is a good idea since it really hides that

PR #4267, opened 2022-11-06
rename sonar modules to sonar-qube

### LINKED ISSUE #4241: Rename sonar* modules to sonar-qube*
Sonar modules should be SonarQube modules. Files to update: 

* JHLiteModuleSlug
* SonarModulesFactory
* SonarModulesFactoryTest
* SonarApplicationService
* SonarModuleConfiguration

kartiktayal:
rename sonar modules to sonar-qube modules

Fix #4241


src/main/java/tech/jhipster/lite/generator/server/sonar/application/SonarQubeApplicationService.java
```diff
@@ -1,18 +1,18 @@
 package tech.jhipster.lite.generator.server.sonar.application;
```
DamnClin, line 1: package needs to be renamed

kartiktayal: Done

kartiktayal:
Not sure if this is an issue, but I have already signed the CLA twice, still it doesn't seem to be applying and is in pending state. Please let me know if I am missing something here.

pascalgrimaud:
You need to check if your 2 commits used the same email in your GitHub account

What I can see (from my phone):
- your GitHub account is @kartiktayal 
- your name in commit is `tayalkartik`
- I can't check yet the email

DamnClin:
@kartiktayal when you click on the link to sign you end up on an empty screen, just wait and the real screen will appear :D (that was the issue I got)

kartiktayal:
> You need to check if your 2 commits used the same email in your GitHub account
> 
> What I can see (from my phone):
> 
> * your GitHub account is @kartiktayal
> * your name in commit is `tayalkartik`
> * I can't check yet the email

@pascalgrimaud Thanks for pointing this out, I have committed the changes using the correct github account again. Do I need to remove the previous commits from the history like a hard reset now, so that there is only one account or it should work after this.
I will be cautious of this moving forward.

PR #4268, opened 2022-11-06
Refactor AsyncSpringLiquibaseTest

### LINKED ISSUE #2627: Refactor AsyncSpringLiquibaseTest
The current implementation of `AsyncSpringLiquibaseTest` is mainly testing implementation details while using a class that should be deleted (LogbackRecorder). A refactoring might be a good idea here

murdos:
SlownessThreshold of AsyncSpringLiquibase is now configurable, this allows speeding up tests.

Fixes #2627


DamnClin:
It is better! Can we delete `LogbackRecorder`?

src/main/resources/generator/server/springboot/dbmigration/liquibase/test/AsyncSpringLiquibaseTest.java.mustache
```diff
@@ -292,19 +169,13 @@ class AsyncSpringLiquibaseTest {
...
-          Thread.sleep(sleep);
-        } catch (InterruptedException x) {
-          // This should never happen
-          throw new Error(x);
-        }
+      try {
+        Thread.sleep(fakeDuration.toMillis());
+      } catch (InterruptedException x) {
+        // This should never happen
+        throw new Error(x);
```
DamnClin, line 176: You can throw an AssertionError

murdos:
> It is better! Can we delete `LogbackRecorder`?

Done! I wasn't aware to was only used by this test :smile:

DamnClin (approved):

DamnClin:
> > It is better! Can we delete `LogbackRecorder`?
> 
> Done! I wasn't aware to was only used by this test smile

Was my first concern with this code :)

PR #4269, opened 2022-11-07
Add Hibernate 2nd level cache

pblanchardie:
fixes #521

PR #4163, opened 2022-10-28
remove JPA properties that do not alter defaults

biergit:

PR #4270, opened 2022-11-07
Caffeine JCache

pblanchardie:
- [x] switch to [Caffeine JCache](https://github.com/ben-manes/caffeine/wiki/JCache)
- [x] provides feature `jcache` instead of `cache` (also updated ehcache), so that JCache-based modules can depend on it
- [ ] be consistent with ehcache: do not depent on "simple" cache and provide a specific configuration class (or depend on a new jcache module as it was done in the past but too heavy)

This PR makes the caffeine module compatible with PR #4269.

If we don't want that, we could also say that Caffeine is not an option for Hibernate 2nd-level caching.

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/JHLiteFeatureSlug.java
+++ b/src/main/java/tech/jhipster/lite/generator/JHLiteFeatureSlug.java
@@ -6,7 +6,7 @@ public enum JHLiteFeatureSlug implements JHipsterFeatureSlugFactory {
   ANGULAR_AUTHENTICATION("angular-authentication"),
   AUTHENTICATION("authentication"),
   BANNER("banner"),
-  CACHE("cache"),
+  JCACHE("jcache"),
   CLIENT_CORE("client-core"),
   CUCUMBER_AUTHENTICATION("cucumber-authentication"),
   DATABASE_MIGRATION("database-migration"),
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/caffeine/domain/CaffeineCacheModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/cache/caffeine/domain/CaffeineCacheModuleFactory.java
@@ -18,7 +18,8 @@ public class CaffeineCacheModuleFactory {
     return moduleBuilder(properties)
       .documentation(documentationTitle("Caffeine"), SOURCE.template("caffeine.md"))
       .javaDependencies()
-        .addDependency(groupId("com.github.ben-manes.caffeine"), artifactId("caffeine"))
+        .addDependency(groupId("com.github.ben-manes.caffeine"), artifactId("jcache"))
+        .addDependency(groupId("javax.cache"), artifactId("cache-api"))
         .and()
       .build();
     //@formatter:on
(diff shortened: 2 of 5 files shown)


pblanchardie:
fullapp fails because of a conflict: caffeine and ehcache shouldn't be enabled together
may I update the fullapp test to choose ehcache instead of caffeine?

pascalgrimaud:
@pblanchardie : yes just choose ehcache and use caffeine in other build

DamnClin:
I disagree, we should be able to use caffeine and ehcache together, this is a perfectly logic and valid setup!

pascalgrimaud:
In this case, we need to improve the support to not impact each option:
- the one who maintains ehcache, shouldn't care about the one who maintains caffeine

Otherwise, it should be a merged option: ehcache + caffeine

DamnClin:
The current integration have no impact between the 2 caches :⁠-⁠)

pblanchardie:
I agree it can be a valid setup to have both (weird scenario, isn't it?), but the current cache approach does not permit it.
If you prefer that we allow multiple cache implementations rather than restricting to one, the user must selects the one used by Hibernate as the second level cache. A property will also be needed to tell Spring which one to use.

> The current integration have no impact between the 2 caches :⁠-⁠)

If we switch Caffeine to JCache, it breaks.

If we keep the current implementation, it silently fails: there is a conflict on the configuration (which one overwrites the other? looks random) so it looks like it works, but in fact they are exclusive. And caffeine cannot be used as Hibernate 2nd level cache.

So anyway, we have to either make them exclusive or allow multiple options by modifying destinations, amongst other non trivial modifications.

DamnClin:
For me a valid scenario is to use ehcache as shared cache and caffeine as local cache. Also, I personally don't want to use hibernate level 2 cache so I don't want it to be enabled just because I add a cache support.

I'dd prefer to allow to have both caches but, if this turns out to be painfull to do, we must put all caches in a `cache` feature so they are exclusive in the landscape.

What I don't get is that you say caffeine is not supported as hibernate second level cache so I don't really get the point here... Hibernate second level cache activation should be a dedicated module (I personally really don't want that, too much side effects and, when working on a business centric application I don't need it)

pblanchardie:
> For me a valid scenario is to use ehcache as shared cache and caffeine as local cache. 

Got it. So you use both with Spring and need a MultipleCacheManagerConfig (see [baeldung sample](https://www.baeldung.com/spring-multiple-cache-managers))? User would choose which cache manager is the default one (@Primary) during generation? A bit heavy.

> Also, I personally don't want to use hibernate level 2 cache so I don't want it to be enabled just because I add a cache support.

As long as hibernate-jcache is not in the classpath, it won't be enabled.
To be explicit, we could add the hibernate.cache.use_second_level_cache property and set it to false (revert #4163).
Then #4269 would set it to true, as I suggested in reply to Pascal.

> 
> I'dd prefer to allow to have both caches but, if this turns out to be painfull to do, we must put all caches in a `cache` feature so they are exclusive in the landscape.
> 

I was making them exclusive with the `jcache` (and cache) feature because at first sight it will be painful and there is a risk on UX, with too many options if we add other implementations (infinispan, etc.).
IMHO, there are tons of edge cases we may want to support, but I'm not sure this one will be useful - if one wants to configure multiple cache managers, I guess he will manage that without the help of JHipster.

> What I don't get is that you say caffeine is not supported as hibernate second level cache so I don't really get the point here...

Hibernate supports JCache, so this PR is about using the JCache abstraction layer with Caffeine so Hibernate can use it as a second-level cache.
Other option is to have 2 modules: caffeine and caffeine-jcache. One providing jcache, the other not...
Last option is to keep Caffeine as-is, not an option for Hib2ndLevCach.

> Hibernate second level cache activation should be a dedicated module

Absolutely, that's the idea behind #4269 

> (I personally really don't want that, too much side effects and, when working on a business centric application I don't need it)

Recommending not to rely on Hibernate second-level caching in business centric (espacialy hexagonal) applications is another debate and I get the point, but still, it's a widespread, robust and very performant feature that we shouldn't "forget".

pascalgrimaud:
closing this as this PR was opened for too long
we can open a ticket and discuss if needed

PR #4371, opened 2022-11-11
Similar tests should be grouped in a single Parameterized test

qmonmert:


pascalgrimaud (approved):

DamnClin:
I really don't get it, we are loosing the functional requirements, for me this is a regression. Why this change?

qmonmert:
it was a suggestion of my sonar plugin (intelliJ), but you can revert it if you don't want this

PR #4481, opened 2022-11-22
group oauth2 modules under the same feature

Yourgene:
Fixes #4313

I didn't find any relevant test cases to edit/create.

If there's some, please let me know and I will update the PR.

PR #4507, opened 2022-11-24
group springdoc auth modules under same features

### LINKED ISSUE #4313: Add feature for spring boot oauth2 providers
The spring boot oauth2 modules for providers should be in the same feature:

* `spring-boot-oauth2-auth0`
* `spring-boot-oauth2-okta`

Yourgene:
Fixes #4313

In the continuation of PR #4481 (merged too quickly, RIP), this one creates the feature `oauth-springdoc`.

I also added the feature `authentication-springdoc`, I find it more logic and homogeneous this way.

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/JHLiteFeatureSlug.java
+++ b/src/main/java/tech/jhipster/lite/generator/JHLiteFeatureSlug.java
@@ -5,6 +5,7 @@ import tech.jhipster.lite.module.domain.resource.JHipsterFeatureSlugFactory;
 public enum JHLiteFeatureSlug implements JHipsterFeatureSlugFactory {
   ANGULAR_AUTHENTICATION("angular-authentication"),
   AUTHENTICATION("authentication"),
+  AUTHENTICATION_SPRINGDOC("authentication-springdoc"),
   BANNER("banner"),
   CACHE("cache"),
   CLIENT_CORE("client-core"),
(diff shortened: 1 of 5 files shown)


src/main/java/tech/jhipster/lite/generator/JHLiteFeatureSlug.java
```diff
@@ -16,6 +17,7 @@ public enum JHLiteFeatureSlug implements JHipsterFeatureSlugFactory {
   JAVA_BUILD_TOOL("java-build-tool"),
   JPA_PERSISTENCE("jpa-persistence"),
   OAUTH_PROVIDER("oauth-provider"),
+  OAUTH_SPRINGDOC("oauth-springdoc"),
```
DamnClin, line 20: Looks like this is only for external providers oauth, if so name should be explicit on that

Yourgene: you're right, it would be more clear. 

is `oauth-provider-springdoc` OK for you ? 

for me, `external` keyword isn't needed, and will only  break lisibility since the feature it depends, `oauth-provider`, doesn't have it

Yourgene: I repushed with changes commented above

PR #4558, opened 2022-12-02
Add prop type utility for vue

Gnuk:


DamnClin:
This probably needs a documentation since usage and utility isn't really obvious :)

Gnuk:
I've added a documentation entry under `documentation/vue.md` inside the generated code.

PR #4559, opened 2022-12-02
Store project folder

Gnuk:


src/main/webapp/app/module/secondary/RestStoreProjectFolderRepository.ts
```diff
@@ -0,0 +1,22 @@
...
+export class RestStoreProjectFolderRepository implements ProjectFoldersRepository {
+  private rest: RestProjectFolders;
+  private storage: StorageProjectFolder;
+
+  constructor(storage: Storage, axiosHttp: AxiosHttp) {
+    this.storage = new StorageProjectFolder(storage);
+    this.rest = new RestProjectFolders(axiosHttp);
+  }
+
+  get(): Promise<ProjectFolder> {
```
DamnClin, line 16: I think this should not be done alone since we need to sync properties on that

Gnuk: I will make a save to store this, that's why I put this like a draft. So it's ok for me :)

anderskristian:
I would like to know the reasoning for this?
I assume that projects knows where they are located?
'pwd' will give the result --> I don't think that should be stored in the project.
Would that mean that I cannot move a project on my local machine without need to update something in the project?

DamnClin:
> I would like to know the reasoning for this? I assume that projects knows where they are located? 'pwd' will give the result --> I don't think that should be stored in the project. Would that mean that I cannot move a project on my local machine without need to update something in the project?

They idea is to get back on your last project path when you open JHLite :)

pascalgrimaud:
opened for too long, let's close this and discuss if needed

PR #4562, opened 2022-12-02
add landscape scroll on click

Franceq34:
As @Gnuk suggested, I removed the domain part of my previous PR as it's just related to the primary part and there's no use to make the domain aware of this.


Franceq34:
Thanks @DamnClin :D

Issue #4066, opened 2022-10-21
JHLite: Spring Boot 3 - remove Zalando and use custom ControllerAdvice


pascalgrimaud:
Links:
- https://github.com/jhipster/jhipster-lite/issues/3601
- https://github.com/jhipster/jhipster-lite/pull/3832

You can use this branch: https://github.com/jhipster/jhipster-lite/tree/spring-boot-3 and submit a pull request against this branch

**Zalando should be removed and ExceptionTranslator should be reworked**

Then, it should fix the following tests:
- ForcedProjectFolderPrefixIntTest
- ExceptionTranslatorIT
- CucumberTest

![image](https://user-images.githubusercontent.com/9156882/197210925-67170560-1e5a-4ce1-bdbd-57377ceffded.png)

DamnClin:
I think we should updated the error management to handle internationalized error on the server side (so it can be tested) as described in https://github.com/jhipster/jhipster-lite/issues/2794

biergit:
Which module should provide the `ExceptionTranslator` ? Should this not be in a dedicated `web-exception-handling` module?

pascalgrimaud:
@biergit : yes, it should be done in a dedicated module, replacing the current one, Zalando problem

Issue #2794, opened 2022-07-26
Internationalized error modules


DamnClin:
The idea would be to add the possibility to easily add internationalized errors messages on the server side

pascalgrimaud:
increasing this, as the amount of work, related to several PRs linked to this ticket

PR #4565, opened 2022-12-03
International assertion errors

DamnClin:
Related to https://github.com/jhipster/jhipster-lite/issues/4066 and https://github.com/jhipster/jhipster-lite/issues/2794

Starting an internationalization of errors for assertion errors


DamnClin:
Next steps: 

* Update `java-base` to update the exceptions there (with attention payed to the coverage drop from new methods);
* Create a `assertion-error-handler` module to add the handler with messages etc;
* Update `GeneratorException` to be able to have an handler similar to this one (with additional complexity to manage keys);
* Update `java-base` module to have this kind of exceptions;
* Create a module with handler for this exceptions

PR #4746, opened 2022-12-14
[draft] Add possibility to parse a file and generate code

antarus:
Added a first version for a domain language parser.
This dsl is likely to change following a return.

This parser not parse a jdl but a new format (see documentation for format)

Addition of a first java code generator (class, record and enum)

TODO:

- [X] Add parser for primary and secondary
- [ ] Manage Collection - done in lexer
- [ ] Add generator for primary and secondary
- [ ] management of a syntax allowing to quickly create a Value Object 
proposal
class myclass {
   beerId VO UUID
}
would allow to quickly create a record named beerId and taking as argument a UUID - done in lexer

- [ ] Add annotation for Create Read Update Delete for select service to generate - done in lexer
- [ ] Add possibility to extend a class  - done in lexer
- [ ] added a @builder annotation to generate a builder associated with the class 
- [ ] generation of a jpa entity and its relations
- [ ] added a syntax to `copy` the attributes of a class/entity to facilitate the definition of a CRUD - done in lexer


DamnClin:
Thanks for the work! I think this should be in a dedicated project, here are some reasons: 

* This is adding code generation and a new step to run, that mean compilation will break at every branch switch (which is really annoying);
* This is designed to build application around data: this is a totally different usage from the one we had in mind when building JHlite. I think it may be a good idea to have it in a dedicated project explaining it's goals;
* By giving that a quick look I think you are only using `JHipsterProjectFolder`, `GitTestUtil`, `ProjectPath` and `Assert` from JHLite which can easily be replaced / imported;
* This is using JHlite classes in a way that is breaking the current architecture (the architecture test will fail);
* This generator should follow it's own lifecycle.

@pascalgrimaud what do you think? You think we can create a new project in JHipster group?

DamnClin:
I think we should have a direct discussion on that, I'm afraid that we won't go anywhere with PRs / messages. I think @murdos can make it happen :)

antarus:
Hi,
- the generation of code linked to antlr can be removed, we do not often modify the parser which generates this code.
- JHLite allows you to build an application as needed, unlike the jhipster generator which generates a lot at once. In any case, I contribute to this project for that. In my everyday life, I mainly do data but it does not seem incompatible with a hexagonal infrastructure that the generator-jhipster does not manage.
I'm still looking for a good example of CRUD with a hex framework to know what to generate :D

- normally, apart from Assert, I had to copy all the classes in the `dsl` package to avoid complications :D

For the separate project, I don't know at all :D

Regarding generation, Julien often talks about CRUD with jhlite in his speeches, it's still code generation.

DamnClin:
> I'm still looking for a good example of CRUD with a hex framework to know what to generate :D

Don't do hexagonal architecture if you want to build a CRUD application, that's not efficient. Hexagonal Architecture is designed to protect a rich domain model, if you don't need that, this architecture will slow you down.

Those kind of rich domain model are built around actions, not around data. You can't build a domain model from data, that's not the point.

jdubois:
I just added a big bug bounty on this (ping @antarus ) as this is a must-have feature, and it's very complex

pascalgrimaud:
@antarus : after having a quick look on this, I don't think we should merge this, into JHLite.

There are several reasons:

The 1st one is from @DamnClin :

> Don't do hexagonal architecture if you want to build a CRUD application, that's not efficient. Hexagonal Architecture is designed to protect a rich domain model, if you don't need that, this architecture will slow you down.

If we merge this, we'll suggest to our users to build their application around entities, around the database, and I don't want this, because it's not the approach suggested by JHLite. If it's the case, why don't use generator-jhipster instead ?

If you want playing with JDL, building your app around entities, just use generator-jhipster
If you want to build your app around **real** business, then you can use JHLite, and **we don't need JDL**

The 2nd one is the amount of code. I don't know who will maintain this, excepting you. With this architecture, the contributor who provided some features will be responsible to maintain it. And I don't know if you have the bandwidth for that?

So my idea would be:
- create a new project in JHipster Organization
- generate a new project with JHLite dependency
- so, you can report all the code there
- then, we can see if this new project is regularly used, how it is used, etc
- so we'll have more feedbacks

@antarus : what do you think? If you agree with this solution, I can create this new project and you'll be the lead and main maintainer.

jdubois:
I totally agree @pascalgrimaud : I believe this code could have much wider impact than just JHipster Lite.
It could work with also JHipster "classic", but it’s probably not far from working with any Spring Boot project…. I’m thinking start.spring.io here! Imagine if this could work, this would have incredible success.
@antarus how about making this a separate project under the JHipster organization? Or make it part of JHipster Online?

antarus:
hello, 

@murdos  contacted me but he doesn't seem to have seen my answer :D

Personally, I don't use the generator as an application generator but more as an application part generator. The current generator forces me to generate a complete app just for the part that interests me at the moment T (generally resource/screen because I'm not frontend at all :D).

This is partly why, in my dsl proposal, I separate the code from the domain and the infrastructure.

If you think it's better to be a separate project why not.
But we will have to determine what to do or not with this new project without encroaching on jhipster and JHLite generator and in function review the DSL.

Regarding the bandwidth, it is quite variable depending on the job.

pascalgrimaud:
@antarus :
I init this project https://github.com/jhipster/jhipster-lite-jdl and invite you as maintainer
Let me know if you need more config of the repo but you should be able to push and add the CI
Maybe sonarcloud to have the same level at 100% than JHipster Lite

pascalgrimaud:
Is it ok @antarus ?
Can this PR be closed ?

antarus:
@pascalgrimaud 
yes you can delete it. I still have the code on my repo.
I'm waiting to have time and know what and how to do exactly ;)

pascalgrimaud:
only closing this PR, the code is still in your branch of your fork :)
as soon as the project is correctly init in the repo, don't forget to claim the bounty @antarus

Issue #4637, opened 2022-12-07
Upgrade Spring Cloud to 2022.0.0


DamnClin:
Once Spring Cloud 2022.0.0 reach a final version (https://github.com/spring-cloud/spring-cloud-release/milestone/115) we have to revert https://github.com/jhipster/jhipster-lite/pull/4636 and upgrade

DamnClin:
The release is available

PR #4840, opened 2022-12-22
Upgrade spring cloud

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/4637


DamnClin:
I'm really surprised that it's working without doing nothing but...

PR #4907, opened 2023-01-02
Introduce annotation @ExcludeFromGeneratedCodeCoverage

murdos:
This annotation will also be detected by Jacoco since it contains "Generated". 
Uses this new annotation in replacement of `@Generated` in places where it was explicitly added to skip code coverage check.

Having code annotated with `@Generated(reason = "Hard to test implement detail error management")` is a bit weird and misleading, since the code has not been generated, but written manually. And the reason is not here to explain why the code has (not!) been generated, but why we want to exclude it from code coverage report.

### DIFF
--- a/src/main/java/tech/jhipster/lite/ApplicationStartupTraces.java
+++ b/src/main/java/tech/jhipster/lite/ApplicationStartupTraces.java
@@ -9,7 +9,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.core.env.Environment;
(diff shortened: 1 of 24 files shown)


pascalgrimaud:
Maybe a shorter name like 'ExcludeFromCoverage' ?
Let's propose others before changing :-)

DamnClin:
> Unless it must absolutely contain Generated

It must (unless some change were made on Jacoco on that)

pascalgrimaud:
@murdos : what do you think about doing the same for the generated part?

murdos:
> @murdos : what do you think about doing the same for the generated part?

I was planning to do that too, but was waiting for the feedback on the generator :-) 
I'll do that soon.

PR #4954, opened 2023-01-08
Gradle: use TOML version catalog

murdos:
Update https://github.com/jhipster/jhipster-lite/issues/1862


atomfrede (approved):
Nice work!

pascalgrimaud:
Something important to check is dependabot + toml, so we are sure the library will be upgraded automatically

murdos:
> Something important to check is dependabot + toml, so we are sure the library will be upgraded automatically

My plan is to still use dependencies in our pom.xml for all java depencies - including gradle, and have a toml file for dependabot that will only contain gradle plugins (the only part that will be gradle specific).

pascalgrimaud:
> My plan is to still use dependencies in our pom.xml for all java depencies - including gradle

Yes, it's smart! I think it's what we should do, so we don't need to maintain 2 distinct files for dependencies: 1 for maven and 1 for gradle :-)

PR #4556, opened 2022-12-01
WIP: feature: adding cassandra support

Yourgene:
Fixes #877
This is still work in progress.

TODOS:
- [x] add cassandra module 
- [x] add testcontainers support
- [x] add a database migration support
- [x] add a dummy-cassandra module, as for dummy mongodb. 
- [ ] add a datastax connection support ? as for oauth2 okta and auth0 providers. What do you think about that ?

Issue #2611, opened 2022-07-15
Manage spring.factories in modules


DamnClin:
`META-INF/spring.factories` are files loaded very early in spring lifecycle that comes to be very usefull. As those files can't be renamed / duplicated we need a way to manage them in modules.

The idea is to be able to set a value to a property and either:

* Create the property with the value;
* Add value to existing property.

For now we only need that for test spring.factories.

When done this must be used in the MongoDB module

Yourgene:
Hi,
If nobody is already working on this issue, I'll take it. 

2 of 3 cassandra modules I'm working on are using spring.factories

DamnClin:
You can have a look at https://github.com/jhipster/jhipster-lite/tree/main/src/main/java/tech/jhipster/lite/module/domain/javaproperties since it may use some of this logic

pascalgrimaud:
Adding a bounty as it was more work than expected

pascalgrimaud:
Thanks for your work @Yourgene : don't forget to claim your bounty

Yourgene:
Thank you for the merge, and the added bounty :)

PR #4958, opened 2023-01-08
Feature/issue 877 cassandra support p1

### LINKED ISSUE #877: Spring Boot Database: Cassandra
Like generator-jhipster is currently doing, the goal of this issue would be to help users to add configuration for cassandra into their application.

Yourgene:
Fixes #877 

As disscussed in #4556, this first PR adds cassandra module


src/main/resources/generator/server/springboot/database/cassandra/spring.factories.mustache
```diff
@@ -0,0 +1 @@
+org.springframework.context.ApplicationListener={{packageName}}.TestCassandraManager
```
Yourgene, line 1: this usage of spring.factories in tests to start testcontainers is the same as in mongodb and neo4j modules.

There's a drawback using it : if a user selects more than one of these 3 modules, this file will be overriden by the last applied module. Is it acceptable ?

Maybe these 3 modules should be grouped in a nosql feature ?

src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java
```diff
@@ -0,0 +1,86 @@
...
+          toSrcTestJava().append(packagePath).append(CASSANDRA_SECONDARY).append("CassandraJSR310DateConvertersTest.java")
+        )
+        .add(SOURCE.template("cassandra.yml"), toSrcMainDocker().append("cassandra.yml"))
+        .add(SOURCE.template("TestCassandraManager.java"), toSrcTestJava().append(packagePath).append("TestCassandraManager.java"))
+        .add(SOURCE.template("CassandraKeyspaceIT.java"), toSrcTestJava().append(packagePath).append("CassandraKeyspaceIT.java"))
+        .add(SOURCE.template("spring.factories"), to("src/test/resources/META-INF/spring.factories"))
+        .and()
+      .springMainProperties()
+        .set(propertyKey("spring.cassandra.contact-points"), propertyValue("127.0.0.1"))
+        .set(propertyKey("#spring.cassandra.keyspace-name"), propertyValue("yourKeyspace"))
```
Yourgene, line 59: this property is commented : we don't know the keyspace name used by the user. 

some info : 
- starting the app with this property commented works.
- starting the app with this property not commented, but empty fails. 
- in cassanda.md, an instruction about this comment is present

Instead of using a "trick" to have this property commented, is it worth to create an api for it in jhlite ?

.github/workflows/github-actions.yml
```diff
@@ -145,6 +145,7 @@ jobs:
           - consulapp
           - gatewayapp
           - mongodbapp
+          - cassandraapp
```
Yourgene, line 148: for the moment, this app fails even if the cassandra container is started.

my first guess is that the springboot app is launched too quickly, and cassandra is not ready yet. I will take time at the beginning of next week to investigate.

src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java
murdos: We should use a default keyspace, derived from project baseName as done e.g. for the postgres database
https://github.com/jhipster/jhipster-lite/blob/64b14b0c473e6f7f675a34aa89249638ef345901/src/main/java/tech/jhipster/lite/generator/server/springboot/database/postgresql/domain/PostgresqlModuleFactory.java#L49 or in generator-jhipster https://github.com/jhipster/generator-jhipster/blob/fb9e01c93afb54b862dd7d1ffea69d0bd26ba475/generators/server/templates/src/main/resources/config/application-dev.yml.ejs#L164

Yourgene: Hi,

For this to work we also need to add the programmatic keyspace creation in `CassandraDatabaseConfiguration`. The driver doesn't create automatically keyspaces if they don't exist.

I see 2 drawbacks by going this way :
- the keyspace creation is delegated to the java application, and not the dba team (or another entity). In the case of Cassandra, the keyspace creation is something important to consider, especially for the replication strategy. I guess that's one of the reasons that a `spring.cassandra.keyspace-action` is not available.
- This keyspace creation in `CassandraDatabaseConfiguration` will "clash" with the way `cassandra-migration` module works : the latter can be used to create keyspaces and tables through cql scripts (the same way as in jhipster). That's how the keyspace is created in `dummy-cassandra`

That's why I think it's better to not impose a keyspace in this module. However if you don't share that idea,  I'm totally fine to implement your remark :)

What do you think ?

.github/workflows/github-actions.yml
Yourgene: adding sleep solved the issue

src/main/resources/generator/server/springboot/database/cassandra/spring.factories.mustache
pascalgrimaud: For now, we can keep this behavior, as I don't think it's a frequent use case to select all these modules

Anyway, later, it can be improved by appending to the existing file, like we do for application.properties

Yourgene: Alright, let's keep it this way for the moment.

Here's some info about the solution you proposed : 
- Appeding to spring.factories won't work, you can't list more than 1 class in `org.springframework.context.ApplicationListener` :(
- Going the annotation way, with an `@EventListener` and register the bean in a `@TestConfiguration` class won't work either because bean instantiation happens too late, after `ApplicationEnvironmentPreparedEvent` that we listen.

I didn't find anything that works and that is as simple and "light" as the spring.factories method, but I didn't go too far in my investigations.

pascalgrimaud:
I'll review it soon, and will have a look to the existing comment too.
Thanks for being patient :-)

src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java
murdos: Oh, I see.
Then maybe starting with the property commented is the best option.

Yourgene: If @pascalgrimaud is also alright with it, we can keep that solution then :)

Yourgene: and what about the way to do it ? do you think having something like 

```java
.springMainProperties()
        .set(CommentedPropertyKey("x"), propertyValue("y))
```
is worth the effort to implement ?
I'm OK to do it just after if you think it has a value

Yourgene: > I see 2 drawbacks by going this way :

to complete my comment, here's a [note from spring documentation](https://docs.spring.io/spring-data/cassandra/docs/current/reference/html/#cassandra.schema-management) about creating a keyspace in `@Configuration` :

> Keyspace creation allows rapid bootstrapping without the need of external keyspace management. This can be useful for certain scenarios but should be used with care[...]

src/main/resources/generator/server/springboot/database/cassandra/spring.factories.mustache
Yourgene: will be solved by https://github.com/jhipster/jhipster-lite/issues/2611

src/test/java/tech/jhipster/lite/generator/server/springboot/database/domain/CassandraModuleFactoryTest.java
```diff
@@ -0,0 +1,111 @@
+package tech.jhipster.lite.generator.server.springboot.database.domain;
```
pascalgrimaud, line 1: not in correct package, should be in: `database/cassandra/domain`

src/main/resources/generator/server/springboot/database/cassandra/CassandraDatabaseConfiguration.java.mustache
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.data.cassandra.core.convert.CassandraCustomConversions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Configuration
+public class CassandraDatabaseConfiguration {
```
pascalgrimaud, line 13:
```suggestion
class CassandraDatabaseConfiguration {
```

src/main/resources/generator/server/springboot/database/cassandra/CassandraJSR310DateConverters.java.mustache
```diff
@@ -0,0 +1,52 @@
...
+      private ZonedDateTimeToTupleConverter() {}
+
+      @Override
+      public TupleValue convert(@Nonnull ZonedDateTime source) {
+        TupleValue tupleValue = type.newValue();
+        tupleValue.setInstant(0, source.toInstant());
+        tupleValue.setString(1, source.getZone().toString());
+        return tupleValue;
+      }
+    }
```
pascalgrimaud, line 51:
```suggestion
  @WritingConverter
  public static class ZonedDateTimeToTupleConverter implements Converter<ZonedDateTime, TupleValue> {

    private final TupleType type = DataTypes.tupleOf(DataTypes.TIMESTAMP, DataTypes.TEXT);

    public static final ZonedDateTimeToTupleConverter INSTANCE = new ZonedDateTimeToTupleConverter();

    private ZonedDateTimeToTupleConverter() {}

    @Override
    public TupleValue convert(@Nonnull ZonedDateTime source) {
      TupleValue tupleValue = type.newValue();
      tupleValue.setInstant(0, source.toInstant());
      tupleValue.setString(1, source.getZone().toString());
      return tupleValue;
    }
  }
```

src/main/resources/generator/server/springboot/database/cassandra/CassandraKeyspaceIT.java.mustache
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@IntegrationTest
+class CassandraKeyspaceIT {
+
+  private static final String KEYSPACE = "{{ baseName }}";
+  @Autowired
```
pascalgrimaud, line 14:
```suggestion
  private static final String KEYSPACE = "{{ baseName }}";

  @Autowired
```

src/main/resources/generator/server/springboot/database/cassandra/cassandra.md
```diff
@@ -0,0 +1,17 @@
...
+
+```bash
+docker compose -f src/main/docker/cassandra.yml up -d
+```
+
+You can interact with your database by running
+
+```bash
+docker exec -it cassandra cqlsh
+```
```
pascalgrimaud, line 17: ok with this 1st version, but the application won't work out of the box. We should add some instruction to create the keyspace, for example:

```
create keyspace if not exists THEKEYSPACE WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 } AND DURABLE_WRITES = true;
```

src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java
pascalgrimaud: I didn't see this comment before sorry
So I'm in favor of having something working out of the box, like it's done in generator-jhipster

But for this 1st version it's ok, it can be improved later.
The idea behind JHipster is:
- create a project
- start Cassandra docker-compose
- start the application
- it works without doing more

Currently, we need to manually create the keyspace and put the correct keyspace in application.properties
Instead, we should put one by default, as suggested by @murdos  and document it to explain how to edit it

pascalgrimaud:
Thanks a lot for your work here @Yourgene 
Few comments from my side but it's ok
If you don't have time to fix it today, I'll fix it to merge and prepare the next release with Cassandra :)

Yourgene:
> Thanks a lot for your work here @Yourgene Few comments from my side but it's ok If you don't have time to fix it today, I'll fix it to merge and prepare the next release with Cassandra :)

Thanks for your review and comments :) 

I'll have time this afternoon to make requested changes

src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java
Yourgene: I understand the wanted behaviour, and I'm fine with it :) 

I share your idea to let it like that for the first version, and improve it after.

I'll work on it when the 3 cassandra modules are merged

src/test/java/tech/jhipster/lite/generator/server/springboot/database/domain/CassandraModuleFactoryTest.java
Yourgene: thanks, sharp eye :)

updated

src/main/resources/generator/server/springboot/database/cassandra/cassandra.md
Yourgene: as you suggested, added the instructions

Yourgene:
@pascalgrimaud seems good to me. 

If you're satisfied with the current state, I'll squash the commits for the merge

pascalgrimaud:
Great work @Yourgene !!!

Yourgene:
Hi, I claimed the bug bounty for this PR : [link](https://opencollective.com/generator-jhipster/expenses/120420)

pascalgrimaud:
@Yourgene : approved

Yourgene:
@pascalgrimaud Thank you :)

PR #5352, opened 2023-02-13
jhlite custom: ignore all children of a parent module

### LINKED ISSUE #4338: JHLite Custom: ignore all children of a parent module
In JHLite Custom, when ignoring a module, it should ignore all the children too.

It doesn't work, and there is a stacktrace:

```
jhlite-hidden-resources.slugs=custom-jhlite,angular-core
```

You need to add other children too:

```
jhlite-hidden-resources.slugs=custom-jhlite,angular-core,angular-health,angular-oauth2,angular-jwt
```

The problem is:
- there are new modules, which have been coded in JHLite: angular-beer
- there is a new release
- the user upgrades the JHLite custom
- it's broken, unless he/she ignores the new module: angular-beer

renanfranca:
Fix #4338

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
@@ -1,0 +1,11 @@
+package tech.jhipster.lite.module.domain.resource;
+
+import java.util.Collection;
+import tech.jhipster.lite.common.domain.JHipsterCollections;
+
+public record JHipsterHiddenModules(Collection<String> slugs, Collection<JHipsterModuleTag> tags) {
+  public JHipsterHiddenModules(Collection<String> slugs, Collection<JHipsterModuleTag> tags) {
+    this.slugs = JHipsterCollections.immutable(slugs);
+    this.tags = JHipsterCollections.immutable(tags);
+  }
+}
--- a/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
@@ -3,10 +3,13 @@ package tech.jhipster.lite.module.domain.resource;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import tech.jhipster.lite.error.domain.Assert;
 import tech.jhipster.lite.module.domain.JHipsterModule;
 import tech.jhipster.lite.module.domain.JHipsterModuleSlug;
@@ -14,19 +17,116 @@ import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
 
 public class JHipsterModulesResources {
 
+  private static final Logger log = LoggerFactory.getLogger(JHipsterModulesResources.class);
+
   private final Map<JHipsterModuleSlug, JHipsterModuleResource> resources;
 
-  public JHipsterModulesResources(Collection<JHipsterModuleResource> modulesResources) {
+  public JHipsterModulesResources(Collection<JHipsterModuleResource> modulesResources, JHipsterHiddenModules hiddenModules) {
     Assert.field("modulesResources", modulesResources).notEmpty().noNullElement();
+    Assert.notNull("jhipsterHiddenModules", hiddenModules);
 
     assertUniqueSlugs(modulesResources);
 
     resources =
       Collections.unmodifiableMap(
-        modulesResources
-          .stream()
+        removeHiddenModules(modulesResources, hiddenModules)
           .collect(Collectors.toMap(JHipsterModuleResource::slug, Function.identity(), (x, y) -> y, LinkedHashMap::new))
       );
+  }
+
+  private Stream<JHipsterModuleResource> removeHiddenModules(
+    Collection<JHipsterModuleResource> modulesResources,
+    JHipsterHiddenModules hiddenModules
+  ) {
+    Collection<String> nestedDependenciesSlugs = findNestedDependencies(modulesResources, hiddenModules);
+    DisplayHiddenResources displayHiddenResources = findDisplayAndHiddenResources(modulesResources, hiddenModules, nestedDependenciesSlugs);
+    if (displayHiddenResources.hasHiddenResources()) {
+      String hiddenModulesSlugs = displayHiddenResources.hiddenSlugs();
+      log.info("The following modules are hidden: {}.", hiddenModulesSlugs);
+    }
+    return displayHiddenResources.displayedStream();
+  }
+
+  private DisplayHiddenResources findDisplayAndHiddenResources(
+    Collection<JHipsterModuleResource> modulesResources,
+    JHipsterHiddenModules hiddenModules,
+    Collection<String> nestedDependenciesSlugs
+  ) {
+    Map<Boolean, List<JHipsterModuleResource>> partitionedResources = modulesResources
+      .stream()
+      .collect(Collectors.partitioningBy(resource -> allowed(resource, hiddenModules, nestedDependenciesSlugs)));
+    return new DisplayHiddenResources(partitionedResources.get(true), partitionedResources.get(false));
+  }
+
+  private boolean allowed(
+    JHipsterModuleResource resource,
+    JHipsterHiddenModules hiddenModules,
+    Collection<String> nestedDependenciesSlugs
+  ) {
+    return notExcludedSlug(resource, hiddenModules, nestedDependenciesSlugs) && noExcludedTag(resource, hiddenModules);
+  }
+
+  private boolean notExcludedSlug(
+    JHipsterModuleResource resource,
+    JHipsterHiddenModules hiddenModules,
+    Collection<String> nestedDependenciesSlugs
+  ) {
+    return !hiddenModules.slugs().contains(resource.slug().get()) && !nestedDependenciesSlugs.contains(resource.slug().get());
+  }
+
+  private boolean noExcludedTag(JHipsterModuleResource resource, JHipsterHiddenModules hiddenModules) {
+    return hiddenModules.tags().stream().noneMatch(tag -> resource.tags().contains(tag));
+  }
+
+  private Collection<String> findNestedDependencies(
+    Collection<JHipsterModuleResource> modulesResources,
+    JHipsterHiddenModules hiddenModules
+  ) {
+    return findNestedDependenciesBySlugs(hiddenModules.slugs(), modulesResources);
+  }
+
+  private Collection<String> findNestedDependenciesBySlugs(Collection<String> slugs, Collection<JHipsterModuleResource> modulesResources) {
+    return slugs.stream().flatMap(slug -> allSlugsNestedDependenciesOf(slug, modulesResources)).toList();
+  }
+
+  private Stream<String> allSlugsNestedDependenciesOf(String slug, Collection<JHipsterModuleResource> modulesResources) {
+    return allResourcesNestedDependenciesOf(new JHipsterModuleSlug(slug), modulesResources)
+      .map(moduleResource -> moduleResource.slug().get());
+  }
+
+  private Stream<JHipsterModuleResource> allResourcesNestedDependenciesOf(
+    JHipsterModuleSlug slug,
+    Collection<JHipsterModuleResource> modulesResources
+  ) {
+    Collection<JHipsterModuleResource> childrensDependencies = this.getChildrensDependencies(slug, modulesResources);
+    if (noMoreNestedResource(childrensDependencies)) {
+      return Stream.of();
+    }
+    return Stream.concat(
+      childrensDependencies.stream(),
+      childrensDependencies.stream().map(moveToNextNestedResource(modulesResources)).flatMap(t -> t)
+    );
+  }
+
+  private boolean noMoreNestedResource(Collection<JHipsterModuleResource> childrensDependencies) {
+    return childrensDependencies.isEmpty();
+  }
+
+  private Function<JHipsterModuleResource, Stream<JHipsterModuleResource>> moveToNextNestedResource(
+    Collection<JHipsterModuleResource> modulesResources
+  ) {
+    return resource -> this.allResourcesNestedDependenciesOf(resource.slug(), modulesResources);
+  }
+
+  private Collection<JHipsterModuleResource> getChildrensDependencies(
+    JHipsterModuleSlug slug,
+    Collection<JHipsterModuleResource> modulesResources
+  ) {
+    return modulesResources.stream().filter(moduleResource -> isChildrensOf(slug, moduleResource)).toList();
+  }
+
+  private boolean isChildrensOf(JHipsterModuleSlug slug, JHipsterModuleResource moduleResource) {
+    return moduleResource.organization().dependencies().stream().anyMatch(dependency -> dependency.slug().equals(slug));
   }
 
   private void assertUniqueSlugs(Collection<JHipsterModuleResource> modulesResources) {
@@ -61,6 +161,20 @@ public class JHipsterModulesResources {
 
     if (!resources.containsKey(slug)) {
       throw new UnknownSlugException(slug);
+    }
+  }
+
+  private record DisplayHiddenResources(Collection<JHipsterModuleResource> displayed, Collection<JHipsterModuleResource> hidden) {
+    public Stream<JHipsterModuleResource> displayedStream() {
+      return displayed().stream();
+    }
+
+    private boolean hasHiddenResources() {
+      return !hidden().isEmpty();
+    }
+
+    private String hiddenSlugs() {
+      return hidden().stream().map(hidden -> hidden.slug().get()).collect(Collectors.joining(", "));
     }
   }
 }
--- a/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/JHipsterHiddenResourcesProperties.java
+++ b/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/JHipsterHiddenResourcesProperties.java
@@ -2,7 +2,6 @@ package tech.jhipster.lite.module.infrastructure.secondary;
 
 import java.util.Collection;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import tech.jhipster.lite.module.domain.resource.JHipsterModuleResource;
 import tech.jhipster.lite.module.domain.resource.JHipsterModuleTag;
 
 @ConfigurationProperties("jhlite-hidden-resources")
@@ -23,23 +22,11 @@ class JHipsterHiddenResourcesProperties {
     }
   }
 
-  public boolean allowed(JHipsterModuleResource resource) {
-    return notExcludedSlug(resource) && noExcludedTag(resource);
+  public Collection<String> getSlugs() {
+    return slugs;
   }
 
-  private boolean notExcludedSlug(JHipsterModuleResource resource) {
-    if (slugs == null) {
-      return true;
-    }
-
-    return !slugs.contains(resource.slug().get());
-  }
-
-  private boolean noExcludedTag(JHipsterModuleResource resource) {
-    if (tags == null) {
-      return true;
-    }
-
-    return tags.stream().noneMatch(tag -> resource.tags().contains(tag));
+  public Collection<JHipsterModuleTag> getTags() {
+    return tags;
   }
 }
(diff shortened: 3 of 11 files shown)


renanfranca:
I implemented the first request:
> keep current property, and improve the behavior to exclude this module and all dependant module

The second request, I add a comment at the properties `jhlite-hidden-resources.slugs` and increment the documentation:
> maybe, display the excluded modules at start, to avoid "silent" exclusions

@pascalgrimaud, could you tell me where do you want to display the hidden modules?

pascalgrimaud:
@renanfranca : maybe in the logs at startup, would be enough

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/JHipsterHiddenResourcesProperties.java
```diff
@@ -23,6 +29,13 @@ public void setTags(Collection<String> tags) {
     }
   }
 
+  public void addNestedDependencies(Collection<JHipsterModuleResource> modulesResources) {
+    if (slugs == null) {
+      return;
+    }
+    addNestedDependenciesBySlugs(modulesResources);
```
DamnClin, line 36: This can't be done here, the hidden modules logic must be moved in domain to make this calculation

DamnClin (changes requested):

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/JHipsterHiddenResourcesProperties.java
renanfranca: Thank you @DamnClin for pointing this out 😃 ! I pushed a new version to be reviewed 👍

renanfranca:
> @renanfranca : maybe in the logs at startup, would be enough

Thank you @pascalgrimaud , I ll try to do it 👍

renanfranca:
Hello @DamnClin,

This jhispter lite custom module impressed me!!! It actually runs the original jhipister lite project (backend and frontend) only by importing it as a dependency with maven!

That makes me so curious that I've been studying it, I learn how to do it using a classifier tag  (here is the article: https://medium.com/publicis-sapient-france/import-spring-boot-application-as-a-maven-dependency-a25e03e5c3a ). But I didn't know how the hipster lite did it, there wasn't any classifier tag...

What I want to do is to generate an application using jhlite module and it could be able to run with my changes which weren't merged yet.

if you could point me in the right direction, I appreciated it 😄

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenResources.java
```diff
@@ -0,0 +1,103 @@
+package tech.jhipster.lite.module.domain.resource;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+import tech.jhipster.lite.module.domain.JHipsterModuleSlug;
+
+public class JHipsterHiddenResources {
```
DamnClin, line 10: Without looking too much at it I think it would be easier to update `JHipsterModulesResources` to take a `JHipsterHiddenModules` in constructor with `JHipsterHiddenModules` containing slugs and tags. This way we can build modules with hidden ones easily

DamnClin:
> What I want to do is to generate an application using jhlite module and it could be able to run with my changes which weren't merged yet.

* mvn install on your branch on JHipsterLite
* Generate an app with custom JHlite
* Update JHlite version in your generated app to use the snapshot one you just built

Haven't tried it now but I think I remember doing it when working on that feature

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenResources.java
renanfranca: Thank you for the fast response @DamnClin,
I am not so sure if I understand what you want me to do this time. So I implemented what I think I understand and I didn't rebase this time, I pushed a new commit to be easier to talk about the change and to undo it if I did all wrong 😅 .

DamnClin: The idea is to update [JHipsterModulesResources](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java): the class responsible of resources management to build it with the hidden resources and let it do the filtering logic (getting nested modules etc).

Your current `JHipsterHiddenResources` needs a collection of `JHipsterModuleResource` which is, in fact, a `JHipsterModulesResources` so I think it will be easier to do this filtering in `JHipsterModuleResource` or in [JHipsterLandscape](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/module/domain/landscape/JHipsterLandscape.java) which is the next step, the one knowing the dependencies

DamnClin: Aftertought: can't be in Landscape since we also want to hide them in patch, must be in `JHipsterHiddenResources`

DamnClin: No new version here, need a force push?

renanfranca:
> 
> * mvn install on your branch on JHipsterLite
> * Generate an app with custom JHlite
> * Update JHlite version in your generated app to use the snapshot one you just built
> 
> Haven't tried it now but I think I remember doing it when working on that feature

I tried the `./mvnw install` but it didn't install the `jhlite-0.27.1-SNAPSHOT-test.jar` and when I run the generated app with the snapshot I got the following message `Failed to execute goal on project customjhlite: Could not resolve dependencies for project tech.jhipster.customjhlite:customjhlite:jar:0.0.1-SNAPSHOT: Could not find artifact tech.jhipster.lite:jhlite:jar:tests:0.27.1-SNAPSHOT`.

When I looked at the maven local repository and compare the generated files of the jhlite-0.27.0 version, it was generated the jhlite-0.27.0-test.jat (image below)

image

I studied that the `./mvnw install` should be enough to install the `jhlite-0.27.1-SNAPSHOT-test.jar`, Do you remember if have to do additional command?

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenResources.java
renanfranca: I am sorry @DamnClin , I forced push now 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+import tech.jhipster.lite.module.domain.JHipsterModuleSlug;
+
+public class JHipsterHiddenModules {
+
+  private final Collection<String> slugs;
+  private final Collection<JHipsterModuleTag> tags;
+  private final Collection<String> nestedDependenciesSlugs = new HashSet<>();
```
DamnClin, line 14: This should not be here, this mutable set is really hard to understand. Calculating nested hidden dependencies is JHipsterModulesResources responsibility

DamnClin:
Took a quick look to the test classifier and it was only built when deploying, dunno why it was done that way, did https://github.com/jhipster/jhipster-lite/pull/5381 to fix that, waiting for @pascalgrimaud approval

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
renanfranca: @DamnClin, thanks! I'll do it soon!

renanfranca:
> Took a quick look to the test classifier and it was only built when deploying, dunno why it was done that way, did #5381 to fix that, waiting for @pascalgrimaud approval

Nice! good to know 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
renanfranca: Hello @DamnClin , I pushed a new version! Please, let me know what you think about it 😃 .

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
```diff
@@ -0,0 +1,22 @@
+package tech.jhipster.lite.module.domain.resource;
+
+import java.util.Collection;
+
+public class JHipsterHiddenModules {
```
DamnClin, line 5: Can be a record

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private final Collection<String> slugs;
+  private final Collection<JHipsterModuleTag> tags;
+
+  public JHipsterHiddenModules(Collection<String> slugs, Collection<JHipsterModuleTag> tags) {
+    this.slugs = slugs;
```
DamnClin, line 11: You can use JHipsterCollections.immutable(...) to ensure that they are immutable and not null

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +18,116 @@ public class JHipsterModulesResources {
...
+  }
+
+  private List<String> allSlugsNestedDependenciesOf(Collection<JHipsterModuleResource> modulesResources, String slug) {
+    return allResourcesNestedDependenciesOf(new JHipsterModuleSlug(slug), modulesResources)
+      .stream()
+      .map(moduleResource -> moduleResource.slug().get())
+      .toList();
+  }
+
+  private Collection<JHipsterModuleResource> allResourcesNestedDependenciesOf(
```
DamnClin, line 94: Can result in a stream instead of a collection

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +18,116 @@ public class JHipsterModulesResources {
...
+
+  private Collection<String> findNestedDependenciesBySlugs(Collection<String> slugs, Collection<JHipsterModuleResource> modulesResources) {
+    Collection<String> nestedDependencies = new HashSet<>();
+    for (String slug : slugs) {
+      nestedDependencies.addAll(allSlugsNestedDependenciesOf(modulesResources, slug));
+    }
+    return nestedDependencies;
+  }
+
+  private List<String> allSlugsNestedDependenciesOf(Collection<JHipsterModuleResource> modulesResources, String slug) {
```
DamnClin, line 87: Can result in a stream instead of a collection

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +18,116 @@ public class JHipsterModulesResources {
...
+    JHipsterHiddenModules hiddenModules
+  ) {
+    if (hiddenModules.getSlugs() == null) {
+      return List.of();
+    }
+    return findNestedDependenciesBySlugs(hiddenModules.getSlugs(), modulesResources);
+  }
+
+  private Collection<String> findNestedDependenciesBySlugs(Collection<String> slugs, Collection<JHipsterModuleResource> modulesResources) {
+    Collection<String> nestedDependencies = new HashSet<>();
```
DamnClin, line 80: With stream result you can flatMap all streams and toList() them

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterHiddenModules.java
renanfranca: Nice! Done 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Cool! did 👍

pascalgrimaud:
@renanfranca : to ease the test, you can directly add the ignore in jhlite to test, instead of a jhlite-custom

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Thank you @DamnClin! Java Stream is easy to use but harder to master! That was so awesome, the code is much better now! Thanks for leading the steps, flatMap is so powerful!  I pushed the changes in a new commit! Let me know if its good, so I'll go to the log the hidden module feature 👍

renanfranca:
> @renanfranca : to ease the test, you can directly add the ignore in jhlite to test, instead of a jhlite-custom

Thank you @pascalgrimaud! 😃 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +16,98 @@ public class JHipsterModulesResources {
...
+      childrensDependencies.stream(),
+      childrensDependencies.stream().map(moveToNextNestedResource(modulesResources)).flatMap(t -> t)
+    );
+  }
+
+  private boolean noMoreNestedResource(final Collection<JHipsterModuleResource> childrensDependencies) {
+    return childrensDependencies.isEmpty();
+  }
+
+  private Function<? super JHipsterModuleResource, ? extends Stream<JHipsterModuleResource>> moveToNextNestedResource(
```
DamnClin, line 94: You can remove the ? extends and ? super

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +16,98 @@ public class JHipsterModulesResources {
...
+  private Stream<String> allSlugsNestedDependenciesOf(String slug, Collection<JHipsterModuleResource> modulesResources) {
+    return allResourcesNestedDependenciesOf(new JHipsterModuleSlug(slug), modulesResources)
+      .map(moduleResource -> moduleResource.slug().get());
+  }
+
+  private Stream<JHipsterModuleResource> allResourcesNestedDependenciesOf(
+    JHipsterModuleSlug slug,
+    Collection<JHipsterModuleResource> modulesResources
+  ) {
+    final Collection<JHipsterModuleResource> childrensDependencies = this.getChildrensDependencies(slug, modulesResources);
```
DamnClin, line 80: You can remove the final, it is no useless in local var (still usefull for class attributes)

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -16,19 +16,98 @@ public class JHipsterModulesResources {
...
+    if (noMoreNestedResource(childrensDependencies)) {
+      return Stream.of();
+    }
+    return Stream.concat(
+      childrensDependencies.stream(),
+      childrensDependencies.stream().map(moveToNextNestedResource(modulesResources)).flatMap(t -> t)
+    );
+  }
+
+  private boolean noMoreNestedResource(final Collection<JHipsterModuleResource> childrensDependencies) {
```
DamnClin, line 90: You can remove the final modifier

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
DamnClin: No problem, it's really cool that you take them into account! Thanks

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Done 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Nice! Done 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Done 👍

renanfranca: Thanks @DamnClin, I pushed the new commit 👍

renanfranca:
Hello @DamnClin and @pascalgrimaud ,
I pushed the logging hidden modules slugs feature 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -4,31 +4,137 @@
...
+  private void logHiddenModules(
+    Collection<JHipsterModuleResource> modulesResources,
+    JHipsterHiddenModules hiddenModules,
+    Collection<String> nestedDependenciesSlugs
+  ) {
+    StringJoiner hiddenSlugs = new StringJoiner(", ");
+    modulesResources
+      .stream()
+      .filter(resource -> notAllowed(resource, hiddenModules, nestedDependenciesSlugs))
+      .forEach(resource -> hiddenSlugs.add(resource.slug().get()));
```
DamnClin, line 55: collect(Collectors.joining(", ")) instead of a mutation in a foreach but i don't really get the logic here, I think it's better to just get the hidden modules from the previous method and log them at once.

You can test logs using LogsSpy

renanfranca: Hello @DamnClin,
thanks for the LogsSpy, it is really cool!
I had to use Collectors.partitioningBy to get the hidden and not hidden modules at once. Because the hidden modules are defined by the slugs and the tags, so I had to check with the `JHipsterModulesResource` collection.

Maybe I didn't see the better approach you saw... 🤔 . Let me know if I went too far this time 😅 .

I pushed a new commit this time 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -63,4 +160,25 @@ private void assertKnownSlug(JHipsterModuleSlug slug) {
...
+
+    public void logHiddenResources() {
+      if (hiddenIsPresent()) {
+        String hiddenModulesSlugs = hiddenSlugs();
+        log.info("The following modules are hidden: {}.", hiddenModulesSlugs);
+      }
+    }
+
+    private boolean hiddenIsPresent() {
+      return !hidden.isEmpty();
```
DamnClin, line 177: use the accessors, not the properties (coverrage check will fail)

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -63,4 +160,25 @@ private void assertKnownSlug(JHipsterModuleSlug slug) {
...
   }
+
+  private record DisplayHiddenResources(Collection<JHipsterModuleResource> displayed, Collection<JHipsterModuleResource> hidden) {
+    public Stream<JHipsterModuleResource> displayedStream() {
+      return displayed.stream();
+    }
+
+    public void logHiddenResources() {
+      if (hiddenIsPresent()) {
+        String hiddenModulesSlugs = hiddenSlugs();
```
DamnClin, line 171: you can inline this

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
```diff
@@ -63,4 +160,25 @@ private void assertKnownSlug(JHipsterModuleSlug slug) {
...
+        log.info("The following modules are hidden: {}.", hiddenModulesSlugs);
+      }
+    }
+
+    private boolean hiddenIsPresent() {
```
DamnClin, line 176: `hasHiddenResources`?

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: thanks, done 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: Thanks, @DamnClin for the fast response. pushed new version 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResources.java
renanfranca: much better, i'll do it now 👍

DamnClin (approved):

renanfranca:
Bounty Claimed https://opencollective.com/generator-jhipster/expenses/124093 😃 👍 

Thank you @pascalgrimaud !
Thank you @DamnClin for the fast response and for being patient 😄

pascalgrimaud:
@renanfranca : approved, thanks for your work

PR #5381, opened 2023-02-15
Add test classifier for local build

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/5380


pascalgrimaud:
To this line https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start.sh#L22 you need to add:

```
rm *-javadoc.jar *-sources.jar *-tests.jar
```

Issue #5435, opened 2023-02-21
eslint --fix failed without output (ENOENT). (React)


mbaus:
I am new to jhipster lite.  

But every time I want to check in react code I get the following error:

```
git commit -m 'create beers service'                                                                                                                                                                                      on git:main|●1✚2…1
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ .lintstagedrc.js — 1 file
    ❯ {src/**/,}*.{js,ts,tsx,vue} — 1 file
      ✖ eslint --fix [ENOENT]
    ↓ {src/**/,}*.{md,json,yml,html,css,scss,java,xml} — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ eslint --fix failed without output (ENOENT).
husky - pre-commit hook exited with code 1 (error)
```
At this time the solution is to disabel the husky-pre commit. But is not what I want.

DamnClin:
Hello, there is a file in .jhipster with applied modules and there configuration, can't you provide the content of the file so we can try to reproduce?

mbaus:
Hello,
on my last try with jhipster-lite, i followed this example: https://blog.ippon.tech/how-to-create-a-quick-react-and-angular-app-with-jhipster-lite/

I also used the docker container in the specified version. I can't see any difference to the current version in the concrete example.

First I built the project with ./mvnw (it makes no difference running npm first).

After creating the environment, I created the first three files and try to check them in one by one.

The first hurdle in this approach was the execution rights for .husky/pre-commit were not set. (this happens with the unzip but is not a problem.)

The first file can be checked in with the note that the .husky/pre-commit is not executable.

After I set the rights, the second checkin fails.

Here is the content of the .jhipster/modules/history.json file:
``` 
{
  "actions" : [ {
    "module" : "init",
    "date" : "2023-02-21T21:37:08.987820863Z",
    "properties" : {
      "packageName" : "de.dacomb.library",
      "projectName" : "My Test Library",
      "baseName" : "library",
      "serverPort" : 8080,
      "endOfLine" : "lf",
      "indentSize" : 2
    }
  }, {
    "module" : "maven-java",
    "date" : "2023-02-21T21:37:10.006771445Z",
    "properties" : {
      "packageName" : "de.dacomb.library",
      "projectName" : "My Test Library",
      "baseName" : "library",
[...]

Issue #5477, opened 2023-02-24
JHipsterModulePackageJson: remove dependencies


pascalgrimaud:
Similar to JHipsterModuleJavaDependencies, need to add:
- method to remove dependency
- method to remove devDependency

It will help to resolve correctly https://github.com/jhipster/jhipster-lite/issues/5400

juliensadaoui:
I can do it

Issue #5400, opened 2023-02-18
Generates test files with vitest imports but npm scripts run jest


matthewreed26:
Generated the Vue Typescript as shown:
![Screen Shot 2023-02-18 at 6 06 40 PM](https://user-images.githubusercontent.com/15159154/219903472-5f1d382f-2bc3-4edd-b6d6-3e27768666ca.png)

Initially, `npm run test` runs `jest` but tests (seen on the right) are importing from vitest so the tests fail:
![Screen Shot 2023-02-18 at 6 04 51 PM](https://user-images.githubusercontent.com/15159154/219903468-37da6274-2669-45ec-ae4c-81b5b6680c1a.png)

My change of npm script `test` to run `vitest`, passes all tests and exhibits behavior similar to `jest --watch`:
![Screen Shot 2023-02-18 at 6 05 29 PM](https://user-images.githubusercontent.com/15159154/219903470-614236a7-97d9-4b4b-9cc7-674e17aae2c2.png)

There also appears to be something up with the `AppVue` in `Router.spec.ts` as well:
![Screen Shot 2023-02-18 at 6 28 57 PM](https://user-images.githubusercontent.com/15159154/219904028-0a7ab4ba-970e-47c9-8d73-22c8004169be.png)

pascalgrimaud:
Thanks for reporting this @matthewreed26

I think the problem comes from Typescript, which uses Jest. But Vue doesn't need Jest as it uses Vitest.

The goal of typescript was to init some project to do some Katas, so this module needs to be rework

qmonmert:
@pascalgrimaud you reproduce this bug?

pascalgrimaud:
Adding a bounty on this
The goal would be to remove jest dependency and all related scripts in the module which uses Vitest instead

pascalgrimaud:
@qmonmert : approved

PR #5478, opened 2023-02-25
React: eslint --fix fails

qmonmert:
Fix: #5435


mbaus:
> Fix: #5435

I just downloaded the pull request and created a new project with it.

I'm using the same little [tutorial](https://blog.ippon.tech/how-to-create-a-quick-react-and-angular-app-with-jhipster-lite/) so it's easier for everyone to follow the steps.

jhipster-lite uses React with Typescript. Without commit, the code from the example works.

eslint has now been installed in the fix but without typescript support.

```
git commit -m 'add interface' on git:main|●1
✔ Preparing lint-staged...
❯ Running tasks for staged files...
   ❯ .lintstagedrc.js — 1 file
     ❯ {src/**/,}*.{js,ts,tsx,vue} — 1 file
       ✖ eslint --fix [FAILED]
     ↓ {src/**/,}*.{md,json,yml,html,css,scss,java,xml} — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ eslint --fix:

... /retest/src/main/webapp/app/beers/interface.d.ts
   1:8 error Parsing error: Unexpected token type

✖ 1 problem (1 error, 0 warnings)

husky - pre-commit hook exited with code 1 (error)

```

I'm new to react with typescript and eslint. so I have to watch tutorials to learn. these usually go beyond the basic installation.

For me to learn it's not bad because the examples work. Only a minimal installation is required / desired in jhipster-lite. So I can only test here. :-(

Can you still add support for typescript?

My conclusion You are super fast in providing support and troubleshooting. Excellent. Thanks for that.

qmonmert:
thanks @mbaus 
I updated the PR, you can try and disable rules
image
```
      "@typescript-eslint/no-extra-semi": "off",
      "@typescript-eslint/no-explicit-any": "off",
```

mbaus:
@qmonmert 
OK.

Now it works out of the box for the example app.
I copied the three classes into the project and made the necessary changes to App.tsx and index.tsx.

The app works as expected.

I was able to check the code into Git ( I just had to add one more rule: {/* eslint-disable-next-line-react/jsx-key */})

So the code generated by jhipster-lite works as expected if all modules are installed before starting development.

I will test again later if this works if the friontend is only added after the backend development.

but I'm assuming it will work. as long as I don't use the TypeScript Module (there's already #5400 and #5477 for that)

Thanks for the quick solution.

mbaus:
@qmonmert: Hello,

now I have tested the approach to write a remaining service only once.

I did the following:
1. Download your changes from the pull request
1. Start jhipster-lite
1. Create a new empty folder
1. follow the steps to create a web service using the [video](https://www.youtube.com/watch?v=mEECPRZjajI) so that it is jhipster-lite compliant.
1. now follow Denis Sharpe's React jhipster-lite tutorial again
1. in step 7 install react-jwt there is a crash.
Here he can't check react core into git, the eslint screams.

It's too late today to take a closer look at the error.
But I would have expected that after the changes there would be no problem with the attachment.

It surprises me.

image

qmonmert:
@mbaus I don't know Denis Sharpe's React jhipster-lite tutorial, I don't understand the bug 🤔

mbaus:
@qmonmert: Hi,

Today I took a closer look at the problem.
the reference to the tutorial should only serve to better understand the steps.

So, in the tutorial, a web service is developed using TDD. Only the most minimal dependency is used for this. So that only 6 modules are used via jhipster-lite at the beginning and more modules are added later.

So far so good.

So now we have an existing code base and now we want to add react-jwt. Unfortunately, this no longer works at the moment.

Jhipster-lite is the first to add reac-core to the project.
Then this leads to jhipster-lite 16 wanting to add or change files. When trying to check this into git, eslint raises an alarm and react-jwt is no longer added at all.

I've since found out the reason for this.

With a basic installation, no husky is installed. That's how it works.

In the current variant, the react-core project only brings the eslint dependencies with it. But these are not installed yet.

A simple npm install before checkin is missing at this point. That would solve the problem.

I hope I was able to describe the problem in an understandable way.

1. Install all components in one go and it works without any problems.
2. Install individual components and react later. Then eslint is armed and breaks the build until npm install is called.

Sorry, I didn't come up with this simple solution tonight.

Is it possible to run an npm install before committing react-core ? maybe only if husky is already active?

Best regards

qmonmert:
@mbaus ok the problem was that you don't run `npm install` after `apply new`, I don't think that JHLite must run this command cc @pascalgrimaud

mbaus:
@qmonmert No, you are wrong.
At the point where I press the button, the error arises.

husky checks the code to be installed. Since the necessary modules are first added in this step, they are not available.
I can't run npm install at this point.

mbaus:
@qmonmert The next error I found is only marginal.

react-jwt comes with a test class with an empty function.
It's about the src/test/javascript/spec/login/primary/loginModal/index.test.tsx class

By adding the following rule "@typescript-eslint/no-empty-function": "off", to .eslintrc.js, the project can be easily checked in.

And very important. you can only become active in the two steps afterwards. the app fails!!!

mbaus:
> @mbaus ok the problem was that you don't run `npm install` after `apply new`, I don't think that JHLite must run this command cc @pascalgrimaud

@pascalgrimaud: the apply new fails because an npm install is required/mandentory before the "applay new" process does a git commit.

pascalgrimaud:
Lot of discussion here.
Thanks guys for trying to solve this issue, I'll try to find time to provide some advices on this

mbaus:
> Lot of discussion here. Thanks guys for trying to solve this issue, I'll try to find time to provide some advices on this

Hello,
I'm sorry if I'm causing too much trouble.

I was just trying to test the changes against quality assurance criteria. Which is not easy for me because I don't know the specifications. What I know from all jhipster-lite descriptions is that the modules should only be added when needed.
Based on this I created two test cases for the module from the bug. One approach is to activate all modules immediately. This works now.
Another approach is to develop the backend first and then add the frontend. That only works to a limited extent.

If React-Core is added later, the jhipster-lite commit will not work because the eslint plugins are only added to package.json here. The commit fails without calling npm installsall. or you can disable the commit hook for the commits from the jhipster generator.

When adding react-jwt there is an empty method in the added classes. Eslint doesn't accept that. Again, the commit fails. Here it would help to include this rule.

In other places I have found anomalies in this regard. But I don't want to open an error for each one individually.

I am happy to test further and try to document the deviations. To do this, however, I should know what the goals of the individual modules are. In some places there are slight differences that are not noticeable at first glance.

To document or describe it here goes too far, I think.

My concern is to move the project forward.
If I'm annoying a developer with this, I'm sorry.

pascalgrimaud:
no @mbaus : never be sorry to participe to an open source project, to try it, to give feedbacks, etc
Your help is really welcome, so thanks so much to share your experience :)

Just be patient as it can take time to find time to solve it ahaha, as it's not our work day :p

mbaus:
OK. 

This week I'm building a playground using the jhipster-lite framework to find out.

I created a new git repo. and provide the link here: [https://github.com/mbaus/jhl-test](https://github.com/mbaus/jhl-test)

Based on the status of the main branch, I was able to verify the following.

If you use Du React in a new project there are no problems. If Reeact has already been installed and you only go back to the main branch, there are no problems either.

If you start from the main branch and freshly create the node_module folder, then adding React will not work. The commit goes wrong. As a workaround, do an npm install after the failed add and then try again. or check in manually.

In any case, before adding react-jwt, an additional rule must be specified, otherwise the addition will also fail.

So far so good. Because you wanted to look at that again internally.

I have another problem now. I'm trying to use the testing library.

with an npm run test or npm run test:test (to avoid the coverage report) it works so far.
But an npm run build fails.

I get the error:

src/test/javascript/spec/common/primary/app/App.spec.tsx:19:7 - error TS2339: Property 'toHaveTextContent' does not exist on type 'Assertion<HTMLElement>'.

I suspect I made a misconfiguration here.

@qmonmert: Can you have a look at the code if you see my mistake?
Many thanks in advance.

mbaus:
@qmonmert : I finde the Mistake.

I have to change the build config.
this must be something like this:
"build": "tsc -p tsconfig.node.json && vite build --emptyOutDir",

pascalgrimaud:
I updated the bounty, as it was a lot of time to analyze and for the dozen of other PRs @qmonmert 
thanks for your work on JHLite

Issue #2246, opened 2022-06-27
Kipe module


DamnClin:
Create a module to add kipe

pascalgrimaud:
this one is for you, @DamnClin :)

DamnClin:
Yep, working on it :D most of the current PR are prerequisite for that :D

pascalgrimaud:
Increasing the bounty as it's a lot of work

PR #5551, opened 2023-03-05
Kipe expression module

DamnClin:
Related to https://github.com/jhipster/jhipster-lite/issues/2246

This is one part of kipe, the other part is the authorization matrice


src/main/resources/generator/server/springboot/mvc/security/kipe/expression/kipe-expression.md.mustache
```diff
@@ -0,0 +1,20 @@
+# Kipe Expression
+
+Kipe can be pronounced, in french, as "qui peu" which means: who can.
```
NolwennD, line 3:
```suggestion
Kipe can be pronounced, in french, as "qui peut" which means: who can.  
```

PR #4586, opened 2022-12-05
Remove zalando problems: Upgrade to Spring Boot 3

DamnClin:
Close https://github.com/jhipster/jhipster-lite/issues/4111
Close https://github.com/jhipster/jhipster-lite/issues/4066
Close https://github.com/jhipster/jhipster-lite/issues/4351
Close https://github.com/jhipster/jhipster-lite/issues/3975
Close https://github.com/jhipster/jhipster-lite/issues/3750

PR #4366, opened 2022-11-11
configure global jwt bearer auth for swagger documentation

atomfrede:
This configures the swagger/openapi docs to use jwt bearer auth globally/for all endpoints

closes #4273

PR #5591, opened 2023-03-09
remove springdoc-jwt module and improve documentation

### LINKED ISSUE #4273: springdoc-jwt : authorization header didn't exist in requests
Hi,
I tried to use **springdoc-jwt** module on the dummy appplication

Then after my application started I pass the bearer token (got with the [/api/authenticate] endpoint with admin/admin credentials) into an authentication form with swagger UI:
image

When I invoke the getBeers enpoint :

**Actual result**:

curl -X 'GET' 'http://localhost:8080/api/beers' -H 'accept: */*'

Bearer token didn't exist in the request and the request fails with {"title": "Unauthorized",  "status": 401, "detail": "Full authentication is required to access this resource"}

**Expected result**:

curl -X 'GET' \
  'http://localhost:8080/api/beers' \
  -H 'accept: */*' \
  -H 'Authorization: Bearer eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJhZG1pbiIsImF1dGgiOiJST0xFX0FETUlOIiwiZXhwIjoxNjk5Mzk2Njg2fQ.qY_SdBezB1DL70w1gKLtu6UFgXwmtr5RFrfpfQkaAAdT7YXQkGwpPOTwnp8GghHGk7vjf9XaWYjz4VhG3x4iyw'

I have tried to fix that with adding  , security = @SecurityRequirement(name = "bearer-jwt") on each endpoint and It's working but I m not sure it's the good way

renanfranca:
Fix #4273


renanfranca:
Hello @DamnClin,
I create this pull request because I agree with [the message](https://github.com/jhipster/jhipster-lite/pull/4366#issuecomment-1344457980) you wrote:
> This behavior change with https://github.com/jhipster/jhipster-lite/pull/4586 (but no idea if it's better or worse...). When you query a WebService that need authentication you'll have a browser prompt with authentication, you can authenticate here and it'll work (this is the behavior for any autenticated endpoint). Is it ok ? IF so we need to remove the "Authorize" button wich become pretty useless

In my opinion, it's a practical workflow to be prompt to log in and I don't need to generate the token first and then add it using the swagger authorization button. I added some lines to the documentation based on my experience while using it.

DamnClin:
No opinion on that, I'm really not using this authentication scheme. @pascalgrimaud got an advice?

pascalgrimaud:
@renanfranca : don't understand why you delete the module. The fix should not be too hard to fill the token in header

PR #5708, opened 2023-03-20
remove module count from header

### LINKED ISSUE #5550: Remove modules counts header
As we don't have the persisted stats on lite.jhipster.tech this is more missleading than anything else...

msivade:
remove applied module count from header component and view and remove dead code related to statistics in the frontend

Fix #5550


msivade:
I removed module count from frontend (and chose to delete dead code in the frontend). Do we need to remove unused REST API related to statistics as well?

PR #5853, opened 2023-04-01
WIP: Remove Jest dependency

### LINKED ISSUE #5400: Generates test files with vitest imports but npm scripts run jest

qmonmert:
Fix #5400


qmonmert:
@pascalgrimaud `optional-typescript` module is really necessary? 🤔

pascalgrimaud:
@qmonmert : it was coded by a user, I don't remember who and why. Anyway, I don't use it :) But why this question ?

qmonmert:
@pascalgrimaud because (except for Angular) Jest is only necessary for `optional-typescript` (see my second commit) and I don't understand why we have this module 🙂

pascalgrimaud:
Ok so no worry, let this module like this

qmonmert:
@pascalgrimaud ok so you can review when you want

PR #5944, opened 2023-04-11
add startup command in README.md file

medoer:
add startup command for consul/pulsar/kafka/akhq modules


pascalgrimaud:
Thanks @medoer :-)

PR #6091, opened 2023-04-27
Share module properties between landscape and patch screens

### LINKED ISSUE #3371: Share module properties between landscape and patch screens
In there current version module properties are reset each time the user go to another page which can be very annoying. An idea can be to store that

renanfranca:
Fix #3371

### DIFF
--- a/src/main/webapp/app/module/domain/ModuleParametersRepository.ts
+++ b/src/main/webapp/app/module/domain/ModuleParametersRepository.ts
@@ -1,0 +1,6 @@
+import { ModuleParameterType } from '@/module/domain/ModuleParameters';
+
+export interface ModuleParametersRepository {
+  store(map: Map<string, ModuleParameterType>): void;
+  get(): Map<string, ModuleParameterType>;
+}
--- a/src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
+++ b/src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
@@ -13,10 +13,10 @@ import { ModulePropertyDefinition } from '@/module/domain/ModulePropertyDefiniti
 import { ModulePropertiesFormVue } from '../module-properties-form';
 import { ModulePropertyKey } from '@/module/domain/ModulePropertyKey';
 import { ProjectActionsVue } from '../project-actions';
-import { ModuleParameterType } from '@/module/domain/ModuleParameters';
 import { ModuleParameter } from '@/module/domain/ModuleParameter';
 import { ModuleParametersVue } from '../module-parameters';
 import { castValue } from '../PropertyValue';
+import { ModuleParametersRepository } from '@/module/domain/ModuleParametersRepository';
 
 export default defineComponent({
   name: 'ModulesPatchVue',
@@ -35,7 +35,8 @@ export default defineComponent({
     const operationInProgress = ref(false);
     const folderPath = ref('');
     const selectedModule = ref<ComponentModule>();
-    const moduleParameters = ref(new Map<string, ModuleParameterType>());
+    const moduleParameters = inject('moduleParameters') as ModuleParametersRepository;
+    const moduleParametersValues = ref(moduleParameters.get());
     const commitModule = ref(true);
     const appliedModules = ref([] as string[]);
     let searchedText = '';
@@ -105,7 +106,7 @@ export default defineComponent({
     };
 
     const isNotSet = (propertyKey: string): boolean => {
-      const value = moduleParameters.value.get(propertyKey);
+      const value = moduleParametersValues.value.get(propertyKey);
 
       if (typeof value === 'string') {
         return empty(value);
@@ -133,11 +134,13 @@ export default defineComponent({
     };
 
     const updateProperty = (property: ModuleParameter): void => {
-      moduleParameters.value.set(property.key, property.value);
+      moduleParametersValues.value.set(property.key, property.value);
+      moduleParameters.store(moduleParametersValues.value);
     };
 
     const deleteProperty = (key: ModulePropertyKey): void => {
-      moduleParameters.value.delete(key);
+      moduleParametersValues.value.delete(key);
+      moduleParameters.store(moduleParametersValues.value);
     };
 
     const mandatoryProperties = (module: string): ModulePropertyDefinition[] => {
@@ -222,7 +225,7 @@ export default defineComponent({
         .apply(new ModuleSlug(module), {
           projectFolder: folderPath.value,
           commit: commitModule.value,
-          parameters: moduleParameters.value,
+          parameters: moduleParametersValues.value,
         })
         .then(() => {
           operationInProgress.value = false;
@@ -249,13 +252,14 @@ export default defineComponent({
 
       projectHistory.properties.forEach(property => {
         if (unknownProperty(property.key)) {
-          moduleParameters.value.set(property.key, property.value);
+          moduleParametersValues.value.set(property.key, property.value);
         }
       });
+      moduleParameters.store(moduleParametersValues.value);
     };
 
     const unknownProperty = (key: string) => {
-      return !moduleParameters.value.has(key);
+      return !moduleParametersValues.value.has(key);
     };
 
     const appliedModule = (slug: string): boolean => {
@@ -271,7 +275,7 @@ export default defineComponent({
       disabledApplication,
       mandatoryProperties,
       optionalProperties,
-      moduleParameters,
+      moduleParametersValues,
       displayedModulesCount,
       totalModulesCount,
       isTagSelected,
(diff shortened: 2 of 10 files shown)


renanfranca:
Hello @DamnClin,
I implemented it using LocalStorage. When you have the time, please, let me know if it is enough.
Thank you in advance 😄 👍

DamnClin:
I'm afraid I won't be able to have a look at it soon... @Gnuk can you check?

src/main/webapp/app/module/domain/ModuleParametersRepository.ts
```diff
@@ -0,0 +1,8 @@
+import { Ref } from 'vue';
+
+export interface ModuleParametersRepository<K, V> {
+  set(key: K, value: V): void;
+  getAll(): Map<K, V>;
+  delete(key: K): void;
+  getRef(): Ref<Map<K, V>>;
```
Gnuk, line 7: I don't understand why you're using a infrastructure notion on your Repository, I think it's a hexagonal architecture violation. Because inside the domain, it's not normal to have things like Vue.

src/main/webapp/app/module/domain/ModuleParametersRepository.ts
```diff
@@ -0,0 +1,8 @@
+import { Ref } from 'vue';
+
+export interface ModuleParametersRepository<K, V> {
```
Gnuk, line 3: Maybe I'm wrong but you seems to not use the generics here when you're trying to implements this repository.

src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
```diff
@@ -35,7 +36,8 @@ export default defineComponent({
     const operationInProgress = ref(false);
     const folderPath = ref('');
     const selectedModule = ref<ComponentModule>();
-    const moduleParameters = ref(new Map<string, ModuleParameterType>());
+    const moduleParameters = inject('moduleParameters') as ModuleParametersRepository<string, ModuleParameterType>;
+    const moduleParametersValues = moduleParameters.getRef();
```
Gnuk, line 40: It's a bit confusing to me to see a usage of a ref from a repository cause it's a reactive notion from the framework and it's not agnostic.

Gnuk (changes requested):

src/main/webapp/app/module/domain/ModuleParametersRepository.ts
renanfranca: Hello. @Gnuk !
Thank you for the fast response 👏👏

You are right! I will try another approach which avoid hexagonal architecture violation .

It's my first time working with Vie framework.
I am going to implement it as soon as possible,,,,😃

src/main/webapp/app/module/domain/ModuleParametersRepository.ts
renanfranca: Hou are right, I should use in the interface the same type i use  at the repository.

src/main/webapp/app/module/primary/modules-patch/ModulesPatch.component.ts
renanfranca: I understand, thank you! I will try a different approach as soon as possible 👍

renanfranca:
Hello @Gnuk,
I pushed the changes in a new commit `avoid hexagonal architecture violation`. Please, let me know what you think.

src/main/webapp/app/module/secondary/LocalStorageModuleParametersRepository.ts
```diff
@@ -0,0 +1,18 @@
+import { ModuleParametersRepository } from '../domain/ModuleParametersRepository';
+import { ModuleParameterType } from '../domain/ModuleParameters';
+
+export class LocalStorageModuleParametersRepository implements ModuleParametersRepository {
+  private readonly STORAGE_KEY = 'moduleParameters';
+
+  store(map: Map<string, ModuleParameterType>): void {
+    localStorage.setItem(this.STORAGE_KEY, JSON.stringify(Array.from(map.entries())));
```
Gnuk, line 8: Please inject localStorage using a constructor and pass it when you make the instance from the main.

src/test/javascript/spec/module/primary/modules-patch/ModulesPatchComponent.spec.ts
```diff
@@ -682,6 +689,20 @@ const repositoryWithProjectFolders = (): ProjectFoldersRepositoryStub => {
...
+const repositoryWithModuleParameters = (): LocalStorageModuleParametersRepository => {
+  const moduleParameters = new LocalStorageModuleParametersRepository();
+  const stubLocalStorage = {
+    getItem: sinon.stub(),
+    setItem: sinon.stub(),
+    removeItem: sinon.stub(),
+    clear: sinon.stub(),
+    length: 0,
+  };
+  sinon.stub(window, 'localStorage').value(stubLocalStorage);
```
Gnuk, line 701: With a constructor field you have no longer to add a stub on global window.

Gnuk (changes requested):

src/main/webapp/app/module/secondary/LocalStorageModuleParametersRepository.ts
renanfranca: Thanks for the clear instruction @Gnuk,
that's much better!

src/test/javascript/spec/module/primary/modules-patch/ModulesPatchComponent.spec.ts
renanfranca: Thanks!
Nice! Much better!

renanfranca:
Hi @Gnuk,
I pushed a new commit `create a constructor to receive localStorage as a parameter` to be reviewed.

src/test/javascript/spec/module/secondary/LocalStorageModuleParametersRepository.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { LocalStorageModuleParametersRepository } from '@/module/secondary/LocalStorageModuleParametersRepository';
+import { ModuleParameterType } from '@/module/domain/ModuleParameters';
+import { describe, it, expect } from 'vitest';
+
+describe('LocalStorageModuleParametersRepository', () => {
+  const mockLocalStorage = (() => {
```
Gnuk, line 6: Maybe you're creating a fake, so you can name this fakeStorage. Then you don't need to define a localStorage on the window because you're mutating the existing one and it's not needed because now you can just inject your fake. Then you can use it as a function instead of calling it directly after the definition.

At the end you may have something like this:

```typescript
const fakeStorage = (): Storage => {
  //...
};
```

And to use it from a test case (inside an `it`):

```typescript
const storage = fakeStorage();
// ...
```

Gnuk (changes requested):

src/test/javascript/spec/module/secondary/LocalStorageModuleParametersRepository.spec.ts
renanfranca: Thanks, @Gnuk for the clear instructions!
I did as you requested. I pushed a new commit  `use a fakeStorage instead of mockLocalStorage` to be reviewed when you have the time 👍

renanfranca:
Hello @Gnuk,
I remove the sonar bug fix from here because it is handled by this issue https://github.com/jhipster/jhipster-lite/issues/6146 . So I rebase my changes in a single commit, let me know if you think this issue is fixed.

Thank you 😄

src/main/webapp/app/module/secondary/LocalStorageModuleParametersRepository.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+  store(map: Map<string, ModuleParameterType>): void {
+    this.localStorage.setItem(this.STORAGE_KEY, JSON.stringify(Array.from(map.entries())));
+  }
+
+  get(): Map<string, ModuleParameterType> {
+    const storedValue = this.localStorage.getItem(this.STORAGE_KEY);
+    if (storedValue) {
+      return new Map(JSON.parse(storedValue));
+    } else {
```
Gnuk, line 19: Else is not needed because you use return from the previous condition

Gnuk (dismissed):

src/main/webapp/app/module/secondary/LocalStorageModuleParametersRepository.ts
renanfranca: Thank you @Gnuk!

I did as you requested and I pushed a new version with that change 😃 👍

Gnuk (approved):

renanfranca:
@Gnuk, I am sorry! I pressed the request review button by mistake! I didn't know what that button do. Have a nice weekend 😃👍

Gnuk (approved):

renanfranca:
Bounty Claimed https://opencollective.com/generator-jhipster/expenses/138076
Thank you @pascalgrimaud!
Thank you @Gnuk for the detailed reviews and I learn a lot about vue and frontend!

pascalgrimaud:
@Gnuk : if you want to claim 100$ bounty for time spending to review it, plz do it and link here

PR #6114, opened 2023-05-01
ci: auto sonar support.

zhyd1997:
https://github.com/jhipster/generator-jhipster/issues/19467#issuecomment-1528772294


zhyd1997:
Automatically run result in the 1st time:

Screenshot 2023-05-01 at 15 29 22

pascalgrimaud:
This PR should be replaced by this one: https://github.com/jhipster/jhipster-lite/pull/372/files

Both can be tested in parallele:
- custom sonar analysis
- sonarcloud

So with this, we'll see if the new Sonarcloud works or is still broken

zhyd1997:
Continue to https://github.com/jhipster/jhipster-lite/pull/6134

PR #6314, opened 2023-05-18
rest local properties storage

### LINKED ISSUE #6279: Local properties storage do no reset
The local properties storage do not reset when changing folder wich is really annoying since, when you start a new application, you still have properties from the previous one. We need to keep properties only for the same application

renanfranca:
Fix #6279

### DIFF
--- a/src/main/webapp/app/module/primary/landscape/Landscape.component.ts
+++ b/src/main/webapp/app/module/primary/landscape/Landscape.component.ts
@@ -52,10 +52,9 @@ export default defineComponent({
 
     const emphasizedModule = ref<ModuleSlug>();
 
-    const folderPath = ref('');
-
     const moduleParameters = inject('moduleParameters') as ModuleParametersRepository;
-    const moduleParametersValues = ref(moduleParameters.get());
+    const folderPath = ref(moduleParameters.getCurrentFolderPath());
+    const moduleParametersValues = ref(moduleParameters.get(folderPath.value));
 
     let commitModule = true;
 
@@ -72,11 +71,20 @@ export default defineComponent({
         .landscape()
         .then(response => loadLandscape(response))
         .catch(error => console.error(error));
-      projectFolders
-        .get()
-        .then(projectFolder => (folderPath.value = projectFolder))
-        .catch(error => console.error(error));
+      loadProjectFolders();
     });
+
+    const loadProjectFolders = (): void => {
+      if (folderPath.value.length === 0) {
+        projectFolders
+          .get()
+          .then(projectFolder => {
+            folderPath.value = projectFolder;
+            moduleParametersValues.value = moduleParameters.get(folderPath.value);
+          })
+          .catch(error => console.error(error));
+      }
+    };
 
     const startGrabbing = (mouseEvent: MouseEvent): void => {
       if (mouseEvent.preventDefault) {
@@ -339,6 +347,7 @@ export default defineComponent({
 
     const updateFolderPath = (path: string): void => {
       folderPath.value = path;
+      moduleParametersValues.value = moduleParameters.get(folderPath.value);
     };
 
     const projectFolderUpdated = (): void => {
@@ -358,7 +367,7 @@ export default defineComponent({
           moduleParametersValues.value.set(property.key, property.value);
         }
       });
-      moduleParameters.store(moduleParametersValues.value);
+      moduleParameters.store(folderPath.value, moduleParametersValues.value);
     };
 
     const unknownProperty = (key: string) => {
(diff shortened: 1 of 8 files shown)


src/main/webapp/app/module/secondary/LocalStorageModuleParametersRepository.ts
```diff
@@ -1,22 +1,39 @@
...
 import { ModuleParameterType } from '../domain/ModuleParameters';
 
+const STORAGE_KEY_MODULE_PARAMETERS = 'moduleParameters';
+const STORAGE_KEY_FOLDER_PATH = 'folderPath';
+
 export class LocalStorageModuleParametersRepository implements ModuleParametersRepository {
-  private readonly STORAGE_KEY = 'moduleParameters';
   private readonly localStorage: Storage;
 
   constructor(localStorage: Storage) {
```
DamnClin, line 10: Why storing only for one path and having 2 keys? Why not storing a map of properties, one for each path?

renanfranca: Thank you for the feedback @DamnClin,

I've worked diligently to fix this problem while doing my best to ensure that the user experience of JHipster Lite remains intact. If I store a map of properties for each path It won't be possible to load the original properties from the module's history once it was stored at the local storage. Because of this functionality:
- 'Should not replace user setted properties from history'

In the way I implemented it, every time you reload the application (press F5) it shares the properties between landscape and modules patch menu options, and if you set a folder path that has jhilite application it will load the module's history.

Let me know what you think about it 👍

DamnClin:
Finally toke some time to test your branch locally... Sorry for the delay. Still not doing what I expect. The test: 

* build two projects (A and B) with different properties and modules ;
* Restart JHLite ;
* Set path to project A -> should have modules and properties from A (working) ;
* Set path to project B -> should have modules and properties from B: not working, I have modules from B but properties from A which is really really confusing

renanfranca:
> Finally toke some time to test your branch locally... Sorry for the delay. Still not doing what I expect. The test:
> 
> * build two projects (A and B) with different properties and modules ;
> * Restart JHLite ;
> * Set path to project A -> should have modules and properties from A (working) ;
> * Set path to project B -> should have modules and properties from B: not working, I have modules from B but properties from A which is really really confusing

Thank you for your feedback @DamnClin with detailed steps!

I pushed a new commit with changes to work as you expected 👍 

Let me know what you think about it ✌️

DamnClin:
> > Finally toke some time to test your branch locally... Sorry for the delay. Still not doing what I expect. The test:
> > 
> > * build two projects (A and B) with different properties and modules ;
> > * Restart JHLite ;
> > * Set path to project A -> should have modules and properties from A (working) ;
> > * Set path to project B -> should have modules and properties from B: not working, I have modules from B but properties from A which is really really confusing
> 
> Thank you for your feedback @DamnClin with detailed steps!
> 
> I pushed a new commit with changes to work as you expected +1
> 
> Let me know what you think about it v

I think this is still really really fragile, why not storing properties by path in local storage? 

The local storage check should not be exposed from the repository, this should be an implementation detail in the repository itself, the temporal coupling in this implementation makes it really hard to use well

renanfranca:
Thanks @DamnClin,

> I think this is still really really fragile, why not storing properties by path in local storage?

I am going to implement this way, without the need to clear the local storage. And when loading the project history it will overwrite what is in local storage for the current project folder path.

> The local storage check should not be exposed from the repository, this should be an implementation detail in the repository itself, the temporal coupling in this implementation makes it really hard to use well

Is it referring to this code:
`moduleParametersValues.value.clear();`?

I am sorry, I didn't get what do you meaning about:
> The local storage check should not be exposed from the repository

Thank you very much in advance 😃 👍

DamnClin:
> I am going to implement this way, without the need to clear the local storage. And when loading the project history it will overwrite what is in local storage for the current project folder path.

In fact it's not an override, it's a merge because you should take local storage + remote properties with a priority to local storage for the folder.

To use it you should not need any invocation related to local storage, as a client of the repository you just want to get the properties :)

It should not be possible to missuse a repo

renanfranca:
> > I am going to implement this way, without the need to clear the local storage. And when loading the project history it will overwrite what is in local storage for the current project folder path.
> 
> In fact it's not an override, it's a merge because you should take local storage + remote properties with a priority to local storage for the folder.
> 
> To use it you should not need any invocation related to local storage, as a client of the repository you just want to get the properties :)
> 
> It should not be possible to missuse a repo

Thank you for the instructions and for your time @DamnClin!

I pushed a new version, please, let me know what you think about it 😃 ✌️!

src/main/webapp/app/module/primary/landscape/Landscape.component.ts
```diff
@@ -338,6 +347,8 @@ export default defineComponent({
 
     const updateFolderPath = (path: string): void => {
       folderPath.value = path;
+      moduleParameters.storeCurrentFolderPath(folderPath.value);
```
DamnClin, line 350: It is still really easy to missuse the repo since there is a temporal coupling here. The idea is to have a transparent local storage on ModuleParametersRepository, this implementation is way too fragile and have very low chance to work over time

renanfranca: Thank you for pointing this out @DamnClin! I pushed a new version fixing everyplace I did this 👍

DamnClin:
Tried it locally, the behavior looks as expected, thanks for the hard work here!

renanfranca:
> Tried it locally, the behavior looks as expected, thanks for the hard work here!

You are welcome! Thank you too! 😄 👍

PR #6632, opened 2023-06-21
Sonarqube 10.1.0-community

qmonmert:
Fix: #6626


src/main/resources/generator/server/sonar/sonar-project.properties.mustache
```diff
@@ -57,3 +57,7 @@ sonar.issue.ignore.multicriteria.S6206.ruleKey=java:S6206
 # Rule: Delete commented out code in liquibase example
 sonar.issue.ignore.multicriteria.S125.resourceKey=src/main/resources/config/liquibase/master.xml
 sonar.issue.ignore.multicriteria.S125.ruleKey=xml:S125
+
+# Rule: The Singleton design pattern should be used with care
+sonar.issue.ignore.multicriteria.S6548.resourceKey=src/main/java/**/*
+sonar.issue.ignore.multicriteria.S6548.ruleKey=java:S6548
```
pascalgrimaud, line 63: I think this rule can be important, so I'd add it locally, using suppress warning, instead of globally

qmonmert: done 🙂

PR #6687, opened 2023-06-27
Add criteria to stats endpoint

kartiktayal:


kartiktayal:
All checks have completed. Can someone please help with license/cla, I already have it checked. Not sure why gitubActionsbot is getting picked up.

kartiktayal:
@dependabot rebase

pascalgrimaud:
@kartiktayal : it's because you have some other commits, related to github actions bot, in your pull request. So the git history is not really clear. Maybe, you can try this:
- rebase against upstream/main
- start from upstream/main then apply your commits

kartiktayal:
> @kartiktayal : it's because you have some other commits, related to github actions bot, in your pull request. So the git history is not really clear. Maybe, you can try this:
> 
> * rebase against upstream/main
> * start from upstream/main then apply your commits

I have already rebased quite a number of times and resolved the conflicts. I keep getting merge conflicts which is why I asked for help. Let me try again if you think this will resolve the issue

pascalgrimaud:
Thanks @kartiktayal 
What is strange is your conflicts, as you should not have any conflicts, because the class you improve should not be impacted:
image

The statistic package is unchanged since 6 months, so you should not have conflict with main branch:

image

kartiktayal:
@pascalgrimaud , I have rebased the branch but still it does not work, anything I am missing here?

pascalgrimaud:
It's git manipulation :-D
You could try this:
- reset your current branch to be the same than upstream/main
- apply this git diff:

```diff
diff --git a/src/main/java/tech/jhipster/lite/statistic/application/StatisticsApplicationService.java b/src/main/java/tech/jhipster/lite/statistic/application/StatisticsApplicationService.java
index e4913913c..165790b81 100644
--- a/src/main/java/tech/jhipster/lite/statistic/application/StatisticsApplicationService.java
+++ b/src/main/java/tech/jhipster/lite/statistic/application/StatisticsApplicationService.java
@@ -4,6 +4,7 @@ import org.springframework.stereotype.Service;
 import tech.jhipster.lite.statistic.domain.AppliedModule;
 import tech.jhipster.lite.statistic.domain.Statistics;
 import tech.jhipster.lite.statistic.domain.StatisticsRepository;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
 
 @Service
 public class StatisticsApplicationService {
@@ -18,7 +19,7 @@ public class StatisticsApplicationService {
     statistics.save(moduleApplied);
   }
 
-  public Statistics get() {
-    return statistics.get();
+  public Statistics get(StatisticsCriteria criteria) {
+    return statistics.get(criteria);
   }
 }
diff --git a/src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java b/src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java
index b832b1197..ac7acf08d 100644
--- a/src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java
+++ b/src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java
@@ -1,7 +1,14 @@
 package tech.jhipster.lite.statistic.domain;
 
+import org.apache.commons.lang3.ObjectUtils;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
 public interface StatisticsRepository {
   void save(AppliedModule moduleApplied);
 
-  Statistics get();
+  Statistics get(StatisticsCriteria criteria);
+
+  default boolean isAnyCriteriaApplied(StatisticsCriteria criteria) {
+    return (ObjectUtils.isNotEmpty(criteria) && (ObjectUtils.anyNotNull(criteria.startTime(), criteria.endTime(), criteria.moduleSlug())));
+  }
 }
diff --git a/src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java b/src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
new file mode 100644
index 000000000..823150ea1
--- /dev/null
+++ b/src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
@@ -0,0 +1,11 @@
+package tech.jhipster.lite.statistic.domain.criteria;
+
+import java.io.Serializable;
+import java.time.Instant;
+
+/**
+ * Criteria class for {@link tech.jhipster.lite.statistic.domain.Statistics}. This class is used
+ * to receive all the possible filtering options from
+ * the Http GET request parameters.
+ */
+public record StatisticsCriteria(Instant startTime, Instant endTime, String moduleSlug) implements Serializable {}
diff --git a/src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/StatisticsResources.java b/src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/StatisticsResources.java
index b4fb37186..d4ef91c80 100644
--- a/src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/StatisticsResources.java
+++ b/src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/StatisticsResources.java
@@ -7,6 +7,7 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import tech.jhipster.lite.statistic.application.StatisticsApplicationService;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
 
 @RestController
 @Tag(name = "Statistics")
@@ -21,7 +22,7 @@ class StatisticsResources {
 
   @GetMapping
   @ApiResponse(description = "JHipster lite usage statistics", responseCode = "200")
-  ResponseEntity<RestStatistics> getStatistics() {
-    return ResponseEntity.ok(RestStatistics.from(statistics.get()));
+  ResponseEntity<RestStatistics> getStatistics(StatisticsCriteria criteria) {
+    return ResponseEntity.ok(RestStatistics.from(statistics.get(criteria)));
   }
 }
diff --git a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java
index 28302b207..0468a8f7b 100644
--- a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java
+++ b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java
@@ -2,12 +2,14 @@ package tech.jhipster.lite.statistic.infrastructure.secondary;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import org.springframework.stereotype.Repository;
 import tech.jhipster.lite.error.domain.Assert;
 import tech.jhipster.lite.statistic.domain.AppliedModule;
 import tech.jhipster.lite.statistic.domain.Statistics;
 import tech.jhipster.lite.statistic.domain.StatisticsRepository;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
 
 @Repository
 @WithoutMongoDB
@@ -23,8 +25,22 @@ class InMemoryStatisticsRepository implements StatisticsRepository {
   }
 
   @Override
-  public Statistics get() {
-    return new Statistics(appliedModules.size());
+  public Statistics get(StatisticsCriteria criteria) {
+    int appliedModulesCount = 0;
+    if (isAnyCriteriaApplied(criteria)) {
+      for (AppliedModule am : appliedModules) {
+        if (
+          (Objects.isNull(criteria.startTime()) || am.date().isAfter(criteria.startTime())) &&
+          (Objects.isNull(criteria.endTime()) || am.date().isBefore(criteria.endTime())) &&
+          (Objects.isNull(criteria.moduleSlug()) || am.module().slug().equals(criteria.moduleSlug()))
+        ) {
+          appliedModulesCount++;
+        }
+      }
+    } else {
+      appliedModulesCount = appliedModules.size();
+    }
+    return new Statistics(appliedModulesCount);
   }
 
   void clear() {
diff --git a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
index 24d3a5b59..1ae0bf68d 100644
--- a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
+++ b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
@@ -5,6 +5,7 @@ import tech.jhipster.lite.error.domain.Assert;
 import tech.jhipster.lite.statistic.domain.AppliedModule;
 import tech.jhipster.lite.statistic.domain.Statistics;
 import tech.jhipster.lite.statistic.domain.StatisticsRepository;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
 
 @WithMongoDB
 @Repository
@@ -24,7 +25,11 @@ class MongoDBStatisticsRepository implements StatisticsRepository {
   }
 
   @Override
-  public Statistics get() {
-    return new Statistics(statistics.count());
+  public Statistics get(StatisticsCriteria criteria) {
+    if (isAnyCriteriaApplied(criteria)) {
+      return new Statistics(statistics.countByModuleSlugAndDateBetween(criteria.moduleSlug(), criteria.startTime(), criteria.endTime()));
+    } else {
+      return new Statistics(statistics.count());
+    }
   }
 }
diff --git a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/SpringDataMongoDBStatisticsRepository.java b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/SpringDataMongoDBStatisticsRepository.java
index 3bb2f3ae4..3e672d1d9 100644
--- a/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/SpringDataMongoDBStatisticsRepository.java
+++ b/src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/SpringDataMongoDBStatisticsRepository.java
@@ -1,9 +1,13 @@
 package tech.jhipster.lite.statistic.infrastructure.secondary;
 
+import java.time.Instant;
 import java.util.UUID;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Repository;
 
 @WithMongoDB
 @Repository
-interface SpringDataMongoDBStatisticsRepository extends CrudRepository<AppliedModuleDocument, UUID> {}
+interface SpringDataMongoDBStatisticsRepository extends CrudRepository<AppliedModuleDocument, UUID> {
+  long countByModuleSlugAndDateBetween(@Nullable String moduleSlug, @Nullable Instant startTime, @Nullable Instant endTime);
+}
diff --git a/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java b/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
new file mode 100644
index 000000000..4f258fd6c
--- /dev/null
+++ b/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
@@ -0,0 +1,85 @@
+package tech.jhipster.lite.statistic.infrastructure.secondary;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Instant;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import tech.jhipster.lite.UnitTest;
+import tech.jhipster.lite.statistic.domain.AppliedModule;
+import tech.jhipster.lite.statistic.domain.AppliedModuleFixture;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
+@UnitTest
+class InMemoryStatisticsRepositoryTest {
+
+  private static InMemoryStatisticsRepository inMemoryStatisticsRepository;
+
+  @BeforeAll
+  static void setUp() {
+    inMemoryStatisticsRepository = new InMemoryStatisticsRepository();
+  }
+
+  @BeforeEach
+  void beforeEach() {
+    inMemoryStatisticsRepository.clear();
+  }
+
+  @AfterAll
+  static void cleanup() {
+    inMemoryStatisticsRepository.clear();
+  }
+
+  @Test
+  void testGet_WithNullCriteria() {
+    StatisticsCriteria statisticsCriteria = new StatisticsCriteria(null, null, null);
+
+    assertThat(inMemoryStatisticsRepository.get(statisticsCriteria).appliedModules()).isZero();
+  }
+
+  @Test
+  void testGet_WithModuleSlugCriteria() {
+    AppliedModule appliedModule = AppliedModuleFixture.appliedModule();
+    StatisticsCriteria statisticsCriteria = new StatisticsCriteria(null, null, appliedModule.module().slug());
+
+    inMemoryStatisticsRepository.save(appliedModule);
+
+    assertThat(inMemoryStatisticsRepository.get(statisticsCriteria).appliedModules()).isEqualTo(1);
+  }
+
+  @Test
+  void testGet_WithStartTimeCriteria() {
+    AppliedModule appliedModule = AppliedModuleFixture.appliedModule();
+    StatisticsCriteria statisticsCriteria = new StatisticsCriteria(Instant.parse("2021-12-02T10:15:30.00Z"), null, null);
+
+    inMemoryStatisticsRepository.save(appliedModule);
+
+    assertThat(inMemoryStatisticsRepository.get(statisticsCriteria).appliedModules()).isEqualTo(1);
+  }
+
+  @Test
+  void testGet_WithEndTimeCriteria() {
+    AppliedModule appliedModule = AppliedModuleFixture.appliedModule();
+    StatisticsCriteria statisticsCriteria = new StatisticsCriteria(null, Instant.parse("2021-12-04T10:15:30.00Z"), null);
+
+    inMemoryStatisticsRepository.save(appliedModule);
+
+    assertThat(inMemoryStatisticsRepository.get(statisticsCriteria).appliedModules()).isEqualTo(1);
+  }
+
+  @Test
+  void testGet_WithFalseCriteria() {
+    AppliedModule appliedModule = AppliedModuleFixture.appliedModule();
+    StatisticsCriteria statisticsCriteria = new StatisticsCriteria(
+      Instant.parse("2022-12-04T10:15:30.00Z"),
+      Instant.parse("2020-12-04T10:15:30.00Z"),
+      "slug"
+    );
+
+    inMemoryStatisticsRepository.save(appliedModule);
+
+    assertThat(inMemoryStatisticsRepository.get(statisticsCriteria).appliedModules()).isZero();
+  }
+}
diff --git a/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java b/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java
index 902bb5f3e..a624639db 100644
--- a/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java
+++ b/src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java
@@ -1,7 +1,7 @@
 package tech.jhipster.lite.statistic.infrastructure.secondary;
 
-import static org.assertj.core.api.Assertions.*;
-import static tech.jhipster.lite.statistic.domain.AppliedModuleFixture.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static tech.jhipster.lite.statistic.domain.AppliedModuleFixture.appliedModule;
 
 import java.util.Collections;
 import org.junit.jupiter.api.AfterAll;
@@ -15,6 +15,7 @@ import org.testcontainers.containers.MongoDBContainer;
 import org.testcontainers.utility.DockerImageName;
 import tech.jhipster.lite.IntegrationTest;
 import tech.jhipster.lite.statistic.domain.StatisticsRepository;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
 
 @IntegrationTest
 @EnabledOnOs(OS.LINUX)
@@ -71,7 +72,13 @@ class MongoDBStatisticsRepositoryIntTest {
   @Test
   void shouldUpdateAndGetStatistics() {
     statistics.save(appliedModule());
+    assertThat(statistics.get(null).appliedModules()).isEqualTo(1);
+  }
 
-    assertThat(statistics.get().appliedModules()).isEqualTo(1);
+  @Test
+  void shouldUpdateAndGetStatisticsWithIncorrectCriteria() {
+    statistics.save(appliedModule());
+    StatisticsCriteria criteria = new StatisticsCriteria(null, null, "slug");
+    assertThat(statistics.get(criteria).appliedModules()).isZero();
   }
 }
```

Then, commit and push

kartiktayal:
You want me to reset my branch to the HEAD of upstream/main. Is that correct?
If so, will I have to not commit all my changes again? or am I misunderstanding?

If possible, can you share the commands with me so I do not make this mistake

pascalgrimaud:
My suggestion is to solve these strange co-author commits with github actions bot, as I don't know what you did with your local repo:
image

It's not good to follow clearly what I suggest because I don't know your local configuration:
- your git origin
- your git upstream
- your email config, etc

Still there are several solutions, it depends on you:

To solve that, you can manipulate git with reset, git patch etc

Or, you can simply start a new dev:
- copy the packages statistics (src and test) in another folder
- start in a fresh clear branch, which is sync with upstream/main
- paste the packages statictics (src and test) with your changes

DamnClin:
Really sorry, kinda busy those days, I'll review it in the upcoming days

src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java
```diff
@@ -1,7 +1,14 @@
...
+import org.apache.commons.lang3.ObjectUtils;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
 public interface StatisticsRepository {
   void save(AppliedModule moduleApplied);
 
-  Statistics get();
+  Statistics get(StatisticsCriteria criteria);
+
+  default boolean isAnyCriteriaApplied(StatisticsCriteria criteria) {
```
DamnClin, line 11: This needs to be part of the criteria or be deleted

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
```diff
@@ -0,0 +1,11 @@
...
+
+import java.io.Serializable;
+import java.time.Instant;
+
+/**
+ * Criteria class for {@link tech.jhipster.lite.statistic.domain.Statistics}. This class is used
+ * to receive all the possible filtering options from
+ * the Http GET request parameters.
+ */
+public record StatisticsCriteria(Instant startTime, Instant endTime, String moduleSlug) implements Serializable {}
```
DamnClin, line 11: I don't like the design of this criteria. For me criterion are optional hence should be in Optional attributes which probably means switching to a class with a builder (so this will also fix the ambiguity on instants).

Also you should use types since module slugs have a peculiar format you will avoid unmatching formats

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/StatisticsResources.java
```diff
@@ -21,7 +22,7 @@ public StatisticsResources(StatisticsApplicationService statistics) {
 
   @GetMapping
   @ApiResponse(description = "JHipster lite usage statistics", responseCode = "200")
-  ResponseEntity<RestStatistics> getStatistics() {
-    return ResponseEntity.ok(RestStatistics.from(statistics.get()));
+  ResponseEntity<RestStatistics> getStatistics(StatisticsCriteria criteria) {
```
DamnClin, line 25: You can't use domain criteria here. You need a `RestStatisticsCriteria` so you'll be able to have OpenAPI documentation for the endpoint.

This makes even more sens with criteria having optional criterion (`RestStatisticsCriteria` and `StatisticsCriteria` structure will be different)

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java
```diff
@@ -23,8 +25,22 @@ public void save(AppliedModule moduleApplied) {
   }
 
   @Override
-  public Statistics get() {
-    return new Statistics(appliedModules.size());
+  public Statistics get(StatisticsCriteria criteria) {
+    int appliedModulesCount = 0;
+    if (isAnyCriteriaApplied(criteria)) {
```
DamnClin, line 30: You just need to filter and count in a stream, this needs to be refactored

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
```diff
@@ -24,7 +25,11 @@ public void save(AppliedModule moduleApplied) {
   }
 
   @Override
-  public Statistics get() {
-    return new Statistics(statistics.count());
+  public Statistics get(StatisticsCriteria criteria) {
+    if (isAnyCriteriaApplied(criteria)) {
+      return new Statistics(statistics.countByModuleSlugAndDateBetween(criteria.moduleSlug(), criteria.startTime(), criteria.endTime()));
```
DamnClin, line 30: I don't think that works since all criterion are optional. You should build a query with the criterion that are actually set.

Another problem: you need an index for that search

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
```diff
@@ -24,7 +25,11 @@ public void save(AppliedModule moduleApplied) {
...
+    } else {
```
DamnClin, line 31: Else keyword is not needed

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+import tech.jhipster.lite.statistic.domain.AppliedModuleFixture;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
+@UnitTest
+class InMemoryStatisticsRepositoryTest {
+
+  private static InMemoryStatisticsRepository inMemoryStatisticsRepository;
+
+  @BeforeAll
+  static void setUp() {
```
DamnClin, line 21: Not needed, you can instantiate it a declaration

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+    inMemoryStatisticsRepository = new InMemoryStatisticsRepository();
+  }
+
+  @BeforeEach
+  void beforeEach() {
```
DamnClin, line 26: Just declare a not static attribute and this won't be needed

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+    inMemoryStatisticsRepository.clear();
+  }
+
+  @AfterAll
+  static void cleanup() {
```
DamnClin, line 31: Just declare a not static attribute and this won't be needed

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+    inMemoryStatisticsRepository.clear();
+  }
+
+  @Test
+  void testGet_WithNullCriteria() {
```
DamnClin, line 36: we tend to use shouldXXXWhen|ForXXX tests namming

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java
```diff
@@ -1,7 +1,7 @@
 package tech.jhipster.lite.statistic.infrastructure.secondary;
 
-import static org.assertj.core.api.Assertions.*;
-import static tech.jhipster.lite.statistic.domain.AppliedModuleFixture.*;
+import static org.assertj.core.api.Assertions.assertThat;
```
DamnClin, line 3: static imports should start a 1 import

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepositoryIntTest.java
```diff
@@ -71,7 +72,13 @@ void shouldGetStatisticsFromMongoInCloudEnvironment() {
   @Test
   void shouldUpdateAndGetStatistics() {
     statistics.save(appliedModule());
+    assertThat(statistics.get(null).appliedModules()).isEqualTo(1);
+  }
 
-    assertThat(statistics.get().appliedModules()).isEqualTo(1);
```
DamnClin, line 75: Missing test cases for dates

DamnClin (changes requested):
Some design flows + missing component test in https://github.com/jhipster/jhipster-lite/blob/main/src/test/features/statistics.feature

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
kartiktayal: Just trying to understand the difference between RestStatisticsCriteria and StatisticsCriteria.
RestStatisticsCriteria is going to be taken as Input in Rest endpoint where it would have no validations other than slug since they have types since it can be null with Swagger annotations for documentation.
And this RestStatisticsCriteria wil be used to create StatisticsCriteria which will also have null values even when using builder.
Maybe I am not able to understand the use of optional here?

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
kartiktayal: I did think of building a criteria, but the criteria API are not available due to missing dependency in the project.
I tried changing the interface from CrudRepo to JpaRepo, but it is not available. Should I add the dependency or am I missing something here?

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
kartiktayal: Do we not need to cleanup after the tests since it is a goof practice to purge the test data. Methods annotated with `@AfterAll` a should be void and static which is why the attribute is static.

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
DamnClin: This is not a JPA repo since it is backed by mongo. If you need a dependency you can add it :)

src/test/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepositoryTest.java
DamnClin: It isn't a good practice really to have shared data between tests since here it's really easy to have a new InMemoryRepo for each test: just declare it in an attribute

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
DamnClin: The RestXXX will have OpenApi annotation for documentation and you need a builder for StatisticsCriteria allowing build from primitives

src/main/java/tech/jhipster/lite/generator/JHLiteModuleSlug.java
```diff
@@ -123,4 +125,12 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
   public String get() {
     return slug;
   }
+
+  public static JHLiteModuleSlug fromString(@NotNull String slug) {
```
DamnClin, line 129: No BeanValidation allowed here

src/main/java/tech/jhipster/lite/generator/JHLiteModuleSlug.java
```diff
@@ -123,4 +125,12 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
...
+    return Arrays
```
DamnClin, line 130: I think it's way better to build a static map associating key and values to do this

src/main/java/tech/jhipster/lite/generator/JHLiteModuleSlug.java
```diff
@@ -123,4 +125,12 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
...
+      .stream(JHLiteModuleSlug.values())
+      .filter(val -> slug.equals(val.get()))
+      .findFirst()
+      .orElseThrow(IllegalArgumentException::new);
```
DamnClin, line 134: We don't use generic exceptions, only GeneratorException allowing us to provide user friendly error messages

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+import tech.jhipster.lite.generator.JHLiteModuleSlug;
+
+/**
+ * Criteria class for {@link tech.jhipster.lite.statistic.domain.Statistics}.
+ */
+public class StatisticsCriteria {
+
+  private final Instant startTime;
```
DamnClin, line 13: Don't store nullable values, store optional, it will be obvious that they can be null

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
```diff
@@ -0,0 +1,65 @@
...
+    this.endTime = Optional.ofNullable(builder.endTime).filter(StringUtils::isNotBlank).map(Instant::parse).orElse(null);
+    this.moduleSlug =
+      Optional.ofNullable(builder.moduleSlug).filter(StringUtils::isNotBlank).map(JHLiteModuleSlug::fromString).orElse(null);
+  }
+
+  public static StatisticsCriteriaBuilder builder() {
+    return new StatisticsCriteriaBuilder();
+  }
+
+  public Instant getStartTime() {
```
DamnClin, line 28: Domain accessors can never result in nullable values, all fields must be Optional

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/criteria/RestStatisticsCriteria.java
```diff
@@ -0,0 +1,28 @@
+package tech.jhipster.lite.statistic.infrastructure.primary.criteria;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
+import java.util.Optional;
+import org.apache.commons.lang3.ObjectUtils;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
+@Schema(name = "RestStatisticsCriteria", description = "JHipster lite rest statistics criteria")
+public record RestStatisticsCriteria(
```
DamnClin, line 10: package private

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/criteria/RestStatisticsCriteria.java
```diff
@@ -0,0 +1,28 @@
...
+  @Schema(description = "Start time to apply filter", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String startTime,
```
DamnClin, line 11: fields must be private

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/criteria/RestStatisticsCriteria.java
```diff
@@ -0,0 +1,28 @@
...
+  @Schema(description = "End time to apply filter", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String endTime,
+  @Schema(description = "Module slug to apply filter with", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String moduleSlug
+) {
+  public boolean isAnyCriteriaApplied() {
+    return ObjectUtils.anyNotNull(this.startTime, this.endTime, this.moduleSlug);
+  }
+
+  public StatisticsCriteria toDomain() {
+    return Optional
```
DamnClin, line 20: Never result in null values, this is really missleading, here this is just an empty criteria

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/InMemoryStatisticsRepository.java
```diff
@@ -23,8 +25,22 @@ public void save(AppliedModule moduleApplied) {
   }
 
   @Override
-  public Statistics get() {
-    return new Statistics(appliedModules.size());
+  public Statistics get(StatisticsCriteria criteria) {
+    long appliedModulesCount = 0;
+    if (Objects.nonNull(criteria)) {
```
DamnClin, line 30: THis will need a full rework with null safe domain

src/main/java/tech/jhipster/lite/statistic/infrastructure/secondary/MongoDBStatisticsRepository.java
```diff
@@ -24,7 +34,27 @@ public void save(AppliedModule moduleApplied) {
...
+      Query query = generateQuery(criteria);
+      return new Statistics(mongoTemplate.count(query, AppliedModuleDocument.class));
+    }
     return new Statistics(statistics.count());
   }
+
+  private Query generateQuery(@NotNull StatisticsCriteria criteria) {
+    Query query = new Query();
+    List<Criteria> criteriaList = new ArrayList<>();
+    if (Objects.nonNull(criteria.getStartTime())) {
```
DamnClin, line 48: This will need a full rework with null safe domain

src/test/features/statistics.feature
```diff
@@ -7,11 +7,23 @@ Feature: application statistics
...
     When I get statistics
     Then I should have statistics
       | Applied modules | 3 |
+
+  Scenario: Should have 1 applied module when specific criteria is applied
+    Given I apply modules to default project
+      | maven-java  |
+      | spring-boot |
+      | mariadb     |
+    When I get statistics with criteria
```
DamnClin, line 24: time filter should be tested in component tests

src/test/java/tech/jhipster/lite/statistic/domain/AppliedModuleFixture.java
```diff
@@ -18,6 +19,16 @@ public static AppliedModule appliedModule() {
       .properties(moduleProperties());
   }
 
+  public static AppliedModule appliedModule(JHLiteModuleSlug slug) {
```
DamnClin, line 22: The previous method should be refactored to rely on this one

DamnClin (changes requested):
Domain model should never rely on nullable values, this change will have impact on the whole PR

kartiktayal:
Thank you for such prompt response on the PR. Will take care of all the review comments.

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/criteria/RestStatisticsCriteria.java
kartiktayal: When I am changing the visibility to package-private, I cannot access it in `StatisticsResources`, do you want me to move it outside on the same level as StatisticsResources?

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/criteria/RestStatisticsCriteria.java
kartiktayal: Correct me if I am wrong but the data in records are held in private final fields, so my question here is do you want me to change it to a normal class or is there something I am missing here?

src/main/java/tech/jhipster/lite/statistic/domain/criteria/StatisticsCriteria.java
kartiktayal: Correct me if my understanding deviates from the expectation, but from what I can understand, having all fields as Optional is not a good idea because Optional is supposed to be used as a return type. IntelliJ is also showing a warning in place as
`'Optional<Instant>' used as type for field 'startTime' `
for when I make the fields optional, say
`private final Optional<Instant> startTime;`
Please do correct me if this was not your intention.

src/main/java/tech/jhipster/lite/generator/JHLiteModuleSlug.java
```diff
@@ -123,4 +130,8 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
   public String get() {
     return slug;
   }
+
+  public static JHLiteModuleSlug fromString(String slug) {
```
DamnClin, line 134: You need to handle the case of unknown slug, either by resulting in an optional or by throwing an exception. No null value should ever be returned by the domain

src/main/java/tech/jhipster/lite/generator/JHLiteModuleSlug.java
```diff
@@ -113,6 +117,9 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
   VUE_CORE("vue-core"),
   VUE_PINIA("vue-pinia");
 
+  private static final Map<String, JHLiteModuleSlug> moduleSlugMap = Arrays
```
DamnClin, line 120: `Stream.of` instead of `Arrays.stream`

src/main/java/tech/jhipster/lite/statistic/application/StatisticsApplicationService.java
```diff
@@ -18,7 +20,7 @@ public void moduleApplied(AppliedModule moduleApplied) {
     statistics.save(moduleApplied);
   }
 
-  public Statistics get() {
-    return statistics.get();
+  public Statistics get(Optional<StatisticsCriteria> criteria) {
```
DamnClin, line 23: Never ask for an Optional as parameter. StatisticsCriteria can have all Optional fields but still beeing mandatory

src/main/java/tech/jhipster/lite/statistic/domain/StatisticsRepository.java
```diff
@@ -1,7 +1,10 @@
 package tech.jhipster.lite.statistic.domain;
 
+import java.util.Optional;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
 public interface StatisticsRepository {
   void save(AppliedModule moduleApplied);
 
-  Statistics get();
+  Statistics get(Optional<StatisticsCriteria> criteria);
```
DamnClin, line 9: Never ask for an Optional as parameter. StatisticsCriteria can have all Optional fields but still beeing mandatory

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/RestStatisticsCriteria.java
```diff
@@ -0,0 +1,25 @@
+package tech.jhipster.lite.statistic.infrastructure.primary;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
+import java.util.Optional;
+import org.apache.commons.lang3.ObjectUtils;
+import tech.jhipster.lite.statistic.domain.criteria.StatisticsCriteria;
+
+@Schema(name = "RestStatisticsCriteria", description = "JHipster lite rest statistics criteria")
+record RestStatisticsCriteria(
```
DamnClin, line 10: Times should be instant, this way we'll get format in documentation and validation during déserialization

src/main/java/tech/jhipster/lite/statistic/infrastructure/primary/RestStatisticsCriteria.java
```diff
@@ -0,0 +1,25 @@
...
+  @Schema(description = "Start time to apply filter", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String startTime,
+  @Schema(description = "End time to apply filter", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String endTime,
+  @Schema(description = "Module slug to apply filter with", nullable = true, requiredMode = RequiredMode.NOT_REQUIRED) String moduleSlug
+) {
+  public boolean isAnyCriteriaApplied() {
+    return ObjectUtils.anyNotNull(this.startTime, this.endTime, this.moduleSlug);
+  }
+
+  public Optional<StatisticsCriteria> toDomain() {
+    if (isAnyCriteriaApplied()) {
```
DamnClin, line 20: Checking this is not the responsibility of the primary adapter. You can just create a Criteria, it will be enought

kartiktayal:
@DamnClin , gentle reminder for review. I believe all review comments have been addressed.

DamnClin:
> @DamnClin , gentle reminder for review. I believe all review comments have been addressed.

Sorry, totally missed the message! Doing it right now

src/test/features/statistics.feature
```diff
@@ -7,11 +7,47 @@ Feature: application statistics
...
+      | startTime  |             |
+      | endTime    |             |
+      | moduleSlug | spring-boot |
+    Then I should have statistics
+      | Applied modules | 1 |
+
+  Scenario: Should have 3 applied module when specific criteria is applied
```
DamnClin, line 31: Can be a good idea to actually filter some to check the updated stat

src/test/java/tech/jhipster/lite/statistic/infrastructure/primary/StatsSteps.java
```diff
@@ -1,12 +1,15 @@
 package tech.jhipster.lite.statistic.infrastructure.primary;
 
-import static tech.jhipster.lite.cucumber.CucumberAssertions.*;
+import static tech.jhipster.lite.cucumber.CucumberAssertions.assertThatLastResponse;
```
DamnClin, line 3: static imports should be * (it's an IDE configuration)

DamnClin (approved):

src/test/features/statistics.feature
kartiktayal: @DamnClin , can you please elaborate more on the request here. I have added the criteria for all the parameters in the feature file. Anything more to be added on this?

DamnClin: The search result in the same result so if the criteria does nothing the test still pass

kartiktayal: Do you want to have a search with incorrect date or incorrect module? Can you give me an example to understand the context a bit better.

DamnClin: A test is meant to validate something. Here, if you remove your feature and run the test it stills pass since it's expecting the same result. You need to see the test failing without the feature

kartiktayal: I have added the case where the API would return 0 in case there are no modules applied. I am still a little confused on failing the test case since the API would return 0 in case there are no applied modules. Please let me know if this was not the intention.

DamnClin: Ok, just keep Scenario: Should have 1 applied module when specific criteria is applied, the other new ones are doing nothing (keep the older ones)

src/test/java/tech/jhipster/lite/statistic/infrastructure/primary/StatsSteps.java
kartiktayal: Done.

src/test/features/statistics.feature
kartiktayal: Done.

kartiktayal:
@DamnClin , all review comments have been addressed.

Issue #6651, opened 2023-06-23
Improve CI to not launch same tests


pascalgrimaud:
For each builds, we should compare:
- the hash of the generated project from main branch
- the hash of the generated project from pull request
- if it's different, launch the build
- otherwise, it should be success

With this change, we won't launch same builds for each pull requests

pascalgrimaud:
@renanfranca : I saw a small issue. Every code in main branch should launch all tests. This new logic should apply only for pull request. What do you think?
Otherwise, there are no test anymore in main branch

renanfranca:
> @renanfranca : I saw a small issue. Every code in main branch should launch all tests. This new logic should apply only for pull request. What do you think? Otherwise, there are no test anymore in main branch

I am going to skip the new steps, even the job generate-main-branch-jar if it is not in a pull request. I am going to create a new pull request for that. Thanks 👍

PR #6718, opened 2023-06-29
ci: improved ci to not launch some tests.

zhyd1997:
Close: https://github.com/jhipster/jhipster-lite/issues/6651

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -181,15 +181,22 @@ jobs:
           ls -al /tmp/jhlite/${{ matrix.app }}
           MD5SUM_POM_XML=$(md5sum /tmp/jhlite/${{ matrix.app }}/pom.xml | cut -d ' ' -f 1)
           echo $MD5SUM_POM_XML
-          if [ "${{ env.md5sum_pom_xml }}" == "${{ steps.cache_md5sum.outputs.md5sum_pom_xml }}" ]; then
+          cd /tmp/jhlite/
+          tar -cvf ${{ matrix.app }}_archive.tar --exclude="node_modules" --exclude=".gitattributes" --exclude=".git" ${{ matrix.app }}
+          MD5SUM_${{ matrix.app }}=$(md5sum ${{ matrix.app }}_archive.tar | cut -d ' ' -f 1)
+          echo $MD5SUM_${{ matrix.app }}
+          if [ "${{ env['md5sum_${{ matrix.app }}'] }}" == "${{ steps.cache_md5sum.outputs['MD5SUM_${{ matrix.app }}'] }}" ]; then
             echo "Cache is up to date. Stopping workflow..."
             echo "cache_up_to_date=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Thanks @zhyd1997 !
I'll have a look soon :-)

.github/workflows/github-actions.yml
```diff
@@ -181,15 +181,25 @@ jobs:
           ls -al /tmp/jhlite/${{ matrix.app }}
           MD5SUM_POM_XML=$(md5sum /tmp/jhlite/${{ matrix.app }}/pom.xml | cut -d ' ' -f 1)
           echo $MD5SUM_POM_XML
+          if [ "${{ env.md5sum_pom_xml }}" == "${{ steps.cache_md5sum.outputs.md5sum_pom_xml }}" ]; then
```
pascalgrimaud, line 184: I don't think it's enough
We should compare the md5 of the generated project, including all existing files
Because if there is a change in package.json, with this solution, it won't be detect so no build

Some ideas which need to be discussed:
- generate a tar or zip, then calculate the md5
- use something like md5deep or checksumdir

pascalgrimaud (changes requested):

.github/workflows/github-actions.yml
zhyd1997: > We should compare the md5 of the generated project, including all existing files

Expensive Operation

pascalgrimaud: locally, it's very fast, so not an expensive operation, just need to clear some files before
that's why it's not an easy ticket :)

.github/workflows/github-actions.yml
```diff
@@ -181,15 +181,32 @@ jobs:
           ls -al /tmp/jhlite/${{ matrix.app }}
           MD5SUM_POM_XML=$(md5sum /tmp/jhlite/${{ matrix.app }}/pom.xml | cut -d ' ' -f 1)
           echo $MD5SUM_POM_XML
+          cd /tmp/jhlite/
+          tar -cvf ${{ matrix.app }}_archive.tar --exclude="node_modules" --exclude=".gitattributes" --exclude=".git" ${{ matrix.app }}
+          MD5SUM_${{ matrix.app }}=$(md5sum ${{ matrix.app }}_archive.tar | cut -d ' ' -f 1)
+          echo $MD5SUM_${{ matrix.app }}
+          if [ "${{ env['md5sum_${{ matrix.app }}'] }}" == "${{ steps.cache_md5sum.outputs['MD5SUM_${{ matrix.app }}'] }}" ]; then
```
zhyd1997, line 188: this line of code need to update:

Currently it equals to

`if [ "" == "" ]; then`

zhyd1997:
Hi @pascalgrimaud 

Sorry 🙏  for long time no reply, seems that it's hard for me to access env variable in github actions.

Can someone continue to work on this task based on this commit and your idea? 👀

pascalgrimaud:
Thanks @zhyd1997 : fine, I'll improve your work as soon as I can

PR #6804, opened 2023-07-07
Rename @Generated to @ExcludeFromGeneratedCodeCoverage

murdos:
See https://github.com/jhipster/jhipster-lite/pull/4907#issuecomment-1371967487


DamnClin:
BTW https://github.com/jacoco/jacoco/pull/1472 has been merged and will be release in the next jacoco release. Is it a good to mark those cases so it will be easy to remove them?

Issue #6879, opened 2023-07-14
Dedicated spring profiles


pascalgrimaud:
I'm starting to add some spring profiles:
- angular
- react
- vue
- svelte

The goal is to hide the frameworks you don't use

PR #6880, opened 2023-07-14
Add dedicated spring profiles

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/6879

### DIFF
--- a/src/main/resources/config/application-angular.properties
+++ b/src/main/resources/config/application-angular.properties
@@ -1,0 +1,1 @@
+jhlite-hidden-resources.slugs=gradle-java,spring-boot-kafka-dummy-producer-consumer,react-core,svelte-core,vue-core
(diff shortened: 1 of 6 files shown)


src/main/resources/config/application-angular.properties
```diff
@@ -0,0 +1 @@
+jhlite-hidden-resources.slugs=gradle-java,spring-boot-kafka-dummy-producer-consumer,react-core,svelte-core,vue-core
```
DamnClin, line 1: We should add a common hidden slugs property in application.properties and use it here to avoid repetition

pascalgrimaud: yes it's a good idea, I think it can be done in another PR
I'm merging this one to not have conflict when disabling Svelte, as suggested in other ticket

PR #6941, opened 2023-07-20
Improve logback configuration in generated apps

pascalgrimaud:


DanielFran:
@pascalgrimaud in the generator-jhipster we are now using only the console-appender by default
https://github.com/jhipster/jhipster-sample-app/blob/main/src/main/resources/logback-spring.xml#L14

pascalgrimaud:
thanks @DanielFran 
I'll have a look to have the most default config as possible

DanielFran:
@pascalgrimaud 

Following https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto.logging.logback, legacy base.xml file is provided for compatibility with earlier versions of Spring Boot 1.
base.xml includes actually default + console appended + file appender

In generator jhipster we review the configuration of logback to define only console appender. For that it uses only default+console-appender XML
Example from spring-boot documentation:
````xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
    <include resource="org/springframework/boot/logging/logback/console-appender.xml" />
    <root level="INFO">
        <appender-ref ref="CONSOLE" />
    </root>
    <logger name="org.springframework.web" level="DEBUG"/>
</configuration>
````

So, I believe you need to review the config, adding console-appender and keeping base.xml is not good.

pascalgrimaud:
Can you provide a PR @DanielFran plz ?
It will help you to understand our templates in this projet :-)

PR #6958, opened 2023-07-21
Logback enhancements

DanielFran:
Note:
It is not needed to define the shutdown hook explicitly since version 1.1.10 of logback. Version 1.1.10 onwards, logback takes care of stopping the current logback-classic context (and all the appenders) when the web-app is stopped or reloaded. Here's the updated doc: https://logback.qos.ch/manual/configuration.html#webShutdownHook

Related with https://github.com/jhipster/jhipster-lite/pull/6941


DanielFran:
@pascalgrimaud not sure we should keep also the file and async appenders examples

pascalgrimaud:
Thanks @DanielFran

PR #6976, opened 2023-07-24
Organize features in thematic folders to better apprend them

murdos:
We have a lot a features (files), having folders might help working with them

WDYT?


pascalgrimaud:
I'm fine with this change

DamnClin (approved):

PR #7036, opened 2023-07-30
Include actual value for date compare assertions

### LINKED ISSUE #7031: Display source time in time comparaison exceptions
In [NotBeforeTimeException](https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/javatool/base/main/NotBeforeTimeException.java.mustache) and [NotAfterTimeException](https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/javatool/base/main/NotAfterTimeException.java.mustache) we are only displaying one of the two compared times which can result in harder analysis when running in production.

We should display both times (that mean reworking the building logic of those exceptions and updating [Assert](https://github.com/jhipster/jhipster-lite/blob/main/src/main/resources/generator/server/javatool/base/main/Assert.java.mustache))

adem-hassine:
Close #7031


src/main/java/tech/jhipster/lite/error/domain/NotAfterTimeException.java
```diff
@@ -8,19 +8,25 @@ private NotAfterTimeException(String field, String message) {
     super(field, message);
   }
 
-  public static NotAfterTimeException notStrictlyAfter(String fieldName, Instant other) {
-    return new NotAfterTimeException(fieldName, message(fieldName, "must be strictly after", other));
+  public static NotAfterTimeException notStrictlyAfter(String fieldName, Instant actual, Instant other) {
```
DamnClin, line 11: I don't like having 2 parameters of the same in a public method. Those should be refactored to something without ambiguity (I'd go for a fluent builder as all parameters here are mandatory)

DamnClin (changes requested):

src/main/java/tech/jhipster/lite/error/domain/NotAfterTimeException.java
adem-hassine: So, do you want to introduce the builder pattern for all the fields or only the new one?

DamnClin: Must be for all fields but it's not just a builder, it's one where you can't miss any field at compile time

adem-hassine: Please check again and let me know if this is good. If it is, I can add some blank or empty assertions in the final build function.

src/main/java/tech/jhipster/lite/error/domain/NotAfterTimeException.java
```diff
@@ -8,24 +8,40 @@ private NotAfterTimeException(String field, String message) {
...
-    return new NotAfterTimeException(fieldName, message(fieldName, "must be strictly after", other));
+  public static NotAfterTimeExceptionBuilder field(String fieldName, Instant value) {
+    return new NotAfterTimeExceptionBuilder(fieldName, value);
   }
 
-  public static NotAfterTimeException notAfter(String fieldName, Instant other) {
-    return new NotAfterTimeException(fieldName, message(fieldName, "must be after", other));
-  }
+  record NotAfterTimeExceptionBuilder(String fieldName, Instant value) {
+    public NotAfterTimeException notStrictlyAfter(Instant other) {
```
DamnClin, line 16: strictlyNotAfter sounds better

src/main/java/tech/jhipster/lite/error/domain/NotBeforeTimeException.java
```diff
@@ -8,24 +8,40 @@ private NotBeforeTimeException(String field, String message) {
...
-    return new NotBeforeTimeException(fieldName, message(fieldName, "must be strictly before", other));
+  public static NotBeforeTimeException.NotBeforeTimeExceptionBuilder field(String fieldName, Instant value) {
+    return new NotBeforeTimeException.NotBeforeTimeExceptionBuilder(fieldName, value);
   }
 
-  public static NotBeforeTimeException notBefore(String fieldName, Instant other) {
-    return new NotBeforeTimeException(fieldName, message(fieldName, "must be before", other));
-  }
+  public record NotBeforeTimeExceptionBuilder(String fieldName, Instant value) {
+    public NotBeforeTimeException notStrictlyBefore(Instant other) {
```
DamnClin, line 16: strictlyNotBefore sounds better

DamnClin (commented):
I like it but formatting is off, you should build an app with java-base and prettier, then run prettier (`npm i` and `npm run prettier:format`) to get a formatted version of the templates

DamnClin:
You should have either Fix #issue or Close #issue in your PR comment so the issue will close when merging it

src/main/java/tech/jhipster/lite/error/domain/NotAfterTimeException.java
```diff
@@ -8,24 +8,40 @@ private NotAfterTimeException(String field, String message) {
...
-      .append(" ")
-      .append(other)
-      .append(" but wasn't")
-      .toString();
+    private static String message(String fieldName, Instant actual, String hint, Instant other) {
+      return new StringBuilder()
+        .append("Time in \"")
+        .append(fieldName)
+        .append("\" ")
+        .append("having value")
```
DamnClin, line 33: can update to "having value : " and remove the next lines :)

src/main/java/tech/jhipster/lite/error/domain/NotAfterTimeException.java
DamnClin: I think they must be in the record constructor

DamnClin (approved):

PR #7103, opened 2023-08-04
Generator: Executes integration tests with a configuration derived from application configuration

murdos:
This makes tests more reliable since they use a configuration that is more real (since it's derived from main configuration).
Indeed it's currently possible to have passing tests because of wrong/outdated test configuration, whereas main application fails.

Also this decreases maintenance burden of test configuration (application-test.properties) since it only contains overridden parameters configuration.


DamnClin:
Why keeping the application.properties test file with all properties in place?

murdos:
> Why keeping the application.properties test file with all properties in place?

The application.properties in fact has been renamed to application-test.properties, so it doesn't exist anymore.

PR #7104, opened 2023-08-04
generated-application: Executes integration tests with a configuration derived from application configuration

murdos:
Follow-up to https://github.com/jhipster/jhipster-lite/pull/7103 - for generated applications

This make tests more reliable since they use a configuration that is more real. And this decrease maintenance burden of test configuration since it only contains configuration override through Spring test profile.


DamnClin (approved):
Sorry, I created conflict with main (I got confuse between your PRs...)

PR #7226, opened 2023-08-20
Remove unused imports

qmonmert:

PR #7282, opened 2023-08-24
Use checkstyle to check for unused imports during build

murdos:
Following https://github.com/jhipster/jhipster-lite/pull/7226#issuecomment-1685393587

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" ?>
+<!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" "https://checkstyle.org/dtds/configuration_1_3.dtd">
+<module name="Checker">
+  <module name="TreeWalker">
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
@murdos : what do you think about adding this as a new module too ?

murdos:
> @murdos : what do you think about adding this as a new module too ?

It was indeed my intention. Do you think I should do it in the same PR?

pascalgrimaud:
@murdos : no it's better to do in a separate PR, as it's a really simple module and it will be another good example for new contributors :-)

pascalgrimaud:
@murdos : for your information, just tested it and it works well, excepting for class in src/test, so I think it would be nice to include them too, what do you think?

murdos:
> @murdos : for your information, just tested it and it works well, excepting for class in src/test, so I think it would be nice to include them too, what do you think?

Sure! I thought that this was the default behavior of the maven plugin, but it is not :/

PR #7335, opened 2023-08-28
Run Cucumber tests with JUnit5

### LINKED ISSUE #7303: Run Cucumber tests with JUnit5
We're still running Cucumber tests with JUnit4 / `junit-vintage-engine`.
Since Cucumber supports JUnit5 with [cucumber-junit-platform-engine](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-junit-platform-engine/README.md), we should switch to it and remove dependency to `junit-vintage-engine`

murdos:
Fixes #7303


pascalgrimaud:
@murdos : do you think this PR is ready?

murdos:
Yes, it is. And it's tested with CI 😀

PR #7370, opened 2023-09-02
Double check security configuration

### LINKED ISSUE #5335: Double check security configuration
Following spring security changes we should double check our configuration following https://github.com/jhipster/generator-jhipster/pull/20107

qmonmert:
Fix #5335 

@DamnClin I don't know if it was what you expected?


src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/oauth2/core/domain/OAuth2ModuleFactory.java
```diff
@@ -86,6 +86,7 @@ private void appendJavaFiles(JHipsterModuleBuilder builder, JHipsterModuleProper
         .addTemplate("JwtGrantedAuthorityConverter.java")
         .addTemplate("OAuth2Configuration.java")
         .addTemplate("SecurityConfiguration.java")
+        .addTemplate("SpaWebFilter.java")
```
pascalgrimaud, line 89: Why do you think it is needed in this module?

qmonmert: you're right it's a mistake

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/jwt/domain/JwtAuthenticationModuleFactory.java
```diff
@@ -61,11 +61,13 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
           .addTemplate("JWTFilter.java")
           .addTemplate("JwtReader.java")
           .addTemplate("SecurityConfiguration.java")
+          .addTemplate("SpaWebFilter.java")
```
pascalgrimaud, line 64: what I mean, is: this class is not needed if you only work with backend application, that's why it should be done in another module, instead of by default jwt module

qmonmert: in a new module?

pascalgrimaud: Yes, unless I'm wrong about the use of this class

DamnClin: I also think this should be in a dedicated module

pascalgrimaud:
opened for too long, let's close this and discuss in ticket, as I still this the SpaWebFilter should be done in dedicated module instead of default

PR #7518, opened 2023-09-18
feat(SpyLogs): Split LogsSpy into LogSpy and LogSpyExtension to follow single-responsibility principle

murdos:
This allows not exposing JUnit5 related methods (beforeEach, afterEach, supportsParameter, ...) in actual tests. 
LogSpy could now be injected directly by annotating the field with `@Logs`: this should improve easy of use with other extensions.

I'll also change generated code if we agree on the principle.

### DIFF
--- a/src/test/java/tech/jhipster/lite/Logs.java
+++ b/src/test/java/tech/jhipster/lite/Logs.java
@@ -1,0 +1,13 @@
+package tech.jhipster.lite;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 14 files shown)


DamnClin:
With both annotation and constructor injection ok for me

DamnClin (approved):

murdos:
> I'll also change generated code if we agree on the principle.

I was also wondering if this feature could not deserve to be in its own module, rather than being hidden/embedded in the spring-core module. WDYT?

pascalgrimaud:
I agree too, it deserved to be in its own module so it can be used in existing spring projects :)

PR #7531, opened 2023-09-19
New LogsSpy module

murdos:
extracted from spring-boot-core, with improved features

PR #7569, opened 2023-09-25
Fix missing dependency to logs-spy module

pascalgrimaud:
following https://github.com/jhipster/jhipster-lite/pull/7531


murdos (approved):
Good catch, thank you!

PR #7629, opened 2023-10-02
Refactoring vue-core

### LINKED ISSUE #7587: Refactoring: vue-core module
I think we need to rework on the **vue-core module** and to split between smaller modules.

I think the **vue-core** module should add:
- Vue.js with Vite
- axios
- vue-router (can be discussed)
- all important tools like eslint, typescript, vitest + config, etc

It should not add the home page

Then, the home page should be splitted into 2 modules:
- **vue-home-domain**, which depends on **vue-core**
- for the primary part, which depends on **vue-home-domain**:
  - **vue-home-primary-native**, with default App.html, AppVue.vue etc
  - vue-home-primary-bootstrap
  - vue-home-primary-tailwind (for example)
  - etc

Pinia should be in another dedicated module.

Don't hesitate to comment this, if you have better idea.

abdelfetah18:
Fix #7587 

- Vue.js with Vite
- axios
- all important tools like eslint, typescript, vitest + config, etc

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
@@ -14,16 +14,9 @@ public class VueModulesFactory {
 
   private static final JHipsterSource SOURCE = from("client/vue");
   private static final JHipsterSource DOCUMENTATION_SOURCE = SOURCE.append("documentation");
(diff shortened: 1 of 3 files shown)


DamnClin:
I think we can delete that with the current vue-core module state.

@murdos @pascalgrimaud @qmonmert you agree?

murdos:
@DamnClin : Indeed, this can be closed.

Issue #1956, opened 2022-06-05
Module: Handle spring yaml


DamnClin:
Following https://github.com/jhipster/jhipster-lite/pull/1953 we'll need to add yml spring properties management

swarajsaaj:
Do you want to use yaml for all generated properties files (`application.yml` and `bootstrap.yml`) instead of `application.properties` ? or use both somehow giving an option?

DamnClin:
Depends :⁠-⁠). Give option for init but stick to existing if the file is here. I also want to keep formatting.

In fact, I really want to do it and, since I'm starting to finally feel better (less sick) I'll probably be able to take that soon :⁠-⁠)

swarajsaaj:
Great :) in that case wish you get well soon and take it up when you feel like👍

murdos:
FYI I intend to work on this feature, since I think it would be a nice feature to have :)

murdos:
@pascalgrimaud : I propose to consider this ticket as done as an initial limited state with the work done with the 4 PR mentioned above, and released in version 0.47. 
And follow continued effort with #7939.
WDYT?

pascalgrimaud:
Agree @murdos

murdos:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/174931

PR #7734, opened 2023-10-12
Preparation for Spring YAML configuration

murdos:
In order to avoid a too big PR, here are first steps in supporting YAML configuration file with Spring (see #1956):
- a YamlFileSpringPropertiesHandler using snake-yaml for serialzing YAML
- Change the type of value in PropertyValue from String to Object, with factory methods that only allow creation of PropertyValue with String, Boolean or Number


DamnClin (approved):
LGTM

PR #7747, opened 2023-10-15
(refactoring) Remove application(-.*).properties templates

### LINKED ISSUE #7745: Refactoring: Remove application(-*).properties templates
Prerequisite for #1956 : these files (`application.properties`, `application-local.properties`, `application-test.properties`) should be initialized by declaring the spring properties in `SpringBootCoreModuleFactory`, and not anymore by copying templates.

This will allow using another configuration format like Yaml transparently.

murdos:
These files (application.properties, application-local.properties, application-test.properties) should be initialized by declaring the spring properties in SpringBootCoreModuleFactory, and not anymore by copying templates. This will allow using another configuration format like Yaml transparently. 
Fixes #7745


src/main/resources/generator/server/springboot/core/main/application-local.properties.mustache
```diff
@@ -1,7 +0,0 @@
-#====================================================================
```
murdos, line 1: The only downside is that we're losing this header.
I'm not sure if it's worth trying to recreate it or not.
I've played a bit with adding a block comment on the first property, but we have to cheat to select the first one on alphabetic order.

Issue #7568, opened 2023-09-25
Migration to Java 21


pascalgrimaud:
Different points:
- migrate JHLite engine to Java 21
- migrate generated apps to Java 21

Need to check:
- eclipse-temurin Docker image
- the CI
- etc

murdos:
Temurin 21 release is delayed because of some legal changes around TCK test licensing: https://adoptium.net/fr/blog/2023/09/temurin21-delay/
Meanwhile we can switch to zulu that has already released its 21 version.

pascalgrimaud:
About JHLite Engine:
- should we migrate to Java 21 ? I'd vote yes

About generated application:
- should we migrate to Java 21 ? (I'd vote yes)
- if yes, should we keep compatibility with Java 17 ? (I'd vote no)

Maybe it's time to release official 1.0.0 if we decide to push Java 21 as main version

What do you think @murdos @DamnClin @qmonmert ?

qmonmert:
@pascalgrimaud same

About JHLite Engine:
- should we migrate to Java 21 ? ✅

About generated application:
- should we migrate to Java 21 ? ✅
- if yes, should we keep compatibility with Java 17 ? ❌

murdos:
> About JHLite Engine:
> - should we migrate to Java 21 ?
I vote yes too. 


> About generated application:
> - should we migrate to Java 21 ?
> - if yes, should we keep compatibility with Java 17 ?

I vote for switching to java 21 by default, but keep compatibility with java 17 a least for a few months.
There are companies that might want to use jhipster-lite, that can't yet use java 21 because it's not yet widely available.

murdos:
I'm going to switch generated applications on Java 21 in a incoming PR, unless someone disagree :-)

pascalgrimaud:
I'm in favor of migrating to Java 21 for generated app, as it's the new LTS

murdos:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/179494

pascalgrimaud:
@murdos : approved

PR #7944, opened 2023-11-05
Use Java 21 as minimum runtime for jhipster-lite engine

murdos:
See #7568


anderskristian:
Good idea for jLite assistant 🐸
Will generated java code also require Java 21 ?

murdos:
> Good idea for jLite assistant 🐸 Will generated java code also require Java 21 ?

There's a discussion about that starting from that comment: https://github.com/jhipster/jhipster-lite/issues/7568#issuecomment-1786828602

pascalgrimaud:
cc @jdubois as it will probably impact your deployment

PR #7964, opened 2023-11-06
Fix hexa arch test

cvirieux:
This fixes the shouldNotDependOnOutside test in HexagonalArchTest.
Before this fix, no classes were included in the filter ".domain..".

This test doesn't validate that a domain class doesn't depend on another domain (but shouldNotDependOnOtherBoundedContextDomains checks it). 

I had to move slugs in a new shared kernel package to make it work.

I've made 2 commits to simplify the code review.


murdos:
@cvirieux : thanks for catching this error and fixing it!
I noticed on my project at work that such violations were not detected but I did not took the time to investigate why. And now it's fixed thanks to you :-D

PR #8013, opened 2023-11-09
Keycloak: remove duplicate entries in jhipster-realm.json.mustache

### LINKED ISSUE #7997: Keycloak: improve jhipster-realm.json.mustache
The key "users" is duplicated in the config

image

qmonmert:
Fix #7997

### DIFF
--- a/src/main/resources/generator/server/springboot/mvc/security/oauth2/core/docker/jhipster-realm.json.mustache
+++ b/src/main/resources/generator/server/springboot/mvc/security/oauth2/core/docker/jhipster-realm.json.mustache
@@ -455,25 +455,6 @@
   "webAuthnPolicyPasswordlessCreateTimeout": 0,
   "webAuthnPolicyPasswordlessAvoidSameAuthenticatorRegister": false,
   "webAuthnPolicyPasswordlessAcceptableAaguids": [],
-  "users": [
-    {
-      "id": "f742ba6f-1d8a-4dec-bf15-e02dab508283",
-      "createdTimestamp": 1598681172054,
-      "username": "service-account-internal",
-      "enabled": true,
-      "totp": false,
-      "emailVerified": false,
-      "serviceAccountClientId": "internal",
-      "disableableCredentialTypes": [],
-      "requiredActions": [],
-      "realmRoles": ["offline_access", "uma_authorization"],
-      "clientRoles": {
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
My first idea was to merge all users, to not duplicate the 'users' entry.

Then, I don't know if this user is a default one, provided by Keycloak, or if it was added by JHipster.
If it's provided: we should keep
If not, we should delete as we should not add unused users

qmonmert:
@pascalgrimaud you're right, this user is necessary and it's normal to have two entries in the JSON
On the generator we have also two entries:
https://github.com/jhipster/generator-jhipster/blob/561a0ef2fd8da1c6ba2c653b29e81fd3b4004656/generators/docker/templates/docker/realm-config/jhipster-realm.json.ejs#L502
https://github.com/jhipster/generator-jhipster/blob/561a0ef2fd8da1c6ba2c653b29e81fd3b4004656/generators/docker/templates/docker/realm-config/jhipster-realm.json.ejs#L2587
I think that we can close this PR and the issue

pascalgrimaud:
Because the file from generator-jhipster is not a good one, and it's my fault because I copy from there.
That's why I created the ticket to improve it and to remove old and deprecated stuff

qmonmert:
@pascalgrimaud ok
I tried to start Keycloak without this user and all seems good
I think that is is not provided by Keycloak but I'm not a specialist of Keycloak :)

pascalgrimaud:
That's why I think this file must be reworked completely, using this workflow:
- start a fresh Keycloak without any custom
- add manually what it is needed by a generated JHLite project
- then export the settings (realm, user, etc)

The scope of this PR should be to move the custom users (admin + user), near to `service-account-internal`
The `service-account-internal` should not be removed, as I think it's an internal user of Keycloak

qmonmert:
@pascalgrimaud this is what I tried yesterday, I created a new realm but I struggle to retrieve users ...

pascalgrimaud:
Thanks and no worry, I created this ticket and I know exactly what to do to have a clean json file to import when starting Keycloak

As for my customer, we just upgraded to Keycloak 22 :-)

qmonmert:
ok 🙂

pascalgrimaud:
closing in favor of https://github.com/jhipster/jhipster-lite/pull/8148

PR #8017, opened 2023-11-12
Suggest merge feature

anderskristian:
I suggest that jLite adds a merge feature.
This means that jLite can preserve users custom changes when generating updates to sources.

I have tried to document the  in 

- [documentation/merge/MERGE.md](https://github.com/anderskristian/jhipster-lite-merge/blob/feature/genlogMerge/documentation/merge/MERGE.md) 
- [documentation/merge/GENLOG.md](https://github.com/anderskristian/jhipster-lite-merge/blob/feature/genlogMerge/documentation/merge/GENLOG.md)

- I suggest that GENLOG will become a feature (module) that uses can select.
- I have never made a module. And this module will be a little special.
- The PR code might be a little simple. The merge algorithm can be improved. Right now it searches for matching fragments in a possible too limited way.
- I suggest a  unix like patch file generated file. That the user can merge. This does not yet exist. 
- I would like the merge tool to autodetect EOL character. (This is not implemented)
- And probably some more things.
- files are probably located in a place that you would like another place.

I want some feedback on this feature. If it is wanted I will improve it. Other wise there is not so much reason to do so.


pascalgrimaud:
First of all, thanks for your work and proposing this.
But I think it should be discussed first before you started all this work.

From my personal use of JHLite since more than 1 year:
- I rarely apply again an existing module, because it didn't change a lot
- the only thing I do in my projects is to upgrade the libs
- that's why I think it's overkill to manage this kind of merge conflict
- it's more easier to manually resolve the possible conflict 
- I don't want the same feature 'upgrade' than in generator-jhipster

Another problem: you didn't respect the architecte in JHLite:
- where is the domain?
- we don't use NotNull from jakarta but our own assert
- etc

Sorry but like this, it can't be merged.

Anyway, your code is not lost because you can totally use this feature in your own JHLite Custom project, and there, you can do and code as you want.

anderskristian:
Okay I see that you are not interested in a merge feature.

I hope that LITE will take over all the features that Classic can do. 
**_Otherwise I think that hipster CLASSIC and LITE will lose relevance._**

It is in that light I have made the merge work.

I fully agree that the branch cannot be merged as it is right here. It need to become a "module" - right now it does nothing.

About "discussing" --> I'm new here, so I don't know the forum to discuss things in.

I'm not sure I understand what "domain" should be in this case. This is a "tool".

Using own asserts in stead of not null. That is fine I could change that.

pascalgrimaud:
> Okay I see that you are not interested in a merge feature.

Not exactly, I'm trying to evaluate the effort to maintain the code you propose and the real value of the feature. As I said, I rarely reapply the module in my projects, that's why it's not a feature I will use it. And if I reapply the module, resolving conflict is really easy as it's only few files
About spring properties, I use a dedicated profile, so I don't have conflict

-----

> I hope that LITE will take over all the features that Classic can do.
> Otherwise I think that hipster CLASSIC and LITE will lose relevance.

Ahaha no, it's mentionned in readme: The original JHipster and JHLite are not the same thing, they are not generating the same code and not serving the same purpose!
It's 2 different projects, 2 different approaches

To resume quickly:
- you want to build your application around infrastructure (database, queue, etc) -> use generator-jhipster and use the JDL because it will help you to design your entities first
- you use XDD to design your application (TDD, BDD, DDD) -> JHipster Lite will help you there, because you don't need to think about technology. That's one of the reason why I don't want JDL in JHipster Lite

-----

> It is in that light I have made the merge work.
> When we can declare domain entities (like classic) I think that merge becomes relevant.

As there is no JDL in JHipster Lite, they won't be any conflict like this

-----

> About "discussing" --> I'm new here, so I don't know the forum to discuss things in.

No problem @anderskristian : just open an issue here and discuss :-)

-----

> I'm not sure I understand what "domain" should be in this case. This is a "tool".

Have a look at :
- https://github.com/jhipster/jhipster-lite/blob/main/documentation/hexagonal-global-schema.png
- https://github.com/jhipster/jhipster-lite/blob/main/documentation/hexagonal-flow.png

The tree should be:
- context
  - application
  - domain
  - infrastructure

With an package-info for SharedKernel or BusinessContext

anderskristian:
I will try to complete the PR with all this input.
It will take a little time --> I also have a job to handle
Then we can see if the feature is something that you want.

I'm still in the impression that Jhipster classic will fade away. 
And that something like LITE should take over.
I also think that the problems are bigger than that. But that is another thing 🐸

pascalgrimaud:
Thanks @anderskristian and feel free :)
- you can continue to improve this pull request, if that's what you decide
- or you could wait opinion from @qmonmert @murdos @DamnClin, who are active in this project too
- but in both case, as said, your code won't be lost, you could still use it in your JHLite Custom (private JHLite project)

And yes, take your time. JHipster is an open source project, it's not our main job, we do it during our freetime, including me.

> I'm still in the impression that Jhipster classic will fade away.

IMO, it's because the code is becoming too complex -> so no new contributor, that's simply as this
And that's why I don't want to bring complexity to JHipster Lite for something which can be done manually and easily

pascalgrimaud:
I'm closing this, as I think it's not something the core contributor should maintain
Anyway, this code can be totally used in a custom jhlite instance so it's not lost

anderskristian:
Okay, that is fair
I still think we need to considering making '**lite**' take over **classic's** features
Otherwise we risk both projects becomes irrelevant.

pascalgrimaud:
It has been discussed in our mailing list and the main points are:
- it's 2 different tools with 2 different approaches
- it can't be compared directly

As it's open source, the core team members will maintain the one they want, nothing is mandatory

see https://www.jhipster.tech/jhipster-lite/

PR #8069, opened 2023-11-18
Add package-info.java in dummy feature

### LINKED ISSUE #7965: Add package-info.java in dummy feature
I was playing with the dummy feature and I noticed that package-info.java was not present in the bounded context. 
In my opinion, it suppose to have a BusinessContext like package-info.

image

qmonmert:
Fix #7965


src/main/resources/generator/server/springboot/mvc/dummy/feature/main/package-info.java.mustache
```diff
@@ -0,0 +1,2 @@
+@{{packageName}}.SharedKernel
```
pascalgrimaud, line 1: I think it's business context, not a shared karnel

pascalgrimaud (changes requested):

src/main/resources/generator/server/springboot/mvc/dummy/feature/main/package-info.java.mustache
qmonmert: you're right, I fix

pascalgrimaud (approved):

PR #8159, opened 2023-11-27
Refactored basic smells

dkdal:


murdos:
I personally don't think there were smells and that code is improved.
I think it just relates to different coding habits.

murdos:
@dkdal : see my last comment, I'll close this PR. Anyway, thanks for your contribution.

Issue #4339, opened 2022-11-09
New module to add thymeleaf (+HTMX, Alpine) as "client side" technology


atomfrede:
I would like to add a module which adds

- [x] thymeleaf
- [x]  webjars 
- [x]  htmx
- [x]  alpine 
- [x]  tailwind (or bootstrap)
- [x]  skeleton layout and index files
- [x] reference added webjars in the skeleton index file (PR Pending)

Would you consider this more a server module or a client module or even split the different parts (e.g. webjars a very simple dedicated module)? 
Would you say such a module makes sense or is it not useful to have a "classic" server side rendering technology within the landscape?

This is heavily inspired by what Taming Thymeleaf is doing and what the ttcli brings in https://github.com/wimdeblauwe/ttcli

gmarziou:
I think it's a great idea. 

Personally, I'm tired of all these javascript frameworks, on the long run they require too much effort to keep your app up to date, fix vulnerabilities and maintain skills.
I imagine I'm not the only one.

I think JHipster Lite is the perfect vehicle for such approach.

atomfrede:
@gmarziou Exactly why I would like to have such a module. Most often the decision to use a full blown javascript framework is just rushed because everyone is doing it.

pascalgrimaud:
Adding a 1st bounty, to motivate someone to init this module :)

atomfrede:
I will start with the boot thymeleaf module. I think adding at at this level. The question is where to add the webjars module and htmx. I would say webjars can be added quite early, while htmx should depend on thymeleaf and webjars, so the column after thymeleaf.

<img width="1199" alt="image" src="https://user-images.githubusercontent.com/203401/230740491-985226ed-ce3a-4ee6-877b-42767b017947.png">

DamnClin:
You don't pic the level, you just need to define the dependencies, level will be calculated

atomfrede:
yes basically that's what I meant😅 it depends on spring boot module.

pascalgrimaud:
@renanfranca : I reopen this ticket as I think it's not finished, right ?

renanfranca:
> @renanfranca : I reopen this ticket as I think it's not finished, right ?

@pascalgrimaud : You are right, I will keep working on it 👍

atomfrede:
Awesome @renanfranca! If you need some help or feedback feel free to give me a ping!

renanfranca:
[...]

PR #8229, opened 2023-12-03
add htmx webjars module

atomfrede:
Moving the webjars module to webjars-locator, adding a htmx-webjars module.

updates #4339


pascalgrimaud:
maybe adding a test in the CI ?
see https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh

atomfrede:
Yes good point.

src/main/java/tech/jhipster/lite/generator/server/webjars/locator/application/WebjarsLocatorApplicationService.java
```diff
@@ -0,0 +1,20 @@
+package tech.jhipster.lite.generator.server.webjars.locator.application;
+
+import org.springframework.stereotype.Service;
+import tech.jhipster.lite.generator.server.webjars.locator.domain.WebjarsLocatorModuleFactory;
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+
+@Service
+public class WebjarsLocatorApplicationService {
```
murdos, line 9: FYI we can also design this with only one `Webjars` module with a factory that would expose multiple methods for htmx, webjars-locator, ... and multiple `JHipsterModuleResource` beans in a unique `WebjarsModuleConfiguration`.

atomfrede: Do we have an example for that? How will it be displayed in the ui?

murdos: Yes, see e.g. [src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/infrastructure/primary/CucumberModuleConfiguration.java](https://github.com/jhipster/jhipster-lite/blob/b57eb3e99c3e5189c160fd0b541412a999d08578/src/main/java/tech/jhipster/lite/generator/server/springboot/cucumber/infrastructure/primary/CucumberModuleConfiguration.java#L14)

The UI display is generated from the JHipsterModuleOrganization data of the JHipsterModuleResource, it's not related to the origin (aka the package) of the JHipsterModuleResource. Module resources with the same feature are regrouped, and a module resource is displayed according to its dependencies.

atomfrede: Thanks,  I think its a good idea, such that more webjars can be added easily if needed. Will change it accordingly

murdos (dismissed):

PR #8295, opened 2023-12-06
add skeleton layout and index files

renanfranca:
See #4339

### DIFF
--- a/src/main/resources/generator/server/springboot/thymeleaf/template/src/main/resources/templates/layout/main.html.mustache
+++ b/src/main/resources/generator/server/springboot/thymeleaf/template/src/main/resources/templates/layout/main.html.mustache
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html th:lang="|${#locale.language}-${#locale.country}|"
+      xmlns:th="http://www.thymeleaf.org"
+      xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title></title>
+    <link rel="stylesheet" th:href="@{/css/application.css}">
+</head>
+<body>
+<main layout:fragment="content">
+  <!-- jhipster-needle-thymeleaf-main-content -->
+</main>
+
+<th:block layout:fragment="script-content">
+  <!-- jhipster-needle-thymeleaf-main-script -->
+</th:block>
+</body>
+</html>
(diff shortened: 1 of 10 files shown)


renanfranca:
I added the thymeleaf skeleton layout and index files from the list:
* [x] thymeleaf
* webjars
* htmx
* alpine
* tailwind (or bootstrap)
* [X] skeleton layout and index files

I use the files from the Taming Thymeleaf CLI (https://github.com/wimdeblauwe/ttcli) project.

atomfrede:
Nice work @renanfranca . Combining it afterwards with https://github.com/jhipster/jhipster-lite/pull/8229 we are very close already to what I had in mind. Let's see how it works together eventually.

renanfranca:
> Nice work @renanfranca . Combining it afterwards with #8229 we are very close already to what I had in mind. Let's see how it works together eventually.

Thanks @atomfrede ! After that, I will work on add tailwindcss as an option to the thymeleaf template module. Then if needed we can refactor the module dependencies to make it work together.

src/main/resources/generator/server/springboot/thymeleaf/template/src/main/resources/templates/layout/main.html.mustache
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html th:lang="|${#locale.language}-${#locale.country}|"
+      xmlns:th="http://www.thymeleaf.org"
+      xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title></title>
+    <link rel="stylesheet" th:href="@{/css/application.css}">
+
```
atomfrede, line 10: Should we add a needle or so here, such that we can add scripts here (e.g. htmx via webjars)? Not sure though how it is handled elsewhere

atomfrede:
I would say go for a release, no need to have this rushed imho.

atomfrede:
@pascalgrimaud @renanfranca What do say, where should the modules to add at the webjars to the template be placed? They clearly depend on this module and the webjars locator + e.g. webjars htmx. I would put them as a dedicated module in the thymeleaf template sources folder or would you say a dedicated folder and separate modules

atomfrede:
Yes it helps a lot. Need to think about it a little more, but I tend to say the "htmx-webjar-thymeleaf-template" module can not be applied without the template module, but is clearly optional for the template module as one must not use htmx nor webjars to create a thymeleaf based application. 
Maybe let's merge this one first and I will try one structure to see what works out best.

PR #8296, opened 2023-12-06
Fix gitinfo nature: it's a shared kernel, not a business context

murdos:


murdos:
> By nature, agree it should be shared kernel. But like this, it won't be used somewhere else, as there is only 1 single configuration class. Do you have something in mind?

At work I'm generating living documentation using these annotations (including a context map using https://contextmapper.org/), and this packages pop as a business context with the other real business contexts whereas it should not.

I agree that everything within the `wire` package is most of the time not a shared kernel: would you suggest to rather remove the `package-info.java`?

pascalgrimaud:
No, I was curious :-)

Issue #8026, opened 2023-11-13
JHLite : online version without management info


pascalgrimaud:
See https://lite.jhipster.tech/management/info

There is no version:

```json
{"git":{}}
```

cc @jdubois : I don't know how the project is built but the git version is missing. Maybe it's related to how you clone the project?

zhyd1997:
![IMG_1243](https://github.com/jhipster/jhipster-lite/assets/31362988/2f83f50e-f2a3-49ad-b521-0ed4c778441b)

Works for me

pascalgrimaud:
Still missing the tag version but it's better than before, as @jdubois worked on the graalvm version

zhyd1997:
Some findings: https://github.com/search?q=repo%3Ajhipster%2Fjhipster-lite%20branch&type=code

pascalgrimaud:
Wonder if your solution can work, as currently we have:

![image](https://github.com/jhipster/jhipster-lite/assets/9156882/5a18f5e0-df6b-482c-bc18-19003f9daad9)

So it really depends on how the build is done by Julien :D

zhyd1997:
Yes, tested and the solution worked

just checkout the pr branch

run  ./mnvw and go to the info url

The release info will show

PR #8325, opened 2023-12-08
fix: added git release info.

zhyd1997:
Fix: https://github.com/jhipster/jhipster-lite/issues/8026

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -577,6 +577,7 @@
               <includeOnlyProperty>^git.commit.id.abbrev$</includeOnlyProperty>
               <includeOnlyProperty>^git.commit.id.describe$</includeOnlyProperty>
               <includeOnlyProperty>^git.branch$</includeOnlyProperty>
+              <includeOnlyProperty>^git.build.(time|version)$</includeOnlyProperty>
             </includeOnlyProperties>
             <verbose>false</verbose>
           </configuration>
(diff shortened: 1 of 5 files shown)


zhyd1997:
Hi @pascalgrimaud 

How about this, do we need to include git release time? 👀 

Screenshot 2023-12-08 at 19 36 40

zhyd1997:
and I am not familiar with Java, so do I need to add `git.build` info to these files? 👀 

<details>
  <summary>https://github.com/search?q=repo%3Ajhipster%2Fjhipster-lite+%22git-branch%22&type=code</summary>
  
  Screenshot 2023-12-08 at 19 50 33
</details>

pom.xml
```diff
@@ -577,6 +577,7 @@
               <includeOnlyProperty>^git.commit.id.abbrev$</includeOnlyProperty>
               <includeOnlyProperty>^git.commit.id.describe$</includeOnlyProperty>
               <includeOnlyProperty>^git.branch$</includeOnlyProperty>
+              <includeOnlyProperty>^git.build.(time|version)$</includeOnlyProperty>
```
zhyd1997, line 580: https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/master/docs/using-the-plugin.md

pascalgrimaud:
Indeed, I think it would be better.
You need to update Header.component.ts and Header.spec.ts to use the build.version instead of the current id.describe

See https://github.com/jhipster/jhipster-lite/blob/main/src/main/webapp/app/common/primary/header/Header.component.ts#L23

pascalgrimaud:
About Consul, yes it could be updated, but it's better to do it in another PR, not in this one

zhyd1997:
Morning @pascalgrimaud 

This is the result:

Screenshot 2023-12-09 at 09 46 54

pascalgrimaud:
thanks a lot @zhyd1997 : let's merge this

Issue #8187, opened 2023-11-29
Internationalized errors should depend spring-mvc-server


DamnClin:
`internationalized-errors` should depend on `spring-mvc-server` to have `ControllerAdvice` and other needed dependencies

atomfrede:
@pascalgrimaud I think this can be closed now.

pascalgrimaud:
Thanks @atomfrede !

PR #8329, opened 2023-12-09
updates InternationalizedErrors module dependency

atomfrede:
Corrected the dependency for `InternationalizedErrors` module and correct the api doc for webflux modules, looks like they have been mixed up with each other.

updates #8187


pascalgrimaud:
Thanks @atomfrede

PR #8330, opened 2023-12-10
fix: updated git version in consul module.

zhyd1997:
Continuing #8026


pascalgrimaud:
Thanks for improving this
You need to update this config too: https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/server/springboot/technicaltools/gitinfo/domain/GitInfoModuleFactory.java

PR #8465, opened 2023-12-28
Maven: use maven model for handling java build commands rather than DOM manipulation

### LINKED ISSUE #2413: Refactor MavenCommandHandler
There is a lot of logic duplication in `MavenCommandHandler` for the structure creation part, that needs a refactoring (but it's not that easy due to spaces management :))

murdos:
This greatly simplifies `MavenCommandHandler` since we just have to map from jhipster-lite model to maven model, and don't have to deal anymore with low level xml tree.

BREAKING CHANGE: replace JavaBuildPlugin 'additionalElements' by more specific and fine-grained 'configuration' and 'executions' properties

Adapted modules factories and tests consequently:
- groupId "org.apache.maven.plugins" is default value for plugins, and thus is not generated anymore
- order of xml elements in pom.xml is now a bit different
- (small drawback of this refactoring) it's not possible anymore to generate Xml comment in plugin configuration or executions

Fixes #2413


src/main/java/tech/jhipster/lite/generator/server/springboot/docker/domain/SpringBootDockerModuleFactory.java
```diff
@@ -45,52 +46,52 @@ private JavaBuildPlugin jibPlugin(JHipsterModuleProperties properties) {
...
-        <permissions>
-          <permission>
-            <file>/entrypoint.sh</file>
-            <mode>755</mode>
-          </permission>
-        </permissions>
-      </extraDirectories>
-    </configuration>
-    """.formatted(properties.projectBaseName().get(), properties.serverPort().get());
+  private JavaBuildPluginConfiguration jibPluginAdditionalElements(JHipsterModuleProperties properties) {
```
atomfrede, line 53: Maybe adapt the methodname too, e.g. `jibPluginConfiguration(...)`

atomfrede (commented):
Impressive work @murdos . It makes it much easier too understand!

src/main/java/tech/jhipster/lite/generator/server/springboot/docker/domain/SpringBootDockerModuleFactory.java
murdos: Indeed, thanks for the careful reading :+1:

murdos:
@atomfrede, @pascalgrimaud : do you think this can be merged? This is holding further work on maven features, and could also conflict with work in progress on supporting gradle plugins.

pascalgrimaud:
Yes go ahead

PR #8504, opened 2024-01-03
  Handle Gradle plugins in modules 

### LINKED ISSUE #4955: Handle Gradle plugins in modules
Following #1862 gradle plugins need to be handled.
This will require a bit of refactoring, since maven and gradle plugins have different identifiers (gradle only has an id, whereas maven has groupId + artifactId), and also because for a given feature maven and gradle plugin are different.

murdos:
Fixes #4955

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,6 +19,7 @@
 *.java          text
 *.js            text
 *.json          text
+*.kts           text
 *.less          text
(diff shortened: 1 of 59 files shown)


atomfrede (approved):
Nice work. I really like the api you've created to have maven and gradle plugins handled.

PR #8527, opened 2024-01-05
CI: Skip tests if there are no changes in the generated project

abdelfetah18:


murdos:
IMO https://github.com/jhipster/jhipster-lite/pull/8684 is more robust, since we don't rely on cache that could introduce side effects.
But @pascalgrimaud know better than me on the subject :-)

pascalgrimaud:
Thanks for trying to fix this complex ticket, but as @murdos said, https://github.com/jhipster/jhipster-lite/pull/8684 is better, so I'm closing this one

PR #8587, opened 2024-01-09
Delete prettier cli parameter that has been removed since version 3

murdos:


pascalgrimaud:
For your information, svelte module is not maintained anymore and is not tested in CI

PR #8590, opened 2024-01-09
Declare checkstyle as direct dependency for dependabot

murdos:
With that, dependabot should be able to automatically update version, since it does not seem to detect toolVersion


murdos:
It worked :tada: : https://github.com/jhipster/jhipster-lite/pull/8592

PR #8630, opened 2024-01-14
Test more modules with gradle in CI tests

murdos:
And fix handling of `HikariCP` dependency with gradle


src/main/java/tech/jhipster/lite/generator/server/springboot/technicaltools/gitinfo/domain/GitInfoModuleFactory.java
```diff
@@ -61,7 +61,7 @@ private GradlePlugin gradleGitPropertiesPlugin() {
         """
         gitProperties {
           failOnNoGitDirectory = false
-          keys = ['git.branch','git.commit.id.abbrev','git.commit.id.describe', 'git.build.version']
+          keys = listOf("git.branch", "git.commit.id.abbrev", "git.commit.id.describe", "git.build.version")
```
murdos, line 64: @atomfrede : FYI ;-)

atomfrede: :facepalm: of course its kotlin not groovy.

atomfrede (approved):

PR #8651, opened 2024-01-16
Add local config to use Spring Boot devtools

pascalgrimaud:
- following https://github.com/jhipster/jhipster-lite/pull/8637

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,12 @@ or
 java -jar target/*.jar --spring.profiles.active=vue
 ```
 
+You can start the application using the `local` profile to use Spring Boot devtools. It will **hot restart** your application when classes from your project are compiled.
+
+```bash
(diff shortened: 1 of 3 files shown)


murdos:
@pascalgrimaud : do you really think this is necessary?
From my understanding devtools is automatically disabled by default on production application, according to [Spring documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/using.html#using.devtools):
> Developer tools are automatically disabled when running a fully packaged application. If your application is launched from java -jar or if it is started from a special classloader, then it is considered a “production application”

pascalgrimaud:
Just tried it and you're right @murdos thanks.

By default, if we don't put these properties:
- it's disabled by default on production application
- it enabled by default when running app using Main class or with Maven, and it restarted automatically when building

No need this PR any more, I'm closing it

PR #8686, opened 2024-01-19
Minor refactorings

gzsombor:
Small changes, which I thought make sense.

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/properties/JHipsterServerPort.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/properties/JHipsterServerPort.java
@@ -2,19 +2,18 @@ package tech.jhipster.lite.module.domain.properties;
 
 import tech.jhipster.lite.shared.error.domain.Assert;
 
-public class JHipsterServerPort {
-
+public record JHipsterServerPort(int serverPort) {
   private static final int DEFAULT_PORT = 8080;
 
-  private final int serverPort;
-
-  public JHipsterServerPort(Integer port) {
-    serverPort = buildPort(port);
-
+  public JHipsterServerPort {
     Assert.field("serverPort", serverPort).min(1).max(65_535);
   }
 
-  private int buildPort(Integer port) {
+  public JHipsterServerPort(Integer port) {
+    this(buildPort(port));
+  }
+
+  private static int buildPort(Integer port) {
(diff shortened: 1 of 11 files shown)


src/main/java/tech/jhipster/lite/module/domain/properties/JHipsterServerPort.java
```diff
@@ -2,19 +2,18 @@
...
 
-  private final int serverPort;
+  public JHipsterServerPort {
+    Assert.field("serverPort", serverPort).min(1).max(65_535);
+  }
 
   public JHipsterServerPort(Integer port) {
-    serverPort = buildPort(port);
-
-    Assert.field("serverPort", serverPort).min(1).max(65_535);
```
murdos, line 14: Why are you removing this assertion?

src/main/resources/generator/buildtool/maven/pom.xml.mustache
```diff
@@ -14,7 +14,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <java.version>21</java.version>
+    <java.version>{{javaVersion}}</java.version>
```
murdos, line 17: What is your intent? 
Generated code only compiles with java 21+
By allowing user to change this property to any value, there's great chance that the generated project doesn't work.

murdos (changes requested):

src/main/java/tech/jhipster/lite/module/domain/properties/JHipsterServerPort.java
gzsombor: The assertion is moved into the default constructor, so it's still apply

src/main/resources/generator/buildtool/maven/pom.xml.mustache
gzsombor: The user can't change the property, it is set in the JHipsterModuleProperties as a final variable.
I don't want to repeat the same 21 in every template, and make it clear what's that number is about.
So I can write a template, without the need to constantly update, whenever a new java version is released.

murdos (approved):

pascalgrimaud:
Happy to see you here @gzsombor !
I didn't expect you followed this project. Thanks for your help and your contribution :)

PR #8692, opened 2024-01-20
More minor additions

gzsombor:


src/test/java/tech/jhipster/lite/module/domain/properties/JHipsterServerPortTest.java
```diff
@@ -18,6 +21,12 @@ void shouldGetServerPortFromPort() {
     assertThat(new JHipsterServerPort(9000).get()).isEqualTo(9000);
   }
 
+  @Test
```
murdos, line 24: A `@ParametrizedTest` would be better.
Also we usually use assertj assertions rather than junit ones.

murdos (approved):

PR #8694, opened 2024-01-20
Group maven/gradle wrappers as 'build-launcher' and add dependencies

gzsombor:


src/main/java/tech/jhipster/lite/generator/slug/domain/JHLiteFeatureSlug.java
```diff
@@ -15,6 +15,7 @@ public enum JHLiteFeatureSlug implements JHipsterFeatureSlugFactory {
   DUMMY_SCHEMA("dummy-schema"),
   FRONT_BROWSER_TEST("front-browser-test"),
   JAVA_BUILD_TOOL("java-build-tool"),
+  JAVA_BUILD_LAUNCHER("java-build-launcher"),
```
murdos, line 18: How about "java-build-tool-wrapper"?
I'm not sure that "build-launcher" is meaningful.

gzsombor: That could work as well :smile:

murdos (approved):

pascalgrimaud:
Not sure about this change, as maven-java already added the maven-wrapper

I add this module as standalone for 2 reasons:
- to add this wrapper on existing projects which were not generated with JHLite
- to sometimes upgrade wrapper

So I think it should stay as standalone and not depend on other module

gzsombor:
Yes, I've just noticed, that the maven wrapper is still installed. I would rather remove it from when MAVEN_JAVA is applied. Currently I'm trying to create a more modern project setup, with a declarative version and environment manager like mise/rtx/asdf, so these wrapper executables are very misleading in my project setup.

pascalgrimaud:
Yes it could be possible to remove it from the maven_java, but in this case, it should be rename to maven_pom_xml_java

I would suggest to open a new ticket to discuss about what we should do for this maven-wrapper, because with this change, this module becomes useless :-p

pascalgrimaud:
And renaming module is a small breaking change for existing projects

murdos:
> Yes it could be possible to remove it from the maven_java, but in this case, it should be rename to maven_pom_xml_java

Sorry, I did not notice this issue before merging the PR...
I would also remove the maven wrapper from the maven_java module, I often encounter projects that use maven but don't need the maven wrapper. BTW I just verified: the gradle module also generates the wrapper, but I think I just did that to mimic the maven behavior. I was persuaded that gradle wrapper was an opt-in option with the gradle-wrapper module.

However I don't think we should rename "maven_java", event if we remove the wrapper: even if there's just a pom.xml generated it's a building block of all other modules, and marks the choice of the java build tool.
I don't think you would want to rename the gradle_java module to "gradle_build.gradle_settings_gradle_libs_versions_java" (corresponding to the 3 generated files for gradle: `build.gradle.kts`, `settings.gradle.kts` and `gradle/libs.versions.toml`)

gzsombor:
I've opened a ticket about this https://github.com/jhipster/jhipster-lite/issues/8697 for further discussions

PR #8684, opened 2024-01-19
Improve ci to not launch same tests

renanfranca:
Fix #6651

PR #8711, opened 2024-01-22
CI: reproducible generated projects with jwtBase64Secret property

renanfranca:
Needed by #6651 and #8684


renanfranca:
@pascalgrimaud : your implementation (https://github.com/jhipster/jhipster-lite/pull/8684#issuecomment-1902072365) was easy to follow, Thanks!

Issue #1627, opened 2022-05-07
CI: status-check with skip ci message


pascalgrimaud:
When using [skip ci] in commit message, the status-check job is pending, so only admin can merged, so it's a problem:

![image](https://user-images.githubusercontent.com/9156882/167247081-657ff5d7-3bc0-495c-a417-3973daa6282f.png)

pascalgrimaud:
@renanfranca : I changed your role as admin of this repo, you should be able to access to it

renanfranca:
@pascalgrimaud : I discovered that the words `[skip ci]` and `[ci skip]` are used by github action to skip all jobs by default. That's why the status-check wasn't executed. The message that requires admin's roles to merge is always there and fades away after the pipeline starts.

The variable `github.event.head_commit.message` return an empty string when executed at a pull request.

I am going to create a pull request to implement the feature that skips the ci, is that ok? I tried the word [not ci] and it is not reserved, feel free to suggest a better one 😃👍

pascalgrimaud:
mhh I see. Can't we force to run the status-check even if the words contain `[skip ci]` or `[ci skip]` ?

renanfranca:
> mhh I see. Can't we force to run the status-check even if the words contain `[skip ci]` or `[ci skip]` ?

I changed  status-checks implementation without any conditions and without "needs". Tested without the `[skip ci]` or `[ci skip]` and it runs... with `[skip ci]` or `[ci skip]` it didn't run and any jobs. When using `[skip ci]` or `[ci skip]` no entrance appears at Actions.

https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

renanfranca:
After some research, I realized that if the workflow is trigger by push or pull_request or using branch filtering like it is today:

```yml
name: build
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || (github.ref == 'refs/heads/main' && github.sha) || github.ref }}
  cancel-in-progress: true
on:
  workflow_dispatch:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```
There isn't a way to avoid the default behavior of skipping complete the overflow and letting the status-checks pending when using the tags: 
```
[skip ci]
[ci skip]
[no ci]
[skip actions]
[actions skip]
```

Here are some references:

[...]

PR #8741, opened 2024-01-25
Skip ci with keywords at head commit message or pull request title

renanfranca:
See
- https://github.com/jhipster/jhipster-lite/issues/1627#issuecomment-1912109856

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -11,6 +11,47 @@ on:
     branches:
       - main
 jobs:
+  #------------------------------------------------------
+  # Setup values which is going to be used in other jobs
+  #------------------------------------------------------
+  setup:
+    name: setup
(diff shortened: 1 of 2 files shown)


renanfranca:
@pascalgrimaud : here is the change proposal for ignoring the ci. I saw that the `tests-windows` don't have the if needed to ignore the job, so I kept it that way.

I am looking for an alternative to use the [ci skip] and [skip ci] keywords which is more elegant: https://stackoverflow.com/questions/76103981/how-can-i-override-github-actions-or-cloud-build-skip-ci

renanfranca:
I am going to check that bug at my fork:
image

renanfranca:
> I am going to check that bug at my fork: image

It is fixed: must remove commit line breaks.

pascalgrimaud:
thanks @renanfranca : let's give it a try :)

PR #8757, opened 2024-01-26
Revert "chore(deps-dev): bump husky from 8.0.3 to 9.0.6 in /src/main/resources/generator/dependencies/common"

qmonmert:
Reverts jhipster/jhipster-lite#8749


pascalgrimaud:
why reverting this @qmonmert ? is it broken ?

qmonmert:
@pascalgrimaud I had a doubt after merging manually comparing version (why we don't have 9.0.0, 9.0.1 ...)
but it must be good
https://www.npmjs.com/package/husky?activeTab=versions
image

PR #8760, opened 2024-01-27
Bump jhlite dependency to 1.3.0

murdos:


pascalgrimaud:
The main problem is:
- when generating jhlite custom, it will use the previous release instead of the last one
- in the CI, we can't use the last release as it doesn't exist yet

I don't know how we could solve it. Any idea?

murdos:
@pascalgrimaud : I thought a bit about it, and I think we might solve that by:
- using maven resources filtering to replace jhlite version in `src/main/resources/generator/dependencies/pom.xml` by current jhlite version, using `${project.version}` => that way custom-jhlite will always use the version of jhlite it has been generated with
- in CI, for custom-jhlite tests, extract the jhlite version required by custom-jhlite, and install in the maven local repository the jhlite-jar with the extracted version. `mvn install:install-file -Dfile=jhlite.jar -DgroupId=tech.jhipster.lite -DartifactId=jhlite -Dversion=<extracted-version>`

WDYT?

pascalgrimaud:
Yes sound good
For th CI, we could use snapshot version too

PR #8776, opened 2024-01-29
Add JaCoCo configuration to check minimum coverage

murdos:
This will help checking coverage locally when developing jhlite: a first feedback without the need to wait for CI build and Codecov report.

### DIFF
--- a/.jhipster/modules/history.json
+++ b/.jhipster/modules/history.json
@@ -3,5 +3,25 @@
     "module" : "application-service-hexagonal-architecture-documentation",
     "date" : "2022-07-29T05:45:39.864617462Z",
     "properties" : { }
+  }, {
+    "module" : "maven-java",
+    "date" : "2024-01-29T22:31:20.121046866Z",
+    "properties" : {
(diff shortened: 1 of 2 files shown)


murdos:
After more investigation, some tests are only ran on Linux, using `@EnabledOnOs(OS.LINUX)`, that's why the build fails on Windows.
I'm inclined to move this specific jacoco configuration to a maven profile that would be activate only for Linux.

renanfranca:
@murdos : I checked out this pull request branch, ran the command `./mvnw verify`, and failed. It should pass, isn't it? Am I doing something wrong?

image
image

image
image

murdos:
> @murdos : I checked out this pull request branch, ran the command `./mvnw verify`, and failed. It should pass, isn't it? Am I doing something wrong?

Hmm... it works fine for me, and in CI... Maybe try to clean before, to be sure: `./mvnw clean verify`

renanfranca (approved):
Thank you for your work! 👍

PR #8790, opened 2024-01-30
Fix docker image name in README

murdos:


pascalgrimaud:
Yes, I think this command is for using the image locally

renanfranca:
> Yes, I think this command is for using the image locally

I see... using the local registry. But the previous command `./mvnw spring-boot:build-image -Pnative` from the READ.me instructions pushed the image to docker.io.

murdos:
> Yes, I think this command is for using the image locally

You're right, I've been confused.

PR #8829, opened 2024-02-04
Some typos

qmonmert:


src/test/java/tech/jhipster/lite/module/domain/JHipsterModuleContextTest.java
```diff
@@ -20,7 +20,7 @@ class JHipsterModuleContextTest {
   private LogsSpy logs;
 
   @Test
-  void shouldGetIndentSizeationFromInvalidIndentation() {
+  void shouldGetIndentSizingFromInvalidIndentation() {
```
murdos, line 23: Should it be indent *size* rather than*sizing*?

qmonmert: yep I update

PR #8842, opened 2024-02-05
Fix getter for java version property

murdos:
Convention in jhlite is to not use get prefix


renanfranca (approved):
Thank you for your work :)

PR #8879, opened 2024-02-11
(engine) Activate more checkstyle rules and fix existing code

murdos:


renanfranca (approved):
IMO, the more checks there are, the better. Thank you for your work!

pascalgrimaud:
Thanks for your work @murdos 
Anyway, I'm not totally sure about the final class rules, as in JHLite Custom, you won't be able to extend anymore these classes. Right now, no impact for me, but I think it will reduce the possibility to customize.

murdos:
@pascalgrimaud : these classes are already final because they have a private constructor.
The FinalClass rule just makes sure the contract is explicit when it's already impossible to extend the class.

pascalgrimaud:
Oh I see, so I'm fine

PR #8933, opened 2024-02-18
bump MAVEN to 3.9.6

qmonmert:


src/main/resources/generator/buildtool/maven/pom.xml.mustache
```diff
@@ -18,7 +18,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <maven.version>3.9.5</maven.version>
+    <maven.version>3.9.6</maven.version>
```
murdos, line 21: I don't see the reason to require users to always use the latest maven version.
Everybody does not use maven wrapper, and does not jump on latest release as soon as it's release.
So IMO we should be more flexible, and even require a 3.8.0 version.

qmonmert: thank you for your opinion

qmonmert:
@pascalgrimaud @murdos so we close this?

pascalgrimaud:
I agree with @murdos : we should not force the maven version here, as some users could use an installed version, instead of the wrapper

qmonmert:
@pascalgrimaud ok I close

PR #8986, opened 2024-02-22
(ci) Introduce java-build-tool in build matrix

murdos:
It allows better handling of build cache, and this will help testing more easily different gradle applications.
This will also solve the issue of applying a feature rather than a module: since we can't apply through JHipster REST API a feature 'github-actions' that regroups either 'github-actions-maven' or 'github-actions-gradle', we can either have bash function that return the right module for the java-build-tool or use expression like 'github-actions-${java_build_tool}'

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -394,7 +394,7 @@ jobs:
             build.gradle.kts \
             settings.gradle.kts \
             gradle/libs.versions.toml \
-            gradle/wrapper/gradle-wrapper.properties\
+            gradle/wrapper/gradle-wrapper.properties \
             | md5sum | head -n1 | cut -d " " -f1)
           echo $GRADLE_DEPS_MD5
           echo "md5sum=${GRADLE_DEPS_MD5}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/workflows/github-actions.yml
```diff
@@ -320,7 +323,7 @@ jobs:
         working-directory: ./current-branch/tests-ci/
         run: |
           ./start.sh 7471
-          ./generate.sh ${{ matrix.app }} ${{ matrix.spring-config-format }}
+          ./generate.sh ${{ matrix.app }} ${{ matrix.java-build-tool }} ${{ matrix.spring-config-format }}
```
renanfranca, line 326: Should this change need to be done at the following step too?

```yml
      - name: 'Generation: generate ${{ matrix.app }} from main branch'
        if: github.event_name == 'pull_request'
        working-directory: ./main-branch/tests-ci/
        run: |
          ./start.sh 7471
          ./generate.sh ${{ matrix.app }} ${{ matrix.spring-config-format }}
          ./stop.sh
```

murdos: I thought about it, but I came to the conclusion that this change should be made once this PR is merged, because script of main branc still uses 2 input parameters.

renanfranca: Ohh! That's true, thanks 👍

PR #9003, opened 2024-02-24
bump Vue 3.4.19

qmonmert:


murdos:
I don't understand why you're doing the version bump manually instead of letting dependabot doing it automatically.

qmonmert:
@murdos because the version is not updated since `3.4.13` 

that's why...

image

qmonmert:
and sometimes it's interesting to watch the version bump (ex: #9004 ) ....

murdos:
> @murdos because the version is not updated since `3.4.13`
> 
Hmm... what would be interesting is to investigate why it's not updated automatically :/

qmonmert:
at least I noticed it!

so this PR is not "interesting" ... you can close this PR if you want and find a better solution ...

murdos:
Closing in favor of #9009: we'll then see if dependabot is able to update vuejs automatically.

murdos:
@renanfranca : I did not succeed in founding the root cause of dependabot not doing the update. I've opened #9030 to track the issue for future vuejs releases, and I propose to manually update vuejs version with your PR :+1:

qmonmert:
@murdos ok I update this PR

murdos:
@qmonmert : Thanks!

PR #9005, opened 2024-02-24
bump all eslint dependencies

qmonmert:


murdos:
Updates already done by dependabot:  #9007 and #9008

PR #9009, opened 2024-02-24
Update obsolete node version to unlock dependabot updates

murdos:
Updating obsolete node version declared as engine in package.json should unlock automatic updates of vuejs with dependabot since vuejs is requiring node >=18.12.0.

See https://github.com/dependabot/dependabot-core/issues/1633#issuecomment-577135095 for a similar issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "vue-tsc": "1.8.27"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


renanfranca (approved):
Thank you for the research!

PR #9015, opened 2024-02-24
Remove useless cacheDirectories in package.json

murdos:
- this is a non-standard configuration that is not mentionned in https://docs.npmjs.com/cli/v10/configuring-npm/package-json
- it's a specific heroku configuration, that is not useful for us, especially as its value is the default value: https://devcenter.heroku.com/articles/nodejs-support#custom-caching

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,8 +93,5 @@
   },
   "engines": {
     "node": ">=20.11.1"
-  },
-  "cacheDirectories": [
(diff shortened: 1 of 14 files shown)


murdos:
@pascalgrimaud : any objections? Maybe there was a good reason to configure that, that I'm missing.

Issue #8963, opened 2024-02-20
Handle different frontend build output directory for maven and gradle


renanfranca:
Today the output directory is fixed to the `target` folder. When generating a gradle project the output folder should be `build`.

It will be challenging because you can apply a frontend module without choosing the `java-build-tool`, so the generator can't automatically detect the correct output directory based on `pom.xml` or `build.gradle.kts` existence.

Vue configuration which needs to be fixed `vitest.config.mts`:
```mts
 test: {
    reporters: ['verbose', 'vitest-sonar-reporter'],
    outputFile: {
      'vitest-sonar-reporter': 'target/test-results/TESTS-results-sonar.xml',
    },
```
```mts
      provider: 'istanbul',
      reportsDirectory: 'target/test-results/',
      reporter: ['html', 'json-summary', 'text', 'text-summary', 'lcov', 'clover'],
```

`vite.config.mts`:
```mts
  build: {
    outDir: '../../../target/classes/static',
  },
```

Needed by:
- https://github.com/jhipster/jhipster-lite/pull/8867#issuecomment-1947650464

murdos:
@renanfranca : Could I work on this issue? I've my own idea that I would like to experiment.
My intent is to enrich `JHipsterModuleContext` in `JHipsterModulesApplyer` with additional properties in function of the detected build tool, such as `project.build.directory`, so it could be use in templates.  `project.build.directory` would be valued to `target` for maven (and probably also if no java build tool is found), and to `build` for gradle.

renanfranca:
> @renanfranca : Could I work on this issue? I've my own idea that I would like to experiment.
> My intent is to enrich `JHipsterModuleContext` in `JHipsterModulesApplyer` with additional properties in function of the detected build tool, such as `project.build.directory`, so it could be use in templates.  `project.build.directory` would be valued to `target` for maven (and probably also if no java build tool is found), and to `build` for gradle.
> 
> **Update**: #9029 is a step in that direction

@murdos : Sure, go for it! I over think about some approaches too and I will be happy to review the code when you finish it. I agree to be target folder if didn't detect any build tool.

murdos:
@pascalgrimaud Bounty claimed: https://opencollective.com/generator-jhipster/expenses/203259

pascalgrimaud:
@murdos : approved

PR #9025, opened 2024-02-25
Startup commands improvements

murdos:


renanfranca:
Thanks, Interesting improvements! I am going to take a closer look as it is an opportunity to learn 😊

renanfranca (approved):
@murdos : Thank you very much! 💙


This new switch features makes me want to use the interfaces instead of enums. There was a lot of stuffs going on at the interface StartupCommand:
-  including the supresswarning because it's in different records
- more than one "public" records in the same file which makes sense for me, but sometimes it isn't considere best practices by my co-works

I am going to use this approach at my next issue #8963 !

murdos:
@renanfranca : thanks for the feedback! I've moved all implementations of StartupCommand to their own top-level file :)

renanfranca:
> I've

Nice! You are welcome 👍😁

PR #9029, opened 2024-02-26
Factorize java build tool detection

murdos:
And filter startupCommands in domain, with JHipsterModulesApplyer, rather in infrastructure. This makes `FileSystemStartupCommandsReadmeCommandsHandler` logic much simpler by reducing its responsibility.


renanfranca:
Sorry, I only saw the request review now. I am going to review it when i wake up 😁👍

src/main/java/tech/jhipster/lite/module/domain/JHipsterModulesApplyer.java
```diff
@@ -61,7 +76,7 @@ public JHipsterModuleApplied apply(JHipsterModuleToApply moduleToApply) {
       .filesToMove(module.filesToMove())
       .filesToDelete(module.filesToDelete())
       .replacers(buildReplacers(module))
-      .startupCommands(module.startupCommands())
+      .startupCommands(buildStartupCommands(module.startupCommands(), detectedJavaBuildTool))
```
renanfranca, line 79: Thank you for your work! At this point when the method `buildStartupCommands` is called the template files weren't written on the system file yet.

src/main/java/tech/jhipster/lite/module/domain/JHipsterModulesApplyer.java
```diff
@@ -92,6 +107,30 @@ public JHipsterModuleApplied apply(JHipsterModuleToApply moduleToApply) {
     return moduleApplied;
   }
 
+  private JHipsterStartupCommands buildStartupCommands(
+    JHipsterStartupCommands jHipsterStartupCommands,
+    Optional<JavaBuildTool> detectedJavaBuildTool
+  ) {
```
renanfranca, line 113: IMO, this method should look for the pom.xml and build.gradle.kts at the template files to add (if found here the build tool is detected) and the files to delete (if found here this build tool cant be an option anymore)

renanfranca:
```suggestion
  private JHipsterStartupCommands buildStartupCommands(
    JHipsterStartupCommands jHipsterStartupCommands,
    Optional<JavaBuildTool> detectedJavaBuildTool,
    JHipsterTemplatedFiles filesToAdd,
    JHipsterFilesToDelete filesToDelete
  ) {
```

src/main/java/tech/jhipster/lite/module/domain/JHipsterModulesApplyer.java
renanfranca: It's TDD all the way 😁! Really beautiful solution!

> I've not handled removal of pom.xml and build.gradle.kts since this is probably something that will never happen, so there's no reason to make code more complex. If one day the need appears, we'll evolve code :-)

Sure! I agree!

renanfranca (approved):
Thank you for your work! 😃

qmonmert:
Coverage is not 100%
https://app.codecov.io/gh/jhipster/jhipster-lite/pull/9029?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=jhipster
image

murdos:
> Coverage is not 100% 

Thanks, I'll fix it!

PR #9033, opened 2024-02-26
Use maven-model-helper library to write changes to the POM model

murdos:
This will allow better preservation of original comments and order of existing elements when updating a pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
     <sonar-maven-plugin.version>3.10.0.2594</sonar-maven-plugin.version>
     <apache-commons-io.version>2.15.1</apache-commons-io.version>
     <cucumber.version>7.15.0</cucumber.version>
-    <maven-model.version>3.9.6</maven-model.version>
+    <maven-model-helper.version>26-SNAPSHOT</maven-model-helper.version>
     <night-config.version>3.6.7</night-config.version>
     <testcontainers.version>1.19.6</testcontainers.version>
     <mongock.version>5.4.0</mongock.version>
@@ -160,9 +160,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.maven</groupId>
-      <artifactId>maven-model</artifactId>
-      <version>${maven-model.version}</version>
+      <groupId>io.fabric8</groupId>
+      <artifactId>maven-model-helper</artifactId>
+      <version>${maven-model-helper.version}</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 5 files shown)


murdos:
@gastaldi : FYI :-)

pom.xml
```diff
@@ -71,7 +71,7 @@
     <sonar-maven-plugin.version>3.10.0.2594</sonar-maven-plugin.version>
     <apache-commons-io.version>2.15.1</apache-commons-io.version>
     <cucumber.version>7.15.0</cucumber.version>
-    <maven-model.version>3.9.6</maven-model.version>
+    <maven-model-helper.version>28</maven-model-helper.version>
     <dom4j.version>2.1.4</dom4j.version>
```
gastaldi, line 75: You probably don't need this as it's resolved transitively?

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/maven/MavenCommandHandler.java
```diff
@@ -398,10 +393,11 @@ private Function<MavenPluginConfiguration, Xpp3Dom> toMavenConfiguration() {
...
   private void writePom() {
-    try (Writer fileWriter = Files.newBufferedWriter(pomPath, StandardCharsets.UTF_8)) {
-      StringWriter stringWriter = new StringWriter();
-      mavenWriter.write(stringWriter, pomModel);
-      fileWriter.write(applyIndentation(stringWriter.getBuffer().toString()));
+    Maven.writeModel(pomModel, pomPath);
+
+    try {
+      String pomContent = Files.readString(pomPath, StandardCharsets.UTF_8);
+      Files.writeString(pomPath, applyIndentation(pomContent), StandardCharsets.UTF_8);
```
gastaldi, line 400: Maybe https://github.com/fabric8-launcher/maven-model-helper/issues/117 would help to avoid I/O until needed

murdos: It might indeed :)

pom.xml
murdos: I think this dependency is not used anymore, but I need to check (I'll do that in another PR).
This is however unrelated to this PR, since maven-model-help use jdom, not dom4j :-)

murdos: Removed with https://github.com/jhipster/jhipster-lite/pull/9039

PR #9058, opened 2024-02-27
Fix missing defined lib so no update

### LINKED ISSUE #8258: Bug: missing defined lib so no update
We need to check for each version, if the lib is defined in the pom.xml
Because, when the lib is missing, the version is not updated by dependabot

See https://github.com/jhipster/jhipster-lite/blame/main/src/main/resources/generator/dependencies/pom.xml

image

murdos:
- Added unit tests that check that every property version is associated with a dependency, and that all dependencies are defining a version
- Fixed existing violations
- Added missing `<type>pom</type>` for bom dependencies

Fixes #8258

### DIFF
--- a/src/main/resources/generator/dependencies/pom.xml
+++ b/src/main/resources/generator/dependencies/pom.xml
@@ -180,10 +180,6 @@
         <version>${jib-maven-plugin.version}</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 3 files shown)


murdos:
FYI I've merged the PR to avoid conflict with dependabot updates.
Feel free to review it afterwards, I'm open to comments and I'll handle them if needed :-)

PR #9090, opened 2024-02-29
Sonar: %n should be used in place of \n to produce the platform-specific line separator.

qmonmert:
Fix https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=jhipster_jhipster-lite&open=AY32YPdwSjlKq-G0oiu5&tab=code


qmonmert:
@murdos `\\n` are really necessary in the regex because tests are ✅ without?

murdos:
> @murdos `\\n` are really necessary in the regex because tests are ✅ without?

@qmonmert : Good point. I was not aware until now, but `\s` match return line `\n`, so line separator is indeed not necessary.
I've pushed to your branch a modification of an existing test using a dependency with multiple exclusions, to be sure that this case is covered.

So +1 for removing the line separator :-)

qmonmert:
@murdos seems good

murdos:
Thanks @qmonmert ! 🙂

qmonmert:
@murdos @pascalgrimaud ✅
https://sonarcloud.io/summary/overall?id=jhipster_jhipster-lite
image

PR #9098, opened 2024-03-02
Rename "dummy" feature to "sample" feature

murdos:
See https://github.com/jhipster/jhipster-lite/discussions/7728: using "sample" is much clearer than using "dummy" for generated sample code


pascalgrimaud:
I think this project https://github.com/jhipster/jhipster-lite-sample-app needs to be updated too, @murdos ?

murdos:
@pascalgrimaud : Right! Done :smiley:

PR #9118, opened 2024-03-04
Fix gradle default tasks command

renanfranca:


murdos (approved):

murdos:
Thanks @renanfranca ! I also noticed this issue but I had not figured why 😅

renanfranca:
> Thanks @renanfranca ! I also noticed this issue but I had not figured why 😅

You are welcome! Me too 😅😁. I only find out the why because of my current issue.

PR #9119, opened 2024-03-04
add develocity integration

atomfrede:
This PR adds leverages our own [gradle develocity ](https://ge.jhipster.tech/) to have build scans and build cache support for CI/CD and on local developer machines. @jdubois already added the needed secret to have write access from our github runners to gradle develocity. Anyone can read from the remote cache and of course use the local build cache. 

If you would like to know more about develocity or would like to have an own user account just give me a ping.

We should squash the commits :)

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -39,6 +39,13 @@ jobs:
         run: |
           ./mvnw clean verify
 
+      - name: Archive results
+        uses: actions/upload-artifact@v4
+        with:
+          name: target
+          path: |
+            target/**/*.*
+
   #--------------------------------------------------
   # Build and Tests the project on Linux
   #--------------------------------------------------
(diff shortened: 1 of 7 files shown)


murdos (approved):

murdos:
@atomfrede : Awesome! 
What can you do with your own user account?

atomfrede:
An account gives you permission to write also to the build cache, api access (which I only needed for the gradle develocity experiments during on boarding) and you can e.g. purge/clear the build cache. All other things, like using the build cache, analyzing builds can be done anonymous.

pascalgrimaud:
We need to be sure it will work for JHLite Custom and in private enterprise

atomfrede:
You mean if someone clones the repo and uses it internally? I think there won't be issues as the develocity plugin steps down gracefully when e.g. firewall or so blocks it. Or we just make it opt in and enable it for our ci for example

atomfrede:
Can we somehow reliable test it?

pascalgrimaud:
@murdos would be able to test it after his holidays, as he built a JHLite Custom for his customer :)

murdos:
I tested it quickly in private enterprise, and I've not encountered any issues 🙂

Issue #9032, opened 2024-02-26
VSCode implicitly import java.lang.StringTemplate.STR which conflict with org.yaml.snakeyaml.nodes.Tag.STR


renanfranca:
I am a vscode user and this compilation error (`Type mismatch: cannot convert from StringTemplate.Processor<String,RuntimeException> to Tag`) at class `YamlFileSpringPropertiesHandler.java` is so annoying:
  ```java
    private Node buildScalarNode(Object value) {
      Tag tag = STR;
      if (value instanceof Integer || value instanceof Long) {
        tag = INT;
      } else if (value instanceof Double || value instanceof Float) {
        tag = FLOAT;
      } else if (value instanceof Boolean) {
        tag = BOOL;
      }
  
      return new ScalarNode(tag, value.toString(), null, null, DumperOptions.ScalarStyle.PLAIN);
    }
  ```
  ![image](https://github.com/jhipster/jhipster-lite/assets/6011616/65080c20-63f1-46bb-b3d5-83b543517997)
  
  Every time before starting work on the project I have to explicitly set Tag.STR to solve the problem:
  ```java
    private Node buildScalarNode(Object value) {
      Tag tag = Tag.STR;
      if (value instanceof Integer || value instanceof Long) {
        tag = INT;
      } else if (value instanceof Double || value instanceof Float) {
        tag = FLOAT;
      } else if (value instanceof Boolean) {
        tag = BOOL;
      }
  
      return new ScalarNode(tag, value.toString(), null, null, DumperOptions.ScalarStyle.PLAIN);
    }
```

I opened an issue:
- https://github.com/redhat-developer/vscode-java/issues/3510

renanfranca:
@murdos and @pascalgrimaud : Do you face that problem using intellij?

renanfranca:
> you can open a PR to change existing code :-)

Thank you! 😅☺️

PR #9171, opened 2024-03-08
Fix: Unnecessarily qualified statically imported element

qmonmert:


src/main/java/tech/jhipster/lite/module/infrastructure/secondary/YamlFileSpringPropertiesHandler.java
```diff
@@ -148,7 +148,7 @@ private static Predicate<NodeTuple> nodeTupleKeyEquals(String partialKey) {
   }
 
   private Node buildScalarNode(Object value) {
-    Tag tag = Tag.STR;
+    Tag tag = STR;
```
murdos, line 151: We should keep it as it is, see https://github.com/jhipster/jhipster-lite/issues/9032#issuecomment-1964486864

qmonmert: done

murdos (approved):

PR #9237, opened 2024-03-16
Fix deprecated methods

qmonmert:
https://javadoc.io/doc/io.jsonwebtoken/jjwt-api/latest/deprecated-list.html

### DIFF
--- a/src/main/resources/generator/server/springboot/mvc/security/jwt/authentication/test/infrastructure/primary/JwtReaderTest.java.mustache
+++ b/src/main/resources/generator/server/springboot/mvc/security/jwt/authentication/test/infrastructure/primary/JwtReaderTest.java.mustache
@@ -9,7 +9,6 @@ import {{packageName}}.LogsSpyExtension;
 import {{packageName}}.UnitTest;
 import io.jsonwebtoken.JwtParser;
 import io.jsonwebtoken.Jwts;
-import io.jsonwebtoken.SignatureAlgorithm;
 import io.jsonwebtoken.security.Keys;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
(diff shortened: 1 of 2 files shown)


src/main/resources/generator/server/springboot/mvc/security/jwt/authentication/test/infrastructure/primary/JwtReaderTest.java.mustache
```diff
@@ -87,7 +86,7 @@ class JwtReaderTest {
     return Jwts
       .builder()
       .subject("test")
-      .signWith(KEY, SignatureAlgorithm.HS512)
+      .signWith(KEY, Jwts.SIG.HS256)
```
murdos, line 89: Is there a reason for switching from HS512 to HS256?

qmonmert: yes it's deprecated 
image

murdos: SIG.HS512 is not deprecated: https://javadoc.io/doc/io.jsonwebtoken/jjwt-api/latest/io/jsonwebtoken/Jwts.SIG.html

qmonmert: You don’t understand, it’s the signature which is deprecated like I said in the commit. Generate an app and you will see

qmonmert: image

image

murdos: You should really read what I wrote, twice.
It's not about the signature change, it's about the change of the length of the key, without justification (your commit doesn't explain why you're changing the length of the key).
```suggestion
      .signWith(KEY, Jwts.SIG.HS512)
```
(and yes, I've generated the app, I've seen that using 512 bit still works)

PR #9099, opened 2024-03-02
(frontend-maven-plugin) Install node in target directory

murdos:
As done by generator-jhipster: this avoid an additional top-level directory in project

PR #9239, opened 2024-03-16
make npm tasks cachable again via gradle enterprise plugin

atomfrede:
After setting the `installDirectory` property the npm tasks became not cachable (as it is a file). We don't cache the node/npm installation so we can safely ignore it for the cacheable confiuration for npm install, npm build and npm tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -774,9 +774,6 @@
                           <property>
                             <name>testFailureIgnore</name>
                           </property>
-                          <property>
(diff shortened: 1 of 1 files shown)


murdos (approved):

murdos:
Thanks @atomfrede !
Is this a side effect of #9099? I guess we should now be careful when we change configuration of plugins, or id of executions.

PR #9246, opened 2024-03-17
Useless javadoc

qmonmert:


murdos (commented):
Such javadoc doesn't bring any value, it should rather be deleted.

PR #9248, opened 2024-03-17
(fix) Missing output directory for npm run test in via gradle enterprise plugin configuration

murdos:
Following #9239
Because of that Sonar execution was failing because it could not find `target/test-results/TESTS-results-sonar.xml`


atomfrede (approved):

atomfrede:
Thanks @murdos

PR #9269, opened 2024-03-18
Remove gradle checkstyle workaround fix

murdos:
It doesn't seem necessary anymore with version 10.14.2 of checkstyle


renanfranca (approved):
Thank you for your work! 😊

PR #9332, opened 2024-03-23
Remove H2 dependency from liquibase

### LINKED ISSUE #7516: Liquibase: remove H2 dependency?
It seems to be only used for `SpringLiquibaseUtilTest`

swarajsaaj:
Remove H2 dependency and fix the SpringLiquibaseUtilTest to use datasource property

Fix #7516

### DIFF
--- a/src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
+++ b/src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
@@ -4,22 +4,36 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
 
+import java.net.URISyntaxException;
+import java.util.Properties;
+
 import com.zaxxer.hikari.HikariDataSource;
 import javax.sql.DataSource;
 import liquibase.integration.spring.SpringLiquibase;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.autoconfigure.liquibase.DataSourceClosingSpringLiquibase;
 import org.springframework.boot.autoconfigure.liquibase.LiquibaseProperties;
 import org.springframework.boot.jdbc.DataSourceBuilder;
+import org.springframework.core.io.PathResource;
 import {{packageName}}.UnitTest;
 
 @UnitTest
 class SpringLiquibaseUtilTest {
 
+  private static String datasourceUrl;
+
+  @BeforeAll
+  public static void setup() throws URISyntaxException {
+    datasourceUrl = getTestProperties().getProperty("spring.datasource.url");
+    System.out.println(datasourceUrl);
+  }
+
   @Test
   void createSpringLiquibaseFromLiquibaseDataSource() {
-    DataSource liquibaseDatasource = DataSourceBuilder.create().url("jdbc:h2:mem:liquibase").username("sa").build();
+    DataSource liquibaseDatasource = DataSourceBuilder.create().url(datasourceUrl).username("sa").build();
     LiquibaseProperties liquibaseProperties = null;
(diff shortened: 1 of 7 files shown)


src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
```diff
@@ -204,4 +218,10 @@ class SpringLiquibaseUtilTest {
       )
       .isExactlyInstanceOf(NullPointerException.class);
   }
+
+  private static Properties getTestProperties() throws URISyntaxException {
+    YamlPropertiesFactoryBean yaml = new YamlPropertiesFactoryBean();
+    yaml.setResources(new PathResource(SpringLiquibaseUtilTest.class.getClassLoader().getResource("config/application-test.yml").toURI()));
```
murdos, line 224: This will not always work since Spring configuration can also be using properties format (this is a generator option).

src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
```diff
@@ -4,22 +4,36 @@ import static org.assertj.core.api.Assertions.assertThat;
...
 
 @UnitTest
 class SpringLiquibaseUtilTest {
 
+  private static String datasourceUrl;
+
+  @BeforeAll
+  public static void setup() throws URISyntaxException {
+    datasourceUrl = getTestProperties().getProperty("spring.datasource.url");
+    System.out.println(datasourceUrl);
```
murdos, line 31: Debug statement, should be removed

src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
```diff
@@ -4,22 +4,36 @@ import static org.assertj.core.api.Assertions.assertThat;
...
+  }
+
   @Test
   void createSpringLiquibaseFromLiquibaseDataSource() {
-    DataSource liquibaseDatasource = DataSourceBuilder.create().url("jdbc:h2:mem:liquibase").username("sa").build();
+    DataSource liquibaseDatasource = DataSourceBuilder.create().url(datasourceUrl).username("sa").build();
```
murdos, line 36: Hmm... Does that mean that the target database should already be started?
Or that TestContainers will automatically start a test database?

murdos: Honestly, I don't have a better solution, but I'm thinking that maybe keeping h2 is finally maybe not so problematic.
Another option is to move this `AsyncSpringLiquibase` feature to a separate module, since IMO it is not that much used.
In my experience having standard liquibase experience by default would fit for most people.
@pascalgrimaud : WDYT?

swarajsaaj: > Hmm... Does that mean that the target database should already be started?
Or that TestContainers will automatically start a test database?

Hi @murdos in my understanding the Datasource just checks for presence of the respective database's Driver class availability based on jdbc url, it doesn't really establish a connection with database in this unit test. And further validates that liquibase is able to instantiate itself.

src/main/resources/generator/server/springboot/dbmigration/liquibase/test/SpringLiquibaseUtilTest.java.mustache
murdos: https://github.com/jhipster/jhipster-lite/pull/9337 should help illustrate (and verify correct behavior) for that

swarajsaaj: @murdos Thanks for pointing this out, I missed the properties format. 
I have fixed it in latest commit [here](https://github.com/jhipster/jhipster-lite/pull/9332/files#diff-63e2539e5f9693e4f838b824ebbb0d0ba0dc37529ffd79e71ccacac5689dcbe3R222) that handles both formats.
I am not sure if there is a better way to handle it, because there is no conditional rendering supported in our mustache templates (only placeholder replacing is supported currently).
So added a conditional in the private helper method.

murdos (approved):

PR #9335, opened 2024-03-23
Factorization in JHipsterModulesFixture

murdos:
This is now possible since java build profiles are partially supported with gradle


renanfranca (approved):
Cool refactoring! Thanks 😎! I'll work on the gradle support on monday 👍

PR #9351, opened 2024-03-25
Add new module renovate

### LINKED ISSUE #9060: Module: renovate
It might be nice to have a module to configure [renovate](https://github.com/renovatebot/renovate) for a jhipster-lite generated project, with e.g. docker-compose dependencies update for content in `src/main/docker` (as done in #9059) and good default settings.

swarajsaaj:
Add renovate for generated projects for automatic dependency upgdates

Fix #9060

### DIFF
--- a/src/main/resources/generator/ci/renovate/renovate.json.mustache
+++ b/src/main/resources/generator/ci/renovate/renovate.json.mustache
@@ -1,0 +1,26 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":disableDependencyDashboard",
+    ":automergeMinor",
+    ":automergePr",
+    ":automergeRequireAllStatusChecks",
+    ":label(area: dependencies)"
+  ],
+  "prHourlyLimit": 20,
+  "enabledManagers": [
+    "maven",
+    "maven-wrapper",
+    "gradle",
+    "gradle-wrapper",
+    "docker-compose",
+    "dockerfile",
+    "npm"
+  ],
+  "docker-compose": {
+    "fileMatch": [
+      "^src/main/docker/[^/]*\\.ya?ml$"
+    ]
+  }
+}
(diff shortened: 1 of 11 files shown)


src/main/resources/generator/ci/renovate/renovate.json.mustache
```diff
@@ -0,0 +1,26 @@
...
+  "extends": [
+    "config:recommended",
+    ":disableDependencyDashboard",
+    ":automergeMinor",
+    ":automergePr",
+    ":automergeRequireAllStatusChecks",
+    ":label(area: dependencies)"
+  ],
+  "prHourlyLimit": 20,
+  "enabledManagers": [
```
murdos, line 12: I think we should rather avoid declaring explicitly managers, so they will be auto discovered.

src/test/java/tech/jhipster/lite/generator/ci/renovate/RenovateModuleFactoryTest.java
```diff
@@ -0,0 +1,47 @@
...
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.JHipsterModulesFixture;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+
+@UnitTest
+@ExtendWith(MockitoExtension.class)
+class RenovateModuleFactoryTest {
+
+  @InjectMocks
+  private RenovateModuleFactory factory;
```
murdos, line 23: There's no dependency here, so no need for Mockito, we can just create a new instance using the constructor: 
```suggestion
  private RenovateModuleFactory factory = new RenovateModuleFactory();
```

murdos (commented):
Could you also declare this module in the fullapp that is generated on CI? https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L173
That way the module generation will also be tested in real conditions.

src/test/java/tech/jhipster/lite/generator/ci/renovate/RenovateModuleFactoryTest.java
swarajsaaj: Nice catch. Thanks.

murdos:
Thanks @swarajsaaj :)

PR #9352, opened 2024-03-25
Add toString methods to a couple of objects

gzsombor:


src/main/java/tech/jhipster/lite/module/domain/landscape/JHipsterLandscapeModule.java
```diff
@@ -84,6 +85,16 @@ public boolean equals(Object obj) {
     return new EqualsBuilder().append(module, other.module).isEquals();
   }
 
+  @Override
+  public String toString() {
+    return new ToStringBuilder(this)
```
murdos, line 90: `SHORT_PREFIX_STYLE` is nicer, and is what we're already using in code when `ToStringBuilder` is used: 
```suggestion
    return new ToStringBuilder(this, SHORT_PREFIX_STYLE)
```

src/test/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResourcesTest.java
```diff
@@ -39,4 +39,11 @@ void shouldNotBuildWithDuplicatedSlug() {
       DuplicatedSlugException.class
     );
   }
+
+  @Test
+  void shouldHaveMeaningfullToString() {
+    var resource = defaultModuleResourceBuilder().slug("dummy").build();
+
+    assertThat(resource.toString()).contains("JHipsterModuleResource", "slug=", "dummy");
```
murdos, line 47: With SHORT_PREFIX_STYLE, you can then test the full toString expression:
```suggestion
    assertThat(resource).hasToString("...");
```

src/main/java/tech/jhipster/lite/module/domain/landscape/JHipsterLandscapeModule.java
gzsombor: Done

src/test/java/tech/jhipster/lite/module/domain/resource/JHipsterModulesResourcesTest.java
gzsombor: Even with the SHORT_PREFIX_STYLE, the result is too verbose, that I would rather not put the whole String in the test

murdos (approved):

PR #9353, opened 2024-03-25
Manage maven version similarly as the other dependency versions are m…

gzsombor:
…anaged

### DIFF
--- a/src/main/resources/generator/buildtool/maven/pom.xml.mustache
+++ b/src/main/resources/generator/buildtool/maven/pom.xml.mustache
@@ -18,7 +18,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <maven.version>3.9.5</maven.version>
+    <maven.version>{{mavenVersion}}</maven.version>
   </properties>
 
   <build>
--- a/src/main/resources/generator/buildtool/maven/pom.xml.mustache
+++ b/src/main/resources/generator/buildtool/maven/pom.xml.mustache
@@ -18,7 +18,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <maven.version>{{mavenVersion}}</maven.version>
+    <!-- Minimum supported Maven version -->
+    <maven.version>3.9.5</maven.version>
   </properties>
 
   <build>
--- a/src/main/resources/generator/dependencies/pom.xml
+++ b/src/main/resources/generator/dependencies/pom.xml
@@ -45,6 +45,7 @@
     <archunit-junit5.version>1.2.1</archunit-junit5.version>
     <cucumber.version>7.16.1</cucumber.version>
     <!-- Maven Plugins -->
+    <maven.version>3.9.6</maven.version>
     <os-maven-plugin.version>1.7.1</os-maven-plugin.version>
     <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
     <maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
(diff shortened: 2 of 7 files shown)


src/main/resources/generator/buildtool/maven/pom.xml.mustache
```diff
@@ -18,7 +18,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <maven.version>3.9.5</maven.version>
+    <maven.version>{{mavenVersion}}</maven.version>
```
DanielFran, line 21: @gzsombor @pascalgrimaud @murdos If maven.version is used to enforce minimum maven version with maven-enforcer-plugin (like in the generator-jhipster), this should be version 3.2.5.

See https://maven.apache.org/docs/history.html:
image

src/main/resources/generator/dependencies/pom.xml
```diff
@@ -45,6 +45,7 @@
     <archunit-junit5.version>1.2.1</archunit-junit5.version>
     <cucumber.version>7.16.1</cucumber.version>
     <!-- Maven Plugins -->
+    <maven.version>3.9.6</maven.version>
```
murdos, line 48: Maven wrapper version is already managed by renovate that automatically update it, so I don't really see the added value with theses changes:
image

src/main/resources/generator/buildtool/maven/pom.xml.mustache
murdos: @DanielFran I agree. We previously had a similar discussion: https://github.com/jhipster/jhipster-lite/pull/8933#discussion_r1493830805
I however think that requiring 3.2.5 is way too low. Recent (3.8+) maven versions contains a lot of improvements and it would be a shame to not raise the bar a little. 
If we require 3.2.5 version, then the enforce verification doesn't have much value.

gzsombor: Ok, I'm going to rename the mavenVersion to 'latestMavenVersion' or something like that, and keep 'enforcer-plugin' as is

src/main/resources/generator/dependencies/pom.xml
gzsombor: My main goal is to have a way to get the current maven/gradle version from a module - as I try to create a jhipster module to create [Mise configuration](https://mise.jdx.dev/), so I could have a minimal, declarative project setup.  
Any alternative idea, how and where can I get those values ? I would rather not duplicate that constant in that module.

murdos: The main issue is that by doing that you're breaking automatic update of maven wrapper.
The source of truth for maven wrapper version in jhipster-lite is `src/main/resources/generator/buildtool/maven/.mvn/wrapper/maven-wrapper.properties` since it's handled by our automatic dependencies updater (renovate).

If for your specific needs you need to get the maven wrapper version used by jhipster-lite, you could implement a reader, **in your own module**, that will extract the version from `src/main/resources/generator/buildtool/maven/.mvn/wrapper/maven-wrapper.properties`:
`mise/domain/MavenWrapperVersionReader.java`
```java
interface MavenWrapperVersionReader {
  Version readVersion();
}
```
implemented by
`mise/infractructure/secondary/FilesystemMavenWrapperVersionReader.java`
```java
@Component
class FilesystemMavenWrapperVersionReader implements MavenWrapperVersionReader {
  @Override
  public Version readVersion() {
  // load generator/buildtool/maven/.mvn/wrapper/maven-wrapper.properties content from classpath
  // using a regexp to extract maven version
  }
}

murdos (changes requested):
As stated in [this comment](https://github.com/jhipster/jhipster-lite/pull/9353/files#r1538680751) this is breaking automatic maven wrapper updates with renovate

murdos:
Closing this PR since it's outdated and requires more design discussions.

PR #9366, opened 2024-03-26
Keycloak: Make realmName and clientScopeName customizable

### LINKED ISSUE #8654: Keycloak: allow choosing realm name and client scope
Instead of using hard coding `jhipster` as realm name and client scope, it should be easy to choose the name we want. So nothing is related directly to jhipster and avoid manual renaming

gzsombor:
Fixes #8654


src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/oauth2/core/domain/OAuth2ModuleFactory.java
```diff
@@ -45,21 +48,28 @@ public OAuth2ModuleFactory(DockerImages dockerImages) {
   public JHipsterModule buildModule(JHipsterModuleProperties properties) {
     Assert.notNull("properties", properties);
 
+    var realmName = properties.getOrDefaultString(REALM_NAME, "jhipster_realm");
+    var clientScopeName = properties.getOrDefaultString(CLIENT_SCOPE_NAME, "jhipster_client");
```
pascalgrimaud, line 52: I think we should use the same values than the current existing ones (which are the same than in generator-jhipster): jhipster

gzsombor: Honestly, having jhipster as a default for every imaginable value is not the wisest pattern :grinning:

pascalgrimaud: Ahah, I think the same than you. But I'd prefer keeping jhipster as default because:
- avoid breaking change for existing app
- closed to generator-jhipster so it's a little bit compatible

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/oauth2/core/domain/OAuth2ModuleFactory.java
```diff
@@ -45,21 +50,28 @@ public OAuth2ModuleFactory(DockerImages dockerImages) {
...
+  private void appendKeycloak(JHipsterModuleBuilder builder, String realmName, String clientScopeName) {
     DockerImageVersion keycloakImage = dockerImages.get("quay.io/keycloak/keycloak");
 
-    builder.context().put("dockerKeycloakVersion", keycloakImage.version().get()).put("dockerKeycloakImage", keycloakImage.fullName());
+    builder
+      .context()
+      .put("dockerKeycloakVersion", keycloakImage.version().get())
+      .put("dockerKeycloakImage", keycloakImage.fullName())
+      .put("realmName", realmName)
+      .put("clientScopeName", clientScopeName);
```
pascalgrimaud, line 74: I think we need to add some rules for this, maybe? Avoid space, avoid strange char, etc?

pascalgrimaud:
@gzsombor : something I wanted to tell you -> I'm really happy to see to here, in this project and doing contributions. Thanks so much for helping this project! :heart:

gzsombor:
> @gzsombor : something I wanted to tell you -> I'm really happy to see to here, in this project and doing contributions. Thanks so much for helping this project! ❤️

That is really nice of you! I'm happy if I could help - and have enough time to do that :)

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/security/oauth2/core/domain/OAuth2ModuleFactory.java
gzsombor: Yes, that's a very valid point

src/main/java/tech/jhipster/lite/module/domain/JHipsterSlug.java
```diff
@@ -1,35 +1,20 @@
...
 
-  private static final Pattern SLUG_FORMAT = Pattern.compile("^[a-z0-9-]+$");
-
   private final String slug;
 
   protected JHipsterSlug(String slug) {
-    Assert.notBlank("slug", slug);
-
-    assertFormat(slug);
+    Assert.field("slug", slug).notBlank().urlSafeSingleWord(() -> new InvalidJHipsterSlugException(slug));
```
murdos, line 13: I'm not very fond of this refactoring, I think we loose some expressiveness and ease of understanding.
I'd rather keep each pattern constant hidden in each class even if it looks like duplicated, and don't introduce custom exception generated by `Assert.field(.., ..)`.

murdos (changes requested):

src/main/java/tech/jhipster/lite/module/domain/JHipsterSlug.java
gzsombor: I think, calling a function `urlSafeSingleWord()` with proper and javadoc is lightyear ahead more expressive than trying to decipher `Pattern.compile("^[a-z0-9-]+$")` or to check if two pattern are the same or not.
 We could get rid of these custom InvalidJHipsterModuleTagException/ UrlSafeSingleWordException/InvalidJHipsterSlugException and just use a plain old IllegalArgumentException, that would help reducing the unnecessary verbosity.

Issue #8627, opened 2024-01-14
Support java build profile in gradle


murdos:
The `AddJavaBuildProfile` command is not implemented in `GradleCommandHandler`

- [x] https://github.com/jhipster/jhipster-lite/pull/9163
- [x] https://github.com/jhipster/jhipster-lite/issues/8628
- [x] https://github.com/jhipster/jhipster-lite/pull/9369
- [ ] https://github.com/jhipster/jhipster-lite/pull/9437

renanfranca:
@murdos : I want to try to fix this issue, is it ok for you?

murdos:
@renanfranca : sure :-)

renanfranca:
@murdos : Please, let me know WDYT about the following solution proposal to implement the java build profile in gradle. This proposal only covers this part:

> [ ] Add build profile, with activation
 [ ] Add properties

I generated the `gradleapp` from `generate.sh` and implemented the changes over it published at this repository [gradleapp-profiles-sample](https://github.com/renanfranca/gradleapp-profiles-sample). Here is the PR with the changes [simulate local and linux profile module in gradle](https://github.com/renanfranca/gradleapp-profiles-sample/pull/1/files). **I ran the solution through sonar and it didn't point to any code smells.**

I am going to break down my suggestions here.

### First here is the build command

```java
return moduleBuilder(properties)
      .javaBuildProfiles()
        .addProfile("local")
          .activation("")
          .properties()
            .set(buildPropertyKey("spring.profiles.active"), buildPropertyValue("local"))
            .and()
          .and()
        .and()
        .addProfile("linux")
          .activation(
               """
               System.getProperty("os.name").lowercase().contains("linux")
               """)
          .properties()
            .set(buildPropertyKey("spring.profiles.active"), buildPropertyValue("linux"))
            .and()
          .and()
        .and()
```

### The gradle commands and the output:
```bash
./gradlew --info --rerun-tasks  ➝ spring.profiles.active = local

./gradlew -Pprofile=local --info --rerun-tasks  ➝ spring.profiles.active = local

./gradlew -Pprofile=linux --info --rerun-tasks  ➝ spring.profiles.active = linux
```

### The proposal code
Here is only the changed code from the build.gradle.kts (here is the PR at the gradleapp-profiles-sample [simulate local and linux profile module in gradle](https://github.com/renanfranca/gradleapp-profiles-sample/pull/1/files)):
```kts
tasks.build {
  dependsOn("processResources")
}

[...]

PR #9369, opened 2024-03-26
support dependencies and dependencies management in a build profile

renanfranca:
See
- https://github.com/jhipster/jhipster-lite/issues/8627


src/test/java/tech/jhipster/lite/module/domain/JHipsterModulesFixture.java
```diff
@@ -55,11 +55,6 @@ public static JHipsterModule fullModule() {
           .pluginManagement(mavenEnforcerPluginManagement())
           .plugin(mavenEnforcerPlugin())
           .and()
-        .javaDependencies()
```
murdos, line 58: You can now remove the `gradleSupportedModule()` and `gradleSupportedModuleBuilder` since we a feature parity between maven and gradle for build profile, in order to have only one `module()` method.

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
```diff
@@ -260,15 +297,14 @@ private void addPropertyTo(BuildProperty property, Path buildGradleFile) {
...
-        new MandatoryFileReplacer(
-          new JHipsterProjectFilePath(Path.of(projectFolder.folder()).relativize(buildGradleFile).toString()),
-          replacer
-        )
-      )
+      ContentReplacers.of(new MandatoryFileReplacer(new JHipsterProjectFilePath(projectFolderRelativePathFrom(buildGradleFile)), replacer))
     );
   }
 
+  private String projectFolderRelativePathFrom(Path buildGradleFile) {
```
murdos, line 304: How about returning directly a JHipsterProjectFilePath? You seem to always use this method that way
```suggestion
  private JHipsterProjectFilePath projectFolderRelativePathFrom(Path buildGradleFile) {
    return new JHipsterProjectFilePath(Path.of(projectFolder.folder()).relativize(buildGradleFile).toString());
```

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
```diff
@@ -330,6 +367,10 @@ private void enablePrecompiledScriptPlugins() {
     addFileToProject(from("buildtool/gradle/buildSrc/build.gradle.kts.template"), to("buildSrc/build.gradle.kts"));
   }
 
+  private void injectTomlVersionCatalogLibsIntoScriptPlugins() {
```
murdos, line 370: The method name is a bit misleading (you're rather initiating settings.gradle for precompiled script plugins), and I think you should rather add this `addFileToProject(..)` statement directly in `enablePrecompiledScriptPlugins()`

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
```diff
@@ -188,38 +192,70 @@ private static GradleDependencyScope gradleDependencyScope(JavaDependency depend
...
+      .stream()
+      .filter(dependencySlug -> dependencyExistsFrom(dependencySlug, command.buildProfile()))
+      .forEach(dependencySlug -> {
+        removeDependencyFromBuildGradle(dependencySlug, command.buildProfile());
+        versionsCatalog.removeLibrary(command.dependency());
+      });
+  }
+
+  @ExcludeFromGeneratedCodeCoverage(reason = "The exception handling is hard to test and an implementation detail")
+  private boolean dependencyExistsFrom(DependencySlug dependencySlug, Optional<BuildProfileId> buildProfile) {
```
murdos, line 206: It would be nicer to use and read with the following signature:
```suggestion
private Predicate<DependencySlug> dependencyExistsFrom(Optional<BuildProfileId> buildProfile) {
    return dependencySlug -> {
      try {
        return dependencyLinePattern(dependencySlug, buildProfile).matcher(Files.readString(buildGradleFile(buildProfile))).find();
      } catch (IOException e) {
        throw GeneratorException.technicalError("Error reading build gradle file: " + e.getMessage(), e);
      }
    };
  }
```

That way you could use it as `.filter(dependencyExistsFrom(command.buildProfile()))`

murdos (dismissed):

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
renanfranca: Sure! Nice 👍

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
renanfranca: Thanks! I have to do this more often 👍

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
renanfranca: Thanks! Yes, It is part of the PrecompiledScript prerequisites to work properly 😃

renanfranca:
@murdos : the code coverage wasnt 100% because the annotation `@ExcludeFromGeneratedCodeCoverage` doesn't work for the following method as expected. Did you encounter this situation before?

image

Thanks for the detailed review!

murdos:
@renanfranca : it works for the method, but not for the lambda that is inside the method.
There's currently no way  to tell jacoco to exclude a lambda from coverage calculation 😅

renanfranca:
> There's currently no way to tell jacoco to exclude a lambda from coverage calculation 😅

@murdos: I implemented this fix:

```java
  private Predicate<DependencySlug> dependencyExistsFrom(Optional<BuildProfileId> buildProfile) {
    return dependencySlug -> dependencyExistsFrom(dependencySlug, buildProfile);
  }

  @ExcludeFromGeneratedCodeCoverage(reason = "The exception handling is hard to test and an implementation detail")
  private boolean dependencyExistsFrom(DependencySlug dependencySlug, Optional<BuildProfileId> buildProfile) {
    try {
      return dependencyLinePattern(dependencySlug, buildProfile).matcher(Files.readString(buildGradleFile(buildProfile))).find();
    } catch (IOException e) {
      throw GeneratorException.technicalError("Error reading build gradle file: " + e.getMessage(), e);
    }
  }
```

murdos:
Great work, again @renanfranca ! Thanks!!!

renanfranca:
> Great work, again @renanfranca ! Thanks!!!

You are welcome! Thank you for the kind words 😄

PR #9378, opened 2024-03-27
enh(gradle): minor improvement in handling IOException and test coverage

murdos:


renanfranca (approved):
Thank you for the improvement, much better!

PR #9437, opened 2024-04-03
support gradle plugins in a build profile

### LINKED ISSUE #8627: Support java build profile in gradle

renanfranca:
Fix #8627


src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
```diff
@@ -114,13 +115,13 @@ public void handle(AddDirectJavaDependency command) {
     Assert.notNull(COMMAND, command);
 
     versionsCatalog.addLibrary(command.dependency());
-    addDependencyToBuildGradle(command.dependency(), command.buildProfile());
+    addDependencyToBuildGradle(command.dependency(), buildGradleFile(command.buildProfile()), command.buildProfile().isPresent());
   }
 
-  private void addDependencyToBuildGradle(JavaDependency dependency, Optional<BuildProfileId> buildProfile) {
+  private void addDependencyToBuildGradle(JavaDependency dependency, Path buildGradleFile, boolean forBuildProfile) {
```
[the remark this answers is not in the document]
renanfranca: @murdos : maybe because this profile plugin implementation is counterintuitive, I couldn't think of an elegant way to make the method `buildGradleFile` handle the `PLUGIN_BUILD_GRADLE_FILE = "buildSrc/build.gradle.kts";`

```java
  private Path buildGradleFile(Optional<BuildProfileId> buildProfile) {
    return buildProfile
      .map(buildProfileId -> {
        File scriptPlugin = scriptPluginForProfile(buildProfileId);
        if (!scriptPlugin.exists()) {
          throw new MissingGradleProfileException(buildProfileId);
        }
        return scriptPlugin.toPath();
      })
      .orElse(projectFolder.filePath(BUILD_GRADLE_FILE));
  }
  ```

File: src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java [hunk 2]
```diff
@@ -429,18 +430,31 @@ public void handle(AddGradlePlugin command) {
...
+      case GradleProfilePlugin plugin -> {
+        declarePlugin(
+          """
+          id("%s")\
+          """.formatted(plugin.dependency().get().id().groupId()),
+          command.buildProfile()
+        );
+        versionsCatalog.addLibrary(plugin.dependency().get());
+        addDependencyToBuildGradle(plugin.dependency().get(), projectFolder.filePath(PLUGIN_BUILD_GRADLE_FILE), false);
+      }
```
renanfranca, line 450: @murdos : should I delete `GradlePluginDependency` record and use `JavaDependency` directly at `GradleProfilePlugin`?

The code below will be cleaner:
```suggestion
      case GradleProfilePlugin plugin -> {
        declarePlugin(
          """
          id("%s")\
          """.formatted(plugin.dependency().id().groupId()),
          command.buildProfile()
        );
        versionsCatalog.addLibrary(plugin.dependency());
        addDependencyToBuildGradle(plugin.dependency(), projectFolder.filePath(PLUGIN_BUILD_GRADLE_FILE), false);
      }
```

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePlugin.java
```diff
@@ -2,8 +2,6 @@
 
 import java.util.Optional;
 
-public sealed interface GradlePlugin permits GradleCommunityPlugin, GradleCorePlugin {
-  GradlePluginId id();
-
```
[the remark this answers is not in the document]
murdos: I think it is still required. The plugin you picked as an example is a special case, because the id has the same value as the group.
The example used in [Gradle documentation](https://docs.gradle.org/current/userguide/implementing_gradle_plugins_precompiled.html#sec:applying_external_plugins) is nice, because id, group and artifactId (aka name) are different:
group is [com.bmuschko](https://github.com/bmuschko/gradle-docker-plugin/blob/e5abdc28e483905bacd49ae49738482dbd567a60/build.gradle.kts#L20), name is [gradle-docker-plugin](https://github.com/bmuschko/gradle-docker-plugin/blob/e5abdc28e483905bacd49ae49738482dbd567a60/settings.gradle.kts#L5), but there are 3 bundled plugins, in particular [com.bmuschko.docker-remote-api](https://github.com/bmuschko/gradle-docker-plugin/blob/e5abdc28e483905bacd49ae49738482dbd567a60/build.gradle.kts#L60)

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
murdos: IMO JavaDependency has too much information (scope, classifier, optional, exclusions,...), so maybe `GradlePluginDependency` should only contains:
```java
private final GroupId groupId;
private final ArtifactId artifactId;
private final Optional<VersionSlug> versionSlug;
```
Maybe, internally, GradleCommandHandler could convert a GradlePluginDependency to a JavaDependency, for factorization and internal API reuse, but at least the public API is clean.
WDYT?

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradleProfilePlugin.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public GradleProfilePlugin build() {
+      return new GradleProfilePlugin(this);
+    }
+  }
+
+  public interface GradleProfilePluginDependencyBuilder {
+    GradleProfilePluginOptionalBuilder dependency(GradlePluginDependency dependency);
+
+    default GradleProfilePluginOptionalBuilder dependency(String groupId, String artifactId) {
```
murdos, line 76: This is a fragile method API, because there's nothing preventing the inversion of the groupId and artifactId values when using the method (by confusion for example).
A better signature would be:
```suggestion
    default GradleProfilePluginOptionalBuilder dependency(GroupId groupId, ArtifactId artifactId) {
```

src/main/resources/generator/buildtool/gradle/buildSrc/build.gradle.kts.template
```diff
@@ -5,3 +5,10 @@ plugins {
 repositories {
   mavenCentral()
 }
+
+dependencies {
+  // jhipster-needle-gradle-implementation-dependencies
```
murdos, line 10: How about using a dedicated needle?
And I think the other ones won't be used, so maybe there's no point in adding them.

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePlugin.java
```diff
@@ -2,8 +2,6 @@
...
+public sealed interface GradlePlugin permits GradleCommunityPlugin, GradleCorePlugin, GradleProfilePlugin {
```
murdos, line 5: I think there's a design flaw here: you want to use `GradleCommunityPlugin`, `GradleCorePlugin` only for "main" build plugins, and `GradleProfilePlugin` only for plugins in build profiles.
But there's nothing that prevents mis-usage, and that guides you in using the correct gradle plugin.

This code compiles (in `JHipsterModulesFixture`), but I don't think it generates something that works (unless I'm wrong).
```java
    .gradlePlugins()
      .plugin(jacocoGradlePlugin())
      .plugin(checkstyleGradlePlugin())
      .plugin(gitPropertiesGradlePluginDependency())
      .and()
    .javaBuildProfiles()
      .addProfile(localBuildProfile())
        .activation(buildProfileActivation().activeByDefault(false))
        .gradlePlugins()
          .plugin(gitPropertiesGradlePluginDependency())
          .plugin(jacocoGradlePlugin())
          .and()
```

Also I think `GradleProfilePlugin` is really a `GradleCommunityProfilePlugin`, and `GradleCoreProfilePlugin` should probably exist, if for example you can to apply the (core) plugin `checkstyle` only in a dedicated profile.

I'm not sure what would be the best design here (I would have to code it myself to find the design), maybe having a 3 levels inheritance:
```java
  sealed interface GradlePlugin {}
  
  sealed interface GradleMainBuildPlugin extends GradlePlugin {}
  record GradleCorePlugin() implements GradleMainBuildPlugin {}
  record GradleCommunityPlugin() implements GradleMainBuildPlugin {}

  sealed interface GradleProfilePlugin extends GradlePlugin {}
  record GradleCoreProfilePlugin() implements GradleProfilePlugin {}
  record GradleCommunityProfilePlugin() implements GradleProfilePlugin {}
```

murdos (commented):
I did a first partial review, focused on API design.
I've not yet reviewed GradleCommandHandler, it'll wait until design is settled.

src/main/resources/generator/buildtool/gradle/buildSrc/build.gradle.kts.template
renanfranca: > And I think the other ones won't be used, so maybe there's no point in adding them.

I agree i am going to remove the other needles

> How about using a dedicated needle?

I think we can keep the needle name, since the dependency scope will be handle by the API and always be empty, so the implementation needle is going to be used. So it is going to use the already AddDependency implementation 👍

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePlugin.java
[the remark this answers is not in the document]
renanfranca: > I think it is still required. The plugin you picked as an example is a special case, because the id has the same value as the group.

Thank you for the heads up! I am going to fix that 😊

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradleProfilePlugin.java
renanfranca: Thanks,  I agreed. I'll do it 👍

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
renanfranca: I agree 👍👍👍

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePlugin.java
renanfranca: I didn't think about the core plugin in the profile, thanks! I am going to try this approach 👍

> I would have to code it myself to find the design

I know the feeling.  If I am trying to foresight the complete deseign before start coding,  I got stuck. That's why I love this project, I can easily use TDD here 😁

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradleCommunityPlugin.java
```diff
@@ -19,7 +19,6 @@ private GradleCommunityPlugin(GradleCommunityPluginBuilder builder) {
     pluginSlug = Optional.ofNullable(builder.pluginSlug);
   }
 
-  @Override
```
murdos, line 22: We should keep this one.
Same remark for `GradleCorePlugin`

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/JHipsterModuleGradlePlugins.java
```diff
@@ -27,10 +29,13 @@ public static JHipsterModuleGradlePluginBuilder builder(JHipsterModuleBuilder mo
...
     Assert.notNull("versions", versions);
 
-    return new JavaBuildCommands(plugins.stream().map(toCommands(versions)).toList());
+    return new JavaBuildCommands(plugins.stream().map(toCommands(versions, Optional.empty())).toList());
   }
 
-  private Function<GradlePlugin, JavaBuildCommand> toCommands(JavaDependenciesVersions versions) {
+  private Function<GradleMainBuildPlugin, JavaBuildCommand> toCommands(
+    JavaDependenciesVersions versions,
+    Optional<BuildProfileId> buildProfile
```
murdos, line 37: It seems that this parameter is not used

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/JHipsterModuleGradleProfilePlugins.java
```diff
@@ -0,0 +1,101 @@
+package tech.jhipster.lite.module.domain.gradleplugin;
```
murdos, line 1: I'd rather see this one in `javabuildprofile` package:
```suggestion
package tech.jhipster.lite.module.domain.javabuildprofile;
```

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java
```diff
@@ -464,7 +478,21 @@ private void addPluginConfiguration(GradlePluginConfiguration pluginConfiguratio
...
+      .versionSlug()
+      .map(
+        versionSlug ->
+          JavaDependency.builder()
+            .groupId(plugin.dependency().groupId())
+            .artifactId(plugin.dependency().artifactId())
+            .versionSlug(versionSlug)
+            .build()
+      )
+      .orElse(JavaDependency.builder().groupId(plugin.dependency().groupId()).artifactId(plugin.dependency().artifactId()).build());
```
murdos, line 496: There's some redundancy here. You could factorize by extracting the builder to a variable:
```
var builder = JavaDependency.builder().groupId(plugin.dependency().groupId()).artifactId(plugin.dependency().artifactId());
plugin.versionSlug().ifPresent(builder::versionSlug);
return builder.build();
```

src/test/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandlerTest.java
```diff
@@ -1072,4 +1176,12 @@ private static void assertFileExists(JHipsterProjectFolder projectFolder, String
     Path profileGradlePath = Paths.get(projectFolder.get()).resolve(other);
     assertThat(Files.exists(profileGradlePath)).as(description, profileGradlePath.toString()).isTrue();
   }
+
+  private static GradleProfilePlugin dockerGradlePluginDependency() {
```
murdos, line 1180: It might be interesting to move it to JHipsterModulesFixture

murdos (dismissed):

renanfranca:
Hello,
I keep receiving the following error at `codecov` pipeline job. Would someone be able to help me? Is it wait and try again later?
image
https://github.com/jhipster/jhipster-lite/actions/runs/8599285157/job/23564724606

murdos:
@renanfranca : it's not specific to your branch, this is a generalized issue with the codecov action: https://github.com/codecov/codecov-action/issues/1359

renanfranca:
> codecov action: codecov/codecov-action#1359

Thanks!

renanfranca:
@murdos : please, take a look at [this](https://github.com/dunst-project/dunst/pull/1312#issuecomment-2043042272):

> I'll just update to v4 ... which should work as we're using fedora based images for the coverage reports (IIRC v4 had issues on older ubuntu based images).

He [fixed](https://github.com/dunst-project/dunst/pull/1326) like you did and it seems to working as you can see at the [latest actions](https://github.com/dunst-project/dunst/actions/workflows/main.yml)

renanfranca:
Another possibility of fix (https://github.com/pybop-team/PyBOP/issues/276#issuecomment-2042639656):

> The root problem with this bug was a typo in the secret. This cause the token to not be used by the uploaded, defaulting to non-token uploading. This has now fixed.

PR #9513, opened 2024-04-11
gradle ability to declare resources filtering in a generic way

### LINKED ISSUE #9405: Gradle: ability to declare resources filtering in a generic way
Needed to support gradle in the `spring-boot-local-profile` module.

With maven, this is indirectly declared by declaring and configuring the `maven-resources-plugin`.
But with gradle it can't be done by declaring a plugin.
Maybe for gradle we should support declaring a free configuration block. Not perfect, if multiple modules need to declare resources filtering, but it might be enough in the first instance

renanfranca:
Fix #9405


renanfranca:
After the review and merge here, I can implement the gradle `spring-boot-local-profile` support in a new PR, WDYT?

murdos:
LGTM :-)

renanfranca:
> LGTM :-)

Awesome!!! Thanks 😁 😊 🫂

PR #9551, opened 2024-04-15
support gradle at spring-boot-local-profile module

renanfranca:
See https://github.com/jhipster/jhipster-lite/issues/9405#issuecomment-2028854888

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/localeprofile/domain/LocalProfileModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/localeprofile/domain/LocalProfileModuleFactory.java
@@ -1,10 +1,11 @@
 package tech.jhipster.lite.generator.server.springboot.localeprofile.domain;
 
 import static tech.jhipster.lite.module.domain.JHipsterModule.*;
-import static tech.jhipster.lite.module.domain.replacement.ReplacementCondition.notContainingReplacement;
+import static tech.jhipster.lite.module.domain.replacement.ReplacementCondition.*;
 
 import tech.jhipster.lite.module.domain.JHipsterModule;
 import tech.jhipster.lite.module.domain.buildproperties.PropertyKey;
+import tech.jhipster.lite.module.domain.buildproperties.PropertyValue;
 import tech.jhipster.lite.module.domain.mavenplugin.MavenPlugin;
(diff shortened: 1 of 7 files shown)


src/main/java/tech/jhipster/lite/generator/server/springboot/localeprofile/domain/LocalProfileModuleFactory.java
```diff
@@ -19,6 +20,26 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
 
     //@formatter:off
     return moduleBuilder(properties)
+      .javaBuildProperties()
+        .set(new PropertyKey("spring.profiles.active"), new PropertyValue("-"))
```
renanfranca, line 24: @murdos : that's my workaround for the fact that PropertyValue can't be blank. WDYT?

renanfranca:
This PR is marked as a draft because I didn't generate a grade application with this module and test it yet.

src/main/java/tech/jhipster/lite/generator/server/springboot/localeprofile/domain/LocalProfileModuleFactory.java
murdos: I think we should rather relax the assertions on PropertyValue : it can be blank, we have a valid use-case here 🙂

murdos (approved):

src/main/java/tech/jhipster/lite/generator/server/springboot/localeprofile/domain/LocalProfileModuleFactory.java
```diff
@@ -13,12 +14,34 @@
...
+          }
+
+          tasks.processResources {
+            filesMatching("**/*.yml") {
+              filter { it.replace("@spring.profiles.active@", springProfilesActive) }
+            }
+            filesMatching("**/*.properties") {
+              filter { it.replace("@spring.profiles.active@", springProfilesActive) }
+            }
+          }
```
renanfranca, line 41: @murdos : FYI, I had to change the filesMatching because It only accepts one pattern parameter, and it's using Ant-Style Pattern and not Regex: `@param pattern Ant-style pattern used to match against files' relative paths`. I couldn't manage to cover both files with one pattern.

I added `spring-boot-local-profile` module to CI at gradleapp 👍

PR #9577, opened 2024-04-18
Renovate: group @typescript-eslint dependencies

qmonmert:


murdos:
These updates are already grouped, aren't they?
https://github.com/jhipster/jhipster-lite/pull/9582

qmonmert:
@murdos indeed :)

PR #9596, opened 2024-04-21
Migration from gradle enterprise to develocity

murdos:
gradle-enterprise-maven-extension is now deprecated in favor of develocity-maven-extension


atomfrede (approved):
Finally, I always wondered when the relocation will be done for the artifacts. Thanks @murdos

PR #9608, opened 2024-04-22
Format Cypress tests

qmonmert:


murdos:
FYI this is already handled automatically with https://github.com/jhipster/jhipster-lite/pull/9592/commits/74f4ab96f908618c9a7a932bbd2b36bf40bdb4ee in https://github.com/jhipster/jhipster-lite/pull/9592

PR #9630, opened 2024-04-24
Generate a package.json requiring only the recent major node.js version, not the latest patch version

murdos:
Following #9407, applying the same logic on generated application


renanfranca (approved):
Thanks! All good for me 😁👍

PR #9672, opened 2024-04-28
Dependabot: ignore Node 22 as not LTS

pascalgrimaud:


murdos:
This is because 22 will only be LTS in 6 months?

pascalgrimaud:
No, it's to ignore PR like this: https://github.com/jhipster/jhipster-lite/pull/9641

As soon as Node 22 is becoming LTS, we'll need to update this config

PR #9673, opened 2024-04-28
(develocity): always publish build scans, even in case of build failures

murdos:


murdos:
@atomfrede : we have an issue with uploading of build scans (and also build cache), as you can see from the lack of recent builds (excepted mine) on https://ge.jhipster.tech/
According to the logs, develocity considers we're not authenticated: 
```
[DEBUG] Using the build cache with the following configuration:
  Local build cache: disabled:
      Local build cache was not enabled in the configuration of the build.
  Remote build cache: enabled
      url: https://ge.jhipster.tech/cache/
      authentication: none
      storeEnabled: true
      allowUntrustedServer: false
      allowInsecureProtocol: false
      useExpectContinue: false
```

This is probably the cause of this error:
>[INFO] --- compiler:3.13.0:compile (default-compile) @ jhlite ---
[DEBUG] Remote cache miss
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 838 source files with javac [debug parameters release 21] to target/classes
Warning:  Could not store entry e814bdd9382157f18633b5b0643d304a in remote build cache: Storing entry at 'https://ge.jhipster.tech/cache/e814bdd9382157f18633b5b0643d304a' response status 403: Forbidden

I've tried many things, but I can't find where the error comes from. 
It looks like develocity doesn't find the DEVELOCITY_ACCESS_KEY environnement variable.
Do you have any idea?

atomfrede:
I will have a look. The token should not be expired. Maybe I did a copy and paste error. Will check tomorrow

atomfrede:
@murdos At first glance everything looks fine. The configured access token has been used today, so I suppose at least the configuration is fine.

.mvn/develocity.xml
```diff
@@ -11,7 +11,7 @@
       <fileFingerprints>true</fileFingerprints>
     </capture>
     <publishing>
-      <onlyIf><![CDATA[!buildResult.failures.empty && authenticated]]></onlyIf>
+      <onlyIf>true</onlyIf>
```
atomfrede, line 14: This makes totally sense. Couldn't publish build scans from my local machine during testing.

atomfrede:
@murdos  I have updated the secret, to make sure the content is correct. Should we merge this?

murdos:
> @murdos I have updated the secret, to make sure the content is correct. Should we merge this?

Yes, I think we should merge it anyway, and see how it behaves on main branch builds :)

PR #9730, opened 2024-05-04
Disable develocity remote build cache when building docker images

murdos:
Workaround for recent systematic build failures on some branches


atomfrede (approved):
Makes sense to disable it 👍

PR #9748, opened 2024-05-06
Move the ability to start liquibase asynchronously to its own module

murdos:
The async liquibase feature is not often used in my experience, so I think it could be useful to move it to its own module so we can pick exactly what we need


pascalgrimaud:
@murdos : is it ready? can it be merged?

murdos:
@pascalgrimaud Yes, it's ready

PR #9758, opened 2024-05-07
support context variables at PackageJson and Content Replacer API

### LINKED ISSUE #9744: Replace hard coded `target/` path in `package.json` using `projectBuildDirectory` context value
The values at `package.json` are added/modified/removed without using the `.mustache` template, so it has to use the `projectBuildDirectory` context values in the implementation. In addition, Content Replacer should consider context values because the `target/` is used to add information inside the `package.json`.

Here are the places to change
- Angular Module ([1](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java#L61) and [2](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java#L116))
- [ThymeleafTemplate Module](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/server/springboot/thymeleaf/template/domain/ThymeleafTemplateModuleFactory.java)

renanfranca:
Fix #9744


src/main/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemPackageJsonHandler.java
```diff
@@ -74,6 +73,28 @@ private Path getPackageJsonFile(JHipsterProjectFolder projectFolder) {
...
+          StringBuilder sb = new StringBuilder();
+          while (matcher.find()) {
+            matcher.appendReplacement(sb, Matcher.quoteReplacement(value.toString()));
+          }
+          matcher.appendTail(sb);
+
+          result.setLength(0);
+          result.append(sb);
+        }));
+    return result.toString();
```
renanfranca, line 95: @murdos : I think I could replace this method implementation using the method `tech.jhipster.lite.module.infrastructure.secondary.file.MustacheTemplateRenderer.render`? WDYT?

Thanks in advance ;)

src/main/java/tech/jhipster/lite/module/domain/JHipsterModulesApplyer.java
```diff
@@ -109,6 +110,14 @@ public JHipsterModuleApplied apply(JHipsterModuleToApply moduleToApply) {
     return moduleApplied;
   }
 
+  private JHipsterModulePackageJson buildPackageJson(JHipsterModule module, Optional<JavaBuildTool> detectedJavaBuildTool) {
+    JHipsterModuleContext context = detectedJavaBuildTool
```
murdos, line 114: I think this would be clearer if we extract to a method:
```java
 private Optional<JavaBuildTool> detectedJavaBuildTool(JHipsterModule module) {
    return javaBuildTools
      .detect(module.projectFolder())
      .or(() -> javaBuildTools.detect(module.files()));
}
```

And then use it here and in `buildStartupCommands(..)` (by changing its signature to `buildStartupCommands(JHipsterModule)`), and remove it from the `Optional<JavaBuildTool>` as an input parameter:
```suggestion
    JHipsterModuleContext context = detectedJavaBuildTool(module)
```

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
```diff
@@ -61,6 +65,16 @@ public PackageJsonType type() {
     return type;
   }
 
+  public Optional<JHipsterModuleContext> context() {
+    return context;
+  }
+
+  public JHipsterModulePackageJson withContext(JHipsterModuleContext context) {
+    this.context = Optional.ofNullable(context);
```
murdos, line 73: This breaks the immutability of `JHipsterModulePackageJson`. 
Field `context` should be final, and this method `withContext()` should return a new instance of `JHipsterModulePackageJson` (by internally reusing the builder).

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemPackageJsonHandler.java
murdos: Definitely :) 
However you should inject and manipulate the interface (`TemplateRenderer`) rather than the implementation.

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
renanfranca: Thanks for the warning! I made the change, when you have the time, please let me know WDYT 😉🖖

renanfranca:
I still in draft because I am going to update the modules in this pull request too 🖖

renanfranca:
I identified the sonar bug, and I will fix it later 👍

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
murdos: I've played a bit more with your branch to find a cleaner solution, and ended with changes that I pushed directly to your branch: https://github.com/jhipster/jhipster-lite/pull/9758/commits/7ca0fd71988d4d1a29aa7d5d0c669778bc4c88d4
Please let me know what you think :)

src/test/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemPackageJsonHandlerTest.java
```diff
@@ -426,10 +457,22 @@ private void mockVersion() {
     }
   }
 
+  @SafeVarargs
+  private @NotNull PackageJsonChanges packageJsonChanges(Consumer<JHipsterModulePackageJsonBuilder>... builderConfigurations) {
+    JHipsterModulePackageJsonBuilder builder = emptyBuilder();
+    Stream.of(builderConfigurations).forEach(configuration -> configuration.accept(builder));
+
+    return builder.build().buildChanges(emptyModuleContext());
+  }
```
renanfranca, line 466: I like it, very sophisticate!

src/main/java/tech/jhipster/lite/module/domain/packagejson/PackageJsonChanges.java
```diff
@@ -0,0 +1,66 @@
...
+    Assert.notNull("context", context);
+
+    this.scripts = packageJson.scripts();
+    this.dependencies = packageJson.dependencies();
+    this.dependenciesToRemove = packageJson.dependenciesToRemove();
+    this.devDependencies = packageJson.devDependencies();
+    this.devDependenciesToRemove = packageJson.devDependenciesToRemove();
+    this.type = packageJson.type();
+    this.context = context;
+  }
```
renanfranca, line 27: I like it, no need for the builder. I tried to create a JHipsterModuleChangesPackageJson and it got so complicated that I gave up!

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
```diff
@@ -17,7 +15,6 @@ public final class JHipsterModulePackageJson {
   private final PackageJsonDependencies devDependencies;
   private final PackageJsonDependencies devDependenciesToRemove;
   private final PackageJsonType type;
-  private final Optional<JHipsterModuleContext> context;
```
renanfranca: Nice move! I was uncomfortable with context here because it was available to the public API and the user doesn't need to fill it all.

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
renanfranca: I enjoy it! Now it's fell right now. I can move on! I comment on code to recognize your work and share my thoughts. Thanks 👍

src/main/java/tech/jhipster/lite/module/domain/packagejson/JHipsterModulePackageJson.java
murdos: Yes, this is exactly what pushed me to introduces the new class :)

renanfranca:
I encounter a new issue when [trying to replace this `target` for `{{projectBuildDirectory}}` at AngularModuleFactory](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java#L116) and it didn't work because [it is inserted with replacer. After all, pckageJson API does not support `jestSonar` block.](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java#L101-L104)

I see two possibilities:
1. add support for jestSonar block
2. add support for replacer handle context variable and update it.

I am going to try to implement the second option because this jestSonar block is only needed at AngularModuleFactory and is not generic enough to deserve an implementation at packageJson API.

renanfranca:
@murdos : I implemented `add support for replacer handle context variable and update it`. Depending on your review I could refactor the packageJson API. Thanks in advance and take your time 😉🖖

src/main/java/tech/jhipster/lite/module/domain/packagejson/PackageJsonData.java
```diff
@@ -0,0 +1,10 @@
+package tech.jhipster.lite.module.domain.packagejson;
+
+record PackageJsonData(
```
murdos, line 3: I'm really not found of this (and is the reason why I've not approved this PR earlier). It brings some unneeded coupling. 
If Sonar is not happy about duplication for these pieces of code, we can disable the rule locally with a `@SuppressWarning`.

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemPackageJsonHandler.java
```diff
@@ -74,6 +81,10 @@ private Path getPackageJsonFile(JHipsterProjectFolder projectFolder) {
     return file;
   }
 
+  private String replacePlaceholders(JHipsterModuleContext context, String content) {
```
murdos, line 84: If find it more logical that way, but this is probably subjective :)
```suggestion
  private String replacePlaceholders(String content, JHipsterModuleContext context) {
```

BTW: this is the signature you used in `FileSystemReplacer`

src/main/java/tech/jhipster/lite/module/domain/packagejson/PackageJsonData.java
renanfranca: > I'm really not found of this (and is the reason why I've not approved this PR earlier). It brings some unneeded coupling.

I can't agree more! It was only to make sonar let me go.

> If Sonar is not happy about duplication for these pieces of code, we can disable the rule locally with a @SuppressWarning.

That's great news!!! I am going to revert this change and add the supresswarning!!!

src/main/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemPackageJsonHandler.java
renanfranca: I agree. I copied the method from ...PackageJson...handler and felt strange. It is better to make the method signature logic inside the class context since it is a private method.

renanfranca: > I agree. I copied the method from ...PackageJson...handler and felt strange. It is better to make the method signature logic inside the class context since it is a private method.

Just to share my logic was to follow the other methods that put content in the last parameter:

```java
    String content = readContent(file);
    content = replaceType(indentation, packageJson.type(), content);
    content = replaceScripts(indentation, packageJson.scripts(), content);
    content = replaceDevDependencies(indentation, packageJson.devDependencies(), content);
    content = replaceDependencies(indentation, packageJson.dependencies(), content);
    content = removeDependencies(indentation, packageJson.dependenciesToRemove(), content);
    content = removeDevDependencies(indentation, packageJson.devDependenciesToRemove(), content);

    content = replacePlaceholders(packageJson.context(), content);
    content = cleanupLineBreaks(indentation, content);
    ```

I will change it, no problem 👍

src/main/java/tech/jhipster/lite/module/domain/packagejson/PackageJsonData.java
renanfranca: I remove the need for `PackageJsonChanges.java` using the context from `JHipsterModuleChanges`. This will also address the code duplication issue identified by Sonar as a secondary benefit.

BTW: I didn't find a way to ignore sonar code duplication using `@SuppressWarning`. I am curious if I didn't search enough, I would like to know if you know how 😃

murdos (approved):

PR #9766, opened 2024-05-08
Run frontend tests by default in watch mode

murdos:
While developing it's much easier and idiomatic to just type `npm test` to get tests running continuously than `npm run test:watch`. The other way to run all tests is `npm run test:coverage`, that will also compute (and check if configured) code coverage.


renanfranca:
I agree, and it aligns with the developer experience proposal. But I think the maven frontend server plugin module should be changed to use the `npm run test:coverage` to keep it work as intending.

 I didn't hate the time to try run the `watch` to see if it will pause the .\mvnw verify with frontend server plugin

pascalgrimaud:
I'll have a look on this

murdos:
> But I think the maven frontend server plugin module should be changed to use the `npm run test:coverage` to keep it work as intending.

I'd already changed the maven frontend plugin configuration, it should work as expected :)

renanfranca:
> I'd already changed the maven frontend plugin configuration

I am sorry, I missed that change. I saw no. Everything will work just fine 🙂

qmonmert:
I don't really agree with that, on most projects `npm t` just execute tests and not watch them

murdos:
> I don't really agree with that, on most projects `npm t` just execute tests and not watch them

In order to choose the default behavior, I think we should determine what is the use case associated with manually running tests. In my opinion, you need to run tests when you have changed some code (either test code or production code), and using watch mode frees you from manually running tests after each code change. I personally don't have other use case for running tests locally, the overall test verification being done by the CI.

And I think that's why the default behavior and one of the greatest feature of `vitest` is to run in watch mode by default. `vitest` that we use for all our client code (excepted angular that is not supported by vitest).

Last point, I think with jhipster-lite we try to raise the bar not to follow the current use industry use. Otherwise we would not use hexagonal architecture, requires 100% code coverage, ...

pascalgrimaud:
I share and understand what you suggest @murdos
After thinking about it, it can be a small breaking change for users who use `npm test` in CI

As the default `npm test` is not in watch mode for years, what do you think about adding a new module which will replace the existing npm test by watch mode ?

The module could be named `front-test-watch-mode`, or something like that:
So everybody will be happy with that

Gnuk:
I like the idea to watch by default because the feedback is auto and it's better to encourage people to use TDD approach.

I don't like the idea of making another module just for the watch option.

So, I like the idea of this PR.

Gnuk (approved):
I like the global idea but I don't know if the name `test:coverage` is good because this script is used with `run` and it's also technically possible to combine `--watch` (or default mode) and `--coverage`.

pascalgrimaud:
Ok so I'm fine to have the watch mode by default, but I'd prefer to have these **3 entries** in `package.json` :
- test: launch test:watch -> so with this, it will be easier to switch to test:coverage by default for user who prefer
- test:watch
- test:coverage

Another thing: we need to have the same in JHLite to be consistent, so something to do in this PR or in the next one

Are you ok @murdos ?

murdos:
> Ok so I'm fine to have the watch mode by default, but I'd prefer to have these **3 entries** in `package.json` :

:+1: Seems reasonable, thanks! :)
 
> Another thing: we need to have the same in JHLite to be consistent, so something to do in this PR or in the next one

Good idea, I'll take care of it too.

pascalgrimaud (approved):

PR #9778, opened 2024-05-09
Use the official Kafka image with TestContainers

murdos:
Latest TestContainers version now supports it


renanfranca (dismissed):
Thank you for your work :)

PR #9779, opened 2024-05-09
Sonar: declare properties-maven-plugin configuration in the pluginManagement section

murdos:
And define a default execution id
This allows running sonar analysis directly with the command `mvn properties:read-project-properties sonar:sonar`


renanfranca (approved):
Thanks! I didn't try, but I imagine that I can run `./mvnw clean verify` and after tests were ok and code coverage 100%. I am going to be able to run sonar analysis without the need to run the tests again... cool for developer flow 😎

PR #9792, opened 2024-05-12
feat(maven): replace existing plugin rather than adding a duplicated one

murdos:


src/test/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/maven/MavenCommandHandlerTest.java
```diff
@@ -828,6 +828,88 @@ void shouldAddBuildPluginManagementToPomWithPluginManagementPlugins() {
...
+                        </dependency>
+                        <dependency>
+                          <groupId>org.asciidoctor</groupId>
+                          <artifactId>asciidoctorj-diagram</artifactId>
+                        </dependency>
+                      </dependencies>
+                    </plugin>\
+            """
+          )
+          .containsOnlyOnce(
```
renanfranca, line 902: Nice! I never used it before.

PR #9801, opened 2024-05-13
add imports option at gradle plugin and gradle profile plugin API

### LINKED ISSUE #9788: Implement the option to add imports using the GradlePlugin API in the build.gradle.kts
The GradlePlugin API change:

```java
      .id("com.github.node-gradle.node")
      .pluginSlug("node-gradle")
      .versionSlug("node-gradle")
      .gradleImports()
        .import("com.github.gradle.node.npm.task.NpmTask")
        .and()
      .configuration(
        """
        node {
          version.set("v20.12.2")
          npmVersion.set("10.5.2")
          npmWorkDir.set(file("build"))
        }
        
        val buildTaskUsingNpm = tasks.register<NpmTask>("buildNpm") {
          description = "Build the frontend project using NPM"
          group = "Build"
          dependsOn("npmInstall")
          npmCommand.set(listOf("run", "build"))
          environment.set(mapOf("APP_VERSION" to project.version.toString()))
        }
        
        val testTaskUsingNpm = tasks.register<NpmTask>("testNpm") {
          description = "Test the frontend project using NPM"
          group = "verification"
          dependsOn("npmInstall", "buildNpm")
          npmCommand.set(listOf("run", "test"))
          ignoreExitValue.set(false)
          workingDir.set(projectDir)
          execOverrides {
            standardOutput = System.out
          }
        }
        
        tasks.bootJar {
          dependsOn("buildNpm")
          from("build/classes/static") {
              into("BOOT-INF/classes/static") // This will place the content under 'static' directory in the JAR
          }
        }
        """
      )
      .build();
```

**note:** the configuration is an example, wont be hard coded the versions and etc... the focus in this issue is the `gradleImports()`.

After enable the GradlePlugin API to handle imports, [here is the result](https://github.com/renanfranca/sample-frontend-gradle/pull/2/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R1-R2) in the `build.gradle.kts`.

renanfranca:
Fix #9788


src/test/java/tech/jhipster/lite/module/domain/JHipsterModulesFixture.java
```diff
@@ -482,6 +536,9 @@ public static GradleProfilePlugin dockerGradlePluginDependency() {
       .id("com.bmuschko.docker-remote-api")
       .dependency(groupId("com.bmuschko"), artifactId("gradle-docker-plugin"))
       .versionSlug("docker-plugin")
+      .gradleImports()
```
murdos, line 539: Let's simplify this, without `.gradleImports()` and just 
```suggestion
     .withBuildGradleImport("java.util.Properties")
     .withBuildGradleImport("java.util.foo")
```

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePluginImport.java
```diff
@@ -0,0 +1,13 @@
+package tech.jhipster.lite.module.domain.gradleplugin;
+
+import tech.jhipster.lite.shared.error.domain.Assert;
+
+public record GradlePluginImport(String pluginImport) {
```
murdos, line 5: How about this, since in reality the usage is not restricted to plugins:
```suggestion
public record BuildGradleImport(String import) {
```

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePluginImports.java
```diff
@@ -0,0 +1,54 @@
+package tech.jhipster.lite.module.domain.gradleplugin;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.stream.Stream;
+import tech.jhipster.lite.shared.error.domain.Assert;
+
+public final class GradlePluginImports {
```
murdos, line 9: I think we can remove the builder and just have:
```suggestion
public record GradlePluginImports(Collection<GradlePluginImport> imports) {

  public GradlePluginImports {
    Assert.field("imports", imports).notNull().noNullElement();
  }

  public void forEach(Consumer<GradlePluginImport> consumer) {
    Assert.notNull("consumer", consumer);

    imports.forEach(consumer);
  }

}
```

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradlePlugin.java
```diff
@@ -4,5 +4,6 @@
 
 public sealed interface GradlePlugin permits GradleMainBuildPlugin, GradleProfilePlugin {
   GradlePluginId id();
+  Optional<GradlePluginImports> imports();
```
murdos, line 7: Everywhere you used `Optional<GradlePluginImports>` can IMO be replaced by `GradlePluginImports`: since `GradlePluginImports` is in fact a collection, it can be empty, there's no need to wrap it in an Optional.

src/main/java/tech/jhipster/lite/module/domain/gradleplugin/GradleCommunityPlugin.java
```diff
@@ -44,6 +52,7 @@ public static GradleCommunityPluginIdBuilder builder() {
   private static final class GradleCommunityPluginBuilder implements GradleCommunityPluginIdBuilder, GradleCommunityPluginOptionalBuilder {
 
     private GradlePluginId id;
+    private final GradlePluginImportsBuilder<GradleCommunityPluginOptionalBuilder> imports = GradlePluginImports.builder(this);
```
murdos, line 55: Here you can replace that with:
```suggestion
    private final Collection<GradlePluginImport> imports = new ArrayList<>();
```

and add methods
```java
    @Override
    public GradleCommunityPluginOptionalBuilder withBuildGradleImport(GradlePluginImport import) {
      imports.add(import);
      return this;
    }

    @Override
    public GradleCommunityPluginOptionalBuilder withBuildGradleImport(String import) {
      return withBuildGradleImport(new GradlePluginImport(import));
    }
```

renanfranca: I have to use `gradleImport` as the parameter name because `import` got a compilation error using a reserved keyword.

murdos: Good point! :sweat_smile:

murdos (approved):
Thanks for your contribution! That's once again a great work :-)

renanfranca:
> Thanks for your contribution! That's once again a great work :-)

Thanks for acknowledging my work! It's especially meaningful coming from a top-notch programmer like you 😃✌️

Issue #9491, opened 2024-04-09
Alternative page for Landscape


pascalgrimaud:
Landscape is designed for a modular approach

After some discussion with @jdubois  and @murdos too, we think it could be interesting to have a page for beginners. The page could simply look like the one here: https://start.jhipster.tech/generate-application

It will probably help the users to try JHLite quicker

What do you think?

TO-DO:

- [x] https://github.com/jhipster/jhipster-lite/issues/10379
- [x] https://github.com/jhipster/jhipster-lite/issues/10317
- [x] https://github.com/jhipster/jhipster-lite/issues/10239

murdos:
I was thinking more about this idea this morning, and if the Landscape page could be loaded with a list of preselected modules (and configured properties), we could use preset configuration, and consider this new page as preset builder.

swarajsaaj:
@murdos the presets concept is similar to https://github.com/jhipster/jhipster-lite/issues/2907?

murdos:
@swarajsaaj : yes, there's some overlap. Template and preset are the same idea. 
Replay is another use case, but it could be solved by transforming an history file to a template/preset.

renanfranca:
I assigned myself to this issue because I would like to have a full-stack experience in jhipster-lite by implementing a new feature 😄

renanfranca:
Here is a demo of using the preset configurations 😃💙:


https://github.com/user-attachments/assets/8ef6389f-a363-42a4-b78d-88e9bf0f5389

renanfranca:
A demo video of the merged preset configuration feature 💙😃

https://github.com/user-attachments/assets/261ad08f-1439-49fd-b967-8827d9564691

atomfrede:
Pretty cool! Nice work!

renanfranca:
> Pretty cool! Nice work!

Thank you! I am glad you like it 😊

pascalgrimaud:
Can it be closed @renanfranca ?

PR #9833, opened 2024-05-18
Restructuring of generator frontend tests

murdos:
- Rename webapp tests folder from src/test/javascript to src/test/webapp : 
  - So src/main/webapp matches src/test/webapp
  - Also frontend tests are not limited to javascript code, this also includes components and e2e tests
- Rename webapp unit tests folder to src/test/webapp/unit
- webapp tests: isolate component tests from e2e tests
- rename e2e folder and remove component tests from e2e folder


pascalgrimaud:
Totally agree with this change, although it's a breaking change
Just wonder if `src/test/webapp` could / should be a variable ? in case we have a full typescript application

murdos:
@pascalgrimaud : I'm glad you agree :) I'll try to fix failing test on windows so we can merge it.
The current PR is only for the generator, I wanted to setup things cleanly before trying something for #9491
It's also a prerequisite for something I want to investigate: merging code coverage between unit tests and components tests for frontend (that would allow removing some dumb unit tests).
If things are satisfying, we can then evolve generated apps :)

murdos:
FYI: there's an issue with tikui-serve and component tests ran with Cypress on Windows, I'll need to discuss that with @Gnuk when he's available.

Gnuk:
> FYI: there's an issue with tikui-serve and component tests ran with Cypress on Windows, I'll need to discuss that with @Gnuk when he's available.

I'm back next week :)

murdos:
@pascalgrimaud : this is now ready for merge (there's no breaking change since it only relates to the internal of the generator) :) With the help of @Gnuk I fixed the windows build.

pascalgrimaud:
can you fix the conflict plz @murdos ?

murdos:
@pascalgrimaud : conflicts are fixed

PR #9883, opened 2024-05-23
use standard vite/vitest configuration file extension

murdos:
This is now possible because package-json is configured with type=module


qmonmert (approved):
👍

renanfranca (approved):
Thanks!

PR #9899, opened 2024-05-24
feat: generate clients with a .npmrc config for forcing saving exact versions of npm dependencies

murdos:
This is already in place for the generator, and this enforces a good practice


qmonmert (approved):

renanfranca (approved):
Thank you @murdos ! That is important, as it avoids awkward and annoying situations!

PR #9907, opened 2024-05-25
fix: move gitignore configuration for buildSrc from spring-boot-local-profile module to gradle-java module

murdos:
Otherwise gradle gitignore configuration is added to maven projects with spring-boot-local-profile applied:
image


renanfranca (approved):
I agree! It also helps avoid potential code duplication in the `.gitignore` file.

Thanks 😁!

PR #9930, opened 2024-05-28
chore(clients): reduce duplicated npm dependencies for clients

murdos:
By factorizing dependencies in common npm module, dependency updates by dependabot will be reduced. If at some point one client needs a distinct version from common version, we can declare it in the specialized client module


renanfranca (approved):
Nice! 😁

PR #9963, opened 2024-05-31
Springdoc: fix realm name with OAuth2

pascalgrimaud:
Following https://github.com/jhipster/jhipster-lite/pull/9366


renanfranca (approved):
Thanks! All good for me 😄

PR #9971, opened 2024-06-02
chore: simplify code by using Optional::stream

murdos:


renanfranca (approved):
That's really nice, I learned it last week! Thanks 😁

PR #9972, opened 2024-06-02
enh: use "release" configuration for setting java version rather than "source" and "target"

murdos:
See https://maven.apache.org/plugins/maven-javadoc-plugin/jar-mojo.html#source and https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html


renanfranca (approved):
Thanks! That's a tricky pitfall! Well done!

PR #9997, opened 2024-06-05
fix(java-build-tool): remove misleading wrapper mention in the descri…

murdos:
…ption of java-build-tools module resources


renanfranca (approved):
Thanks!

PR #10014, opened 2024-06-08
React: remove unused dependency

qmonmert:


murdos:
Good catch @qmonmert :)

PR #10015, opened 2024-06-08
Vue: use standart vite/vitest gonfig file extension

qmonmert:


src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
```diff
@@ -45,6 +45,7 @@ public JHipsterModule buildVueModule(JHipsterModuleProperties properties) {
     return ClientsModulesFactory.clientModuleBuilder(properties)
       .documentation(documentationTitle("Vue"), DOCUMENTATION_SOURCE.file("vue.md"))
       .packageJson()
+        .addType("module")
```
murdos, line 48: From memory, it had an impact on eslint configuration, that has to be explicitly renamed to `.eslintrc.cjs`, see https://github.com/jhipster/jhipster-lite/pull/8563

FYI I'm planning to switch eslint config to use module imports, and switch back to simple `.js` extension once we've migration to ESLint v9 (since it supports and detects projects configured with type=module)

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
```diff
@@ -66,21 +67,21 @@ public JHipsterModule buildVueModule(JHipsterModuleProperties properties) {
         .addDevDependency(packageName("sinon"), VUE)
         .addScript(scriptKey("build"), scriptCommand("vue-tsc -p tsconfig.build.json --noEmit && vite build --emptyOutDir"))
         .addScript(scriptKey("dev"), scriptCommand("vite"))
-        .addScript(scriptKey("lint"), scriptCommand("eslint --ext .js,.ts,.vue src"))
+        .addScript(scriptKey("lint"), scriptCommand("eslint --ext .js,.ts,.vue src/main/webapp/app/**/*"))
```
murdos, line 70: Is there a reason to not lint anymore test sources?

qmonmert: I did the same thing that in React app, else it doesn't work with `.eslintrc.js` in `test/javascript/integration` folder

PR #10052, opened 2024-06-12
Update versions in github-actions templates

murdos:


renanfranca (approved):
Thank you for your work @murdos ! 😃

PR #10075, opened 2024-06-15
UI: hide minimap by default

pascalgrimaud:
I'd prefer to hide the minimap as it hides the init module when we open the UI:

image


murdos (approved):

murdos:
Maybe all modules should at least depend on the `init` module.
It does not really make sense to have renovate, gitpod, ... if you have not initialized your project.

pascalgrimaud:
> It does not really make sense to have renovate, gitpod, ... if you have not initialized your project.

Maybe I think it can still work. Let discuss about it in another ticket
But I don't think all modules need init: for example license do not need init :-D

PR #10099, opened 2024-06-17
Sonar: reactivate S2699 rule

qmonmert:


src/main/resources/generator/client/react/src/test/javascript/spec/login/primary/loginForm/index.test.tsx.mustache
```diff
@@ -28,10 +28,6 @@ const login = async () => {
 };
 
 describe('loginForm', () => {
-  it('should render the login button without crashing', () => {
```
qmonmert, line 31: tested in the following test

src/main/resources/generator/client/react/src/test/javascript/spec/login/primary/loginModal/index.test.tsx.mustache
```diff
@@ -14,10 +14,6 @@ const LoginModalRender = (open: boolean) => render(
 </UserInfoContext.Provider>);
 
 describe('test login modal', () => {
-  it('describe should render without crashing', () => {
```
qmonmert, line 17: same

PR #10238, opened 2024-07-09
refactor(VersionsCatalog): move libraryEntriesMatchingDependency meth…

renanfranca:
…od next to the method that uses it first


murdos (approved):
Good Idea 😀

renanfranca:
> Good Idea 😀

Thanks!!!

PR #10264, opened 2024-07-12
feature(protobuf): use ascopes maven protobuf plugin instead of xolstice one

murdos:
This plugin is actively maintained, is simpler to use and aims to be compatible with maven 4

https://github.com/ascopes/protobuf-maven-plugin?tab=readme-ov-file#why-do-we-need-another-plugin


src/main/resources/generator/dependencies/pom.xml
```diff
@@ -54,7 +54,7 @@
     <jqwik.version>1.9.0</jqwik.version>
     <!-- Maven Plugins -->
     <os-maven-plugin.version>1.7.1</os-maven-plugin.version>
-    <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
+    <protobuf-maven-plugin.version>2.3.0</protobuf-maven-plugin.version>
```
ascopes, line 57:
```suggestion
    <protobuf-maven-plugin.version>2.4.0</protobuf-maven-plugin.version>
```

As discussed, the regression you raised should now be fixed that was causing your build failures. Thanks for bearing with!

murdos: Thanks for the quick fix! :)

PR #10269, opened 2024-07-13
Bug: removed SpringDoc apiLicenseUrl and apiExternalDocUrl blank value

### LINKED ISSUE #10262: SpringDoc external URL and licence URL should not be blank
`SpringdocModuleFactory` set `apiLicenseUrl` and `apiExternalDocUrl` to blank which is not valid and will cauyse parsing problem (ex: to use with OpenApi diff).

Either need to remove this properties from the template or valuated them

richersoon:
Close #10262


src/main/java/tech/jhipster/lite/generator/server/springboot/apidocumentation/springdoccore/domain/SpringdocModuleFactory.java
```diff
@@ -42,9 +42,7 @@ private JHipsterModule buildModule(JHipsterModuleProperties properties, SpringDo
         .put("apiTitle", "Project API")
         .put("apiDescription", "Project description API")
         .put("apiLicenseName", "No license")
-        .put("apiLicenseUrl", "")
```
DamnClin, line 45: Also needs to be removed from the template

DamnClin (approved):

src/main/java/tech/jhipster/lite/generator/server/springboot/apidocumentation/springdoccore/domain/SpringdocModuleFactory.java
richersoon: Hi @DamnClin , I already just removed them from the template. Please check again.

DamnClin: Yep, thanks, LGTM :)

PR #10278, opened 2024-07-14
Bug: added check if repository is null in CucumberJpaReset

### LINKED ISSUE #10097: Cucumber JPA reset should not fail without repository
CucumberJpaReset fails when there is no declared repository, should be updated to:

```java
public class CucumberJpaReset {

  @Autowired(required = false)
  private Collection<JpaRepository<?, ?>> repositories;

  @After
  @Before
  @Transactional
  public void wipeData() {
    if (repositories == null) {
      return;
    }

    repositories.forEach(JpaRepository::deleteAllInBatch);
  }
}
```

richersoon:
Close #10097

### DIFF
--- a/src/main/resources/generator/server/springboot/cucumber/CucumberJpaReset.java.mustache
+++ b/src/main/resources/generator/server/springboot/cucumber/CucumberJpaReset.java.mustache
@@ -3,6 +3,7 @@ package {{packageName}}.cucumber;
 import io.cucumber.java.After;
 import io.cucumber.java.Before;
 import java.util.Collection;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.data.jpa.repository.JpaRepository;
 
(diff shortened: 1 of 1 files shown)


src/main/resources/generator/server/springboot/cucumber/CucumberJpaReset.java.mustache
```diff
@@ -18,6 +18,10 @@ public class CucumberJpaReset {
   @Before
   @Transactional
   public void wipeData() {
+    if (repositories == null) {
```
DamnClin, line 21: Not sure that will be enough without de required false on the repositories injection

richersoon: I've also added the required false on repositories injection, please check @DamnClin . thanks

DamnClin (approved):

pascalgrimaud:
Thanks for your contribution @richersoon

PR #10297, opened 2024-07-16
manage landscape preset configuration

### LINKED ISSUE #10239: Manage landscape preset configuration
I propose saving/reading the preset configuration into a JSON file, similar to how `history.json` works. Here is what I think the `.jhipster/modules/preset.json` should look like:

```json
[
  {
    "name": "angular + spring boot",
    "date": "2024-07-09T15:10:00.521252170Z",
    "modules": [
      "init",
      "application-service-hexagonal-architecture-documentation",
      "maven-java",
      "prettier",
      "angular-core",
      "java-base",
      "maven-wrapper",
      "spring-boot",
      "spring-boot-mvc-empty",
      "logs-spy",
      "spring-boot-tomcat"
    ]
  }
]
```

When applying the selected preset, the properties defined by the landscape interface or default will be used, which is why they won't be stored in the json.
```json
    "properties": {
      "packageName": "com.mycompany.myapp",
      "projectName": "JHipster Sample Application",
      "baseName": "jhipsterSampleApplication",
      "serverPort": 8080,
      "endOfLine": "lf",
      "indentSize": 2,
      "springConfigurationFormat": "yaml"
    },
```

renanfranca:
Fix #10239


src/main/java/tech/jhipster/lite/project/infrastructure/secondary/FileSystemProjectsRepository.java
```diff
@@ -84,4 +90,13 @@ public ProjectHistory getHistory(ProjectPath path) {
   private Path historyFilePath(ProjectPath path) {
     return Paths.get(path.get(), HISTORY_FOLDER, HISTORY_FILE);
   }
+
+  @Override
+  public Collection<Preset> getPreset() {
+    try {
+      return json.readValue(projectFiles.readBytes(PRESET_FILE), PersistedPresets.class).toDomain();
```
murdos, line 97: Am i missing something? I thought we decided to not store presets files in project: https://github.com/jhipster/jhipster-lite/issues/10239#issuecomment-2219864428

src/main/java/tech/jhipster/lite/project/infrastructure/secondary/FileSystemProjectsRepository.java
```diff
@@ -84,4 +90,13 @@ public ProjectHistory getHistory(ProjectPath path) {
   private Path historyFilePath(ProjectPath path) {
     return Paths.get(path.get(), HISTORY_FOLDER, HISTORY_FILE);
   }
+
+  @Override
+  public Collection<Preset> getPreset() {
```
murdos, line 95: Since this is a collection, maybe:
```suggestion
  public Collection<Preset> getPresets() {
```

src/main/java/tech/jhipster/lite/project/infrastructure/secondary/FileSystemProjectsRepository.java
renanfranca: @murdos: As far as I understand, the code from `projectFiles.readBytes` reads the file from the resource in the JHipster Lite project and not from the generated project. I use `projectFiles` to avoid duplicating code. Let me know if I misused it 😅

```java
@Override
@ExcludeFromGeneratedCodeCoverage(reason = "The error handling is a hard to test implementation detail")
public byte[] readBytes(String path) {
    Assert.notBlank("path", path);

    try (InputStream input = getInputStream(path)) {
        assertFileExist(path, input);
        return toByteArray(input);
    } catch (IOException e) {
        throw GeneratorException.technicalError("Error closing " + path + ": " + e.getMessage(), e);
    }
}

private void assertFileExist(String path, InputStream input) {
    if (input == null) {
        throw GeneratorException.technicalError("Can't find file: " + path);
    }
}

private InputStream getInputStream(String path) {
    return FileSystemProjectFiles.class.getResourceAsStream(path.replace("\\", SLASH));
}
```

renanfranca: In this case the preset.json file must be created at `src/java/resources` in the jhipster-lite source code 👍

murdos: Oh, my bad, sorry 😅

murdos (approved):

src/main/java/tech/jhipster/lite/project/infrastructure/secondary/FileSystemProjectsRepository.java
renanfranca: It's cool 😉

PR #10367, opened 2024-07-22
return presets through endpoint

### LINKED ISSUE #10317: Implement the rest api which return the presets stored at `src/main/resources`
Should create the endpoint GET `/api/presets`. See #9491 description for the complete roadmap.

renanfranca:
Fix #10317


src/test/java/tech/jhipster/lite/project/infrastructure/secondary/SpiedFileSystemProjectFilesConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.nio.charset.StandardCharsets;
+import org.mockito.Mockito;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Primary;
+import tech.jhipster.lite.module.infrastructure.secondary.FileSystemProjectFiles;
+
+@TestConfiguration
+public class SpiedFileSystemProjectFilesConfiguration {
```
murdos, line 11: Maybe?
```suggestion
public class FakeFileSystemProjectFilesConfiguration {
```

src/test/java/tech/jhipster/lite/project/infrastructure/secondary/SpiedFileSystemProjectFilesConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Bean
+  @Primary
+  public FileSystemProjectFiles fileSystemProjectFiles() {
```
murdos, line 15: Usually beans and spring configuration (class above) don't need to be public, since they're not accessed by other classes.
```suggestion
  FileSystemProjectFiles fileSystemProjectFiles() {
```

murdos (approved):

src/test/java/tech/jhipster/lite/project/infrastructure/secondary/SpiedFileSystemProjectFilesConfiguration.java
renanfranca: @murdos: I followed the naming convention of another class, `MockedProjectFormatterConfiguration.java` (https://github.com/jhipster/jhipster-lite/blob/main/src/test/java/tech/jhipster/lite/project/infrastructure/secondary/MockedProjectFormatterConfiguration.java). Additionally, it's not fake; the spy allows the class to work in its original way. Maybe I should mock the specific behavior of "/preset.json" in a different class using the `@Before` Cucumber annotation. WDYT?

src/test/java/tech/jhipster/lite/project/infrastructure/secondary/SpiedFileSystemProjectFilesConfiguration.java
renanfranca: I will do this 👍

src/test/java/tech/jhipster/lite/project/infrastructure/secondary/SpiedFileSystemProjectFilesConfiguration.java
murdos: Ok, I see. It's just that the name (Spied...) is a bit unusual.
BTW would it work if you put a `preset.conf` at root of test resources?

renanfranca: > Ok, I see. It's just that the name (Spied...) is a bit unusual.

Can I change it to Faked prefix?

> BTW would it work if you put a `preset.conf` at root of test resources?

Did you mean `preset.json`? I tried this path a little,  but I have to change the behavior of `FileSystemProjectFiles.class.getResourceAsStream(path.replace("\\", SLASH));` (https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/FileSystemProjectFiles.java) and it is used by another classes in CucumberTest to access resources at `/generator/....`.

murdos: Ok, then let's use for Faked prefix :)

PR #10389, opened 2024-07-25
Prettier config improvements

murdos:
- node_modules and version control systems directories are already ignored by default: https://prettier.io/docs/en/ignore.html#ignoring-files-prettierignore
- there's no particular reason for ignoring .mvn folder
- use correct formating for .prettierrc

### DIFF
--- a/src/main/resources/generator/prettier/.prettierignore
+++ b/src/main/resources/generator/prettier/.prettierignore
@@ -1,6 +1,5 @@
 target
 build
 package-lock.json
-.mvn
 gradle
 .gradle
--- a/src/main/resources/generator/prettier/.prettierignore
+++ b/src/main/resources/generator/prettier/.prettierignore
@@ -1,8 +1,6 @@
-node_modules
 target
 build
 package-lock.json
-.git
 .mvn
 gradle
 .gradle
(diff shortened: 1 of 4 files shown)


src/main/resources/generator/prettier/.prettierignore
```diff
@@ -1,8 +1,5 @@
-node_modules
 target
 build
 package-lock.json
-.git
-.mvn
```
pascalgrimaud, line 6: the reason was probably because in old maven wrapper we had this file  ̀MavenWrapperDownloader.java`
See https://github.com/jhipster/jhipster-sample-app/tree/v6.10.0/.mvn/wrapper

As now, it's not needed anymore we could indeed remove or let it

pascalgrimaud (dismissed):

qmonmert (approved):
Update this line https://github.com/jhipster/jhipster-lite/blob/e943e8924ccb3ccdfd30cf17e513b902f2f1e89d/src/test/java/tech/jhipster/lite/generator/prettier/domain/PrettierModuleFactoryTest.java#L36 with `.containing("endOfLine: 'crlf'")`

PR #10390, opened 2024-07-25
Angular: Remove unused code in test file

qmonmert:


murdos:
@qmonmert : Good catch! :)

PR #10399, opened 2024-07-26
feat(archUnit): adding equals/hashcode contract check

Eulbobo:
This feature adds a new ArchUnit test that is designed to make a simple equals/hashcode validation.

Every class implementing equals(Object) should also implement hashCode().
You have to implement both or none in each class for this new test to pass.

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/javatool/archunit/domain/ArchUnitModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/javatool/archunit/domain/ArchUnitModuleFactory.java
@@ -27,6 +27,7 @@ public class ArchUnitModuleFactory {
         .add(SOURCE.template("archunit.properties"), to("src/test/resources/archunit.properties"))
         .add(SOURCE.template("AnnotationArchTest.java"), testDestination.append("AnnotationArchTest.java"))
         .add(SOURCE.template("HexagonalArchTest.java"), testDestination.append("HexagonalArchTest.java"))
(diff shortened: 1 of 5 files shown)


murdos (approved):

DamnClin:
For me, this should be in a dedicated module (with a dedicated test) since not everybody will agree with this rule

murdos:
> For me, this should be in a dedicated module (with a dedicated test) since not everybody will agree with this rule
FYI it's me that have encouraged @Eulbobo to put it in the existing module, since IMO it's an universal rule, enforced in the Effective Java book.
In order to understand, do you have examples where you would not want to follow this rule?

DamnClin:
> > For me, this should be in a dedicated module (with a dedicated test) since not everybody will agree with this rule
> 
> FYI it's me that have encouraged @Eulbobo to put it in the existing module, since IMO it's an universal rule, enforced in the Effective Java book. In order to understand, do you have examples where you would not want to follow this rule?

For most domain objects I don't really care about that and I'm afraid that this will create a lot of noise in junior team that will try to unit test those methods & not use builders to make simple implementations for that.

In many contexts, this sound like a bad tradeoff, I think no equals/hashCode is better than bad costly ones

DamnClin:
Totally My Bad here!!! I was reading this will on a meeting and missed the point, I though this was to enforce equals & hashCode on all object

DamnClin (approved):

renanfranca:
Congratulations @Eulbobo for your first contribution here! Thank you! 🎉👏😃

PR #10404, opened 2024-07-27
TS Loader module

DamnClin:
Small module to add a Loader object for frontend applications

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/client/loader/infrastructure/primary/TsLoaderModuleConfiguration.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/loader/infrastructure/primary/TsLoaderModuleConfiguration.java
@@ -1,0 +1,23 @@
+package tech.jhipster.lite.generator.client.loader.infrastructure.primary;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import tech.jhipster.lite.generator.client.loader.application.TsLoaderModuleApplicationService;
+import tech.jhipster.lite.generator.slug.domain.JHLiteModuleSlug;
+import tech.jhipster.lite.module.domain.resource.JHipsterModuleOrganization;
+import tech.jhipster.lite.module.domain.resource.JHipsterModuleResource;
+
+@Configuration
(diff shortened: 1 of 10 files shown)


src/main/java/tech/jhipster/lite/generator/client/loader/infrastructure/primary/TsLoaderModuleConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+@Configuration
+class TsLoaderModuleConfiguration {
+
+  @Bean
+  JHipsterModuleResource tsLoaderModule(TsLoaderModuleApplicationService tsLoader) {
+    return JHipsterModuleResource.builder()
+      .slug(JHLiteModuleSlug.TS_LOADER)
+      .withoutProperties()
+      .apiDoc("Client", "Helper class to represent loading states")
+      .organization(JHipsterModuleOrganization.builder().addDependency(JHLiteModuleSlug.PRETTIER).build())
```
murdos, line 19: Is there a reason for requiring the prettier module?

DamnClin: Not really, was hesitating... Init is probably a better idea

murdos (approved):

src/main/java/tech/jhipster/lite/generator/client/loader/infrastructure/primary/TsLoaderModuleConfiguration.java
murdos: In fact this requires typescript, but our client modules and the typescript module don't have any thing in common, so we would have to choose between client or typescript as prerequisite.
Maybe we should try to have some (all?) of your client modules depend on the typescript module (and remove the typscript setup they embed). WDYT?

DamnClin: Hopefully I'll find some motivation to work on the frontend modules in the next weeks and I'll probably stumble into that problem...

PR #10416, opened 2024-07-27
Angular: add comment in test

qmonmert:


src/main/java/tech/jhipster/lite/generator/client/angular/security/oauth2/domain/AngularOauth2ModuleFactory.java
```diff
@@ -95,8 +95,10 @@ public class AngularOauth2ModuleFactory {
     """
 
     it('should display login component', () => {
+      // WHEN
```
DamnClin, line 98: IMO this is a bad practice, two reasons for that:

* In unit test, this should be Arrange / Act / Assert (Given / When / Then is for gherkin / cucumber tests)
* This don't add any information since the test structure is enough to see it so it's just adding noise to the tests

qmonmert: @DamnClin there are a lot of comments like this in angular app, I remove all?

DamnClin: For me it's better to remove them but I don't know of any universal best practice on that so, in the end, do as you want :)

qmonmert: my idea was to have the same thing on all tests files and it was the only without comment

PR #10419, opened 2024-07-27
fix(module-api): VersionSource should not be required for removing dependencies from packageJson

murdos:
I've deprecated the previous API and introduced a simpler one without VersionSource


qmonmert (approved):

renanfranca (approved):
Thank you for your work! Nice improvement, @murdos !

Issue #10421, opened 2024-07-27
Client modules using typescript should depend on the typescript module


murdos:
Following a [recent discussion](https://github.com/jhipster/jhipster-lite/pull/10404#discussion_r1693936727) on the addition of new modules adding typescript utilities, it demonstrates that we currently have an issue regarding typescript related modules.

Let's pick the `optional-typescript` as an example: it requires the `typescript`. 
So if you have applied for example the `vue-core` module, you can't use it unless you also apply the `typescript` module which will generates conflicting npm configuration (e.g. setting up jest in addition to already setup vitest).

On the contrary the new modules, [ts-loader](https://github.com/jhipster/jhipster-lite/pull/10404) and [ts-pagination-domain](https://github.com/jhipster/jhipster-lite/pull/10405) only requires the `init` module, which is convenient for matching both use cases, but a bit wrong since code will not compile if you don't have already applied a module that'll setup typescript.

My suggestion is to rework the `typescript` module to make it use a modern setup (eslint, vitest), and have at least `vue-core` and `react-core` modules (angular could be nice too, but it has a very specific build tools so this might be a bit more complicated) depend on the `typescript` module.

@pascalgrimaud @qmonmert @DamnClin @renanfranca : WDYT?

Related: https://github.com/jhipster/jhipster-lite/issues/3979

renanfranca:
I agree with this 💯 👌.  It will help new frontend devs 😁👍

maznag:
we need to remove `"type": "module"` from package.json for angular case. 

`npm run test:coverage` fail with this configuration and we get this error : `An unhandled exception occurred: require is not defined`

[error] ReferenceError: require is not defined
    at file:///................/jest.conf.js:1:37
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:400:35)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:427:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1561:24)
    at Module._compile (node:internal/modules/cjs/loader:1712:5)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)

murdos:
[...]

PR #10450, opened 2024-07-29
refactor(client-common): remove ClientsModulesFactory that was only containing lintstaged configuration

murdos:
ClientsModulesFactory was previously doing more things, but it now only append lintstaged configuration. By moving things to each client, we can use a more fined grained configuration (e.g. vue extension only for vue client, tsx for react, ...).

Preparation for #10421


src/main/java/tech/jhipster/lite/generator/server/springboot/thymeleaf/template/domain/ThymeleafTemplateModuleFactory.java
```diff
@@ -83,7 +82,7 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
     Assert.notNull(PROPERTIES, properties);
 
     //@formatter:off
-    return ClientsModulesFactory.clientModuleBuilder(properties)
+    return moduleBuilder(properties)
```
murdos, line 85: @renanfranca : I don't think this was useful for Thymeleaf, since eslint is not even a dependency, so it could be missing.
WDYT?

renanfranca: That was my first issue adding the "frontend" module. I remember using lint in the generated project to help me format the template code, but I am not sure if the `Client....` class helped me. However, there is a Prettier configuration in the deleted `Client....` class.

Maybe I copied and pasted 🤔  and learned while doing...

PR #10459, opened 2024-07-30
Sonar: remove exclusions files not present

qmonmert:


murdos:
Good catch @qmonmert !

PR #10462, opened 2024-07-31
Unify and simplify vue and react vitest config

murdos:
-  make vitest configuration inherit from vite configuration
- use simpler and more canonical include and exclude expressions, relying on default vitest configuration

Preparation for #10421


renanfranca (approved):
Thanks!

qmonmert (approved):

PR #10463, opened 2024-07-31
fix(bug) build fails on windows with autocrlf=lf

### LINKED ISSUE #10398: Build fails on Windows if you cloned with autocrlf=input, eol=lf
Some implementations relies on System.lineSeparator() to split configuration file lines

If you are running on Windows, your default lineSeparator is `\n\r`, but, if you cloned the project with LF end of lines, the split can't be done anymore and build fails
image

Issue https://github.com/jhipster/jhipster-lite/issues/477 provided an improved way to deal with end of lines, so it may be the solution

Maybe I'm the only one using this setup, but I'll try and fix it anyway

Eulbobo:
Changing end of file and properties String manipulation to use Properties class

close #10398

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/javaproperties/PropertyValue.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/javaproperties/PropertyValue.java
@@ -1,7 +1,7 @@
 package tech.jhipster.lite.module.domain.javaproperties;
 
-import java.util.ArrayList;
 import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import tech.jhipster.lite.shared.collection.domain.JHipsterCollections;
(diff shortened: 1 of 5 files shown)


murdos (approved):

DamnClin (commented):
I have some doubts on that. Does this keep the other parts of the file intact (not changing order, not deleting blank lines etc)???

Eulbobo:
No it does not, and I didn't think it should since there was no test for this :/

Order should be preserved, but empty lines will definitly go away.
And a file header comment is automatically generated with generation date for the file (can't remove it, can only provide a default value on properties.store(outputStream, comment))

If those points are mandatory behaviours for the component, they should be tested accordingly beforehand

DamnClin:
> No it does not, and I didn't think it should since there was no test for this :/
> 
> Order should be preserved, but empty lines will definitly go away. And a file header comment is automatically generated with generation date for the file (can't remove it, can only provide a default value on properties.store(outputStream, comment))
> 
> If those points are mandatory behaviours for the component, they should be tested accordingly beforehand

For me they are, we learned that later on but I think this is really important. @pascalgrimaud  any advice on this?

murdos:
IMO these points are indeed mandatory.
We try to preserve as much as possible the existing source file, and the header will generate some additional noise in system versioning changes.

Eulbobo:
First step : adding tests to check those behaviours in existing class
- order in file preserved
- comment preserved
- white line preserved
- new key/values appended at the end of the file
- new values appended at the end of current existing values for key

Something like this :
image
(yes, it works even on my setup, be with a very ugly fix i'm too ashamed to show here)

This should go in another pull request/issue

PR #10468, opened 2024-07-31
Modernize typescript module

murdos:
- simpler tsconfig
- setup vitest
- remove jest setup in optional-ts module

Preparation for #10421


renanfranca (approved):
Thank you, @murdos! I am used to using Vitest now! I really like it. Unit testing is so powerful for testing behavior rather than just testing the code by mocking everything 😅!

PR #10490, opened 2024-08-03
Sonar: remove svelte files

qmonmert:
Svelte is not supported


pascalgrimaud:
In fact, Svelte was supported but disabled because of lack of interest

qmonmert:
@pascalgrimaud ok

PR #10505, opened 2024-08-04
Add tests for Router (improve Sonar)

qmonmert:
Remove `src/main/webapp/app/router/index.ts` from exclusion files on sonar-project.properties


murdos:
I'm not sure that these tests have a real value.
I agree that we should not exclude the router code from sonar coverage, but we should rather merge the code coverage of the unit and components tests, that will allow indirectly verify all routes are used.

murdos:
Anyway, I think we can merge this PR, and re-evaluate when/if we merge frontend code coverage.

murdos (approved):

PR #10561, opened 2024-08-14
Docs: sonar commands in readme

### LINKED ISSUE #10552: Docs: sonar commands in readme
The sonar commands in readme should be in a dedicated section instead of startup
See https://github.com/jhipster/jhipster-lite-sample-app

image

qmonmert:
Fix #10552 

image


pascalgrimaud:
after thinking about it, I'm not sure we should add a new method `sonarAnalysisCommands` as it is used only for Sonar

I would suggest:
- remove the startup section related to Sonar
- let the link to sonar.md
- in sonar.md, update this file and add the command to start sonar + to analyze

What do you think @qmonmert ?

murdos:
I agree, sonarAnalysisCommands is really specific and won't be reusable in other cases.

qmonmert:
@pascalgrimaud ok I understand, I provide a new PR

PR #10596, opened 2024-08-18
Switch to Kafka native image in KafkaTestContainer

### LINKED ISSUE #10502: Kafka: use native image for tests with testcontainers
See https://x.com/gunnarmorling/status/1820022466841907278?t=IgpHq5Oh7yJa3tKFwwqpqg&s=19

richersoon:
Close #10502

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/broker/kafka/domain/KafkaModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/broker/kafka/domain/KafkaModuleFactory.java
@@ -31,6 +31,7 @@ public class KafkaModuleFactory {
     return moduleBuilder(properties)
       .context()
         .put("kafkaDockerImage", dockerImages.get("apache/kafka").fullName())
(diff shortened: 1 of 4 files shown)


src/main/java/tech/jhipster/lite/generator/server/springboot/broker/kafka/domain/KafkaModuleFactory.java
```diff
@@ -31,6 +31,7 @@ public JHipsterModule buildModuleInit(JHipsterModuleProperties properties) {
     return moduleBuilder(properties)
       .context()
         .put("kafkaDockerImage", dockerImages.get("apache/kafka").fullName())
+        .put("kafkaDockerNativeImage", dockerImages.get("apache/kafka-native").fullName())
```
murdos, line 34: I think you can just reuse the `kafkaDockerImage` variable, so we'll also use the native image for docker compose.
And thus only use one Docker image.

murdos (changes requested):

Issue #10613, opened 2024-08-19
Cypress/Playwright: split current module in two modules: e2e tests and component-tests


murdos:
The current `cypress` module generates something that looks like frontend component tests (using the Martin Folwer definition, not the Cypress definition of component tests): things are generated in `src/test/webapp/component`, npm script is `test:component`, ...

We should probably improve the current module to also automatically start the frontend server, so component tests can be launched without prerequisites by launching `npm run test:component` (see current generator configuration).

And create a new one that would generate Cypress configuration for running E2E.

murdos:
@pascalgrimaud : bounty claimed https://opencollective.com/generator-jhipster/expenses/224383

pascalgrimaud:
@murdos : approved

PR #10623, opened 2024-08-20
feat(cypress): automatically start dev server before launching frontend component tests

murdos:
See #10613


renanfranca (approved):
Thanks! Sorry for the delay!

PR #10638, opened 2024-08-22
Replace eslint-plugin-prettier usage by eslint-config-prettier

### LINKED ISSUE #10626: Replace eslint-plugin-prettier usage by eslint-config-prettier
See the Prettier's official recommendations: https://prettier.io/docs/en/integrating-with-linters.html

We should also configure lintstaged to run `eslint --fix`, then `prettier --write` instead of only `eslint --fix`

murdos:
Fixes #10626


renanfranca (approved):
Thanks @murdos ! Good luck on the e2e quest 😊

PR #10639, opened 2024-08-23
add react internationalization

### LINKED ISSUE #10413: Add frontend internalionalization
Not that it is really usefull in JHLite context BUT the generator is a good example so we should provide an example for this really common usecase

fabienpuissant:
add react internationalization

create react internalization module using i18next 23.14.0 with language detector see https://react.i18next.com/latest/using-with-hooks


src/main/java/tech/jhipster/lite/generator/client/tools/reacti18n/application/ReactI18nApplicationService.java
```diff
@@ -0,0 +1,20 @@
+package tech.jhipster.lite.generator.client.tools.reacti18n.application;
```
murdos, line 1: A better location for the whole module would be 
```suggestion
package tech.jhipster.lite.generator.client.react.i18n.application;
```
There are already `core` and `security` subpackages in the `tech.jhipster.lite.generator.client.react` package :)

src/main/java/tech/jhipster/lite/generator/client/tools/reacti18n/domain/ReactI18nModuleFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.file.JHipsterSource;
+import tech.jhipster.lite.module.domain.packagejson.VersionSource;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+import tech.jhipster.lite.shared.error.domain.Assert;
+
+public class ReactI18nModuleFactory {
+
+  private static final JHipsterSource APP_SOURCE = from("client/common/reacti18n/src/main/webapp/app");
```
murdos, line 13: You should probably create a new `core` subdirectory in `src/main/resources/generator/client/react` where you'll move existing resources, and create a subdirectory `src/main/resources/generator/client/react/i18n`.

See how `src/main/resources/generator/client/angular` is structured as an example.

src/main/java/tech/jhipster/lite/generator/client/tools/reacti18n/domain/ReactI18nModuleFactory.java
```diff
@@ -0,0 +1,70 @@
...
+  private static final String INDEX = "src/main/webapp/";
+  private static final String INDEX_TEST = "src/test/webapp/unit/common/primary/app/";
+
+  public JHipsterModule buildModule(JHipsterModuleProperties properties) {
+    Assert.notNull("properties", properties);
+
+    //@formatter:off
+    return moduleBuilder(properties)
+      .packageJson()
+      .addDependency(packageName("i18next"), VersionSource.REACT)
```
murdos, line 25: If the three `i18next-*` dependencies are going to be used also by Angular and Vue, we can put them in the `COMMON` VersionSource

src/main/java/tech/jhipster/lite/generator/client/tools/reacti18n/infrastructure/primary/ReactI18nModuleConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Configuration
+public class ReactI18nModuleConfiguration {
+
+  @Bean
+  JHipsterModuleResource i18nModule(ReactI18nApplicationService i18n) {
+    return JHipsterModuleResource.builder()
+      .slug(REACT_I18N)
+      .propertiesDefinition(JHipsterModulePropertiesDefinition.builder().build())
+      .apiDoc("react i18n", "Add react internationalization")
```
murdos, line 22: The first parameter is the group where the module will appear in the [Patches view](https://lite.jhipster.tech/patches)
```suggestion
      .apiDoc("Frontend - React", "Add react internationalization")
```

src/main/resources/generator/client/common/reacti18n/src/main/webapp/app/i18n.ts.mustache
```diff
@@ -0,0 +1,24 @@
...
+
+  .init({
+    fallbackLng: 'en',
+    debug: false,
+
+    interpolation: {
+      escapeValue: false,
+    },
+    backend: {
+      loadPath: '../assets/locales/{{ lng }}/{{ ns }}.json',
```
murdos, line 20: `{{ ... }}` is also the variable interpolation used by the Mustache templating that is used by the jhipster-lite generator.
If there's no dynamic content in this file, you'd better remove the `.mustache` suffix in the file name, to avoid some unexpected side effects.

murdos: I think you forgot to rename this file.

murdos:
> add react internationalization
> 
> Fix #10413

@DamnClin : if I'm not mistaken, your issue #10413 was more about adding i18n to the generator itself (i.e. to the application that runs jhlite)?

I think @fabienpuissant that your PR is also an interesting feature, but we should probably not close the mentioned issue:)

fabienpuissant:
> > add react internationalization
> > Fix #10413
> 
> @DamnClin : if I'm not mistaken, your issue #10413 was more about adding i18n to the generator itself (i.e. to the application that runs jhlite)?
> 
> I think @fabienpuissant that your PR is also an interesting feature, but we should probably not close the mentioned issue:)

Yes maybe I did'nt get the point of Colin, cause I think that i18n on the front end app is a nice to have :)

murdos:
Congrats @fabienpuissant for your first contribution! :)

PR #10647, opened 2024-08-23
feat(landscape): reduce opacity of non selectable modules

murdos:
image

### DIFF
--- a/src/main/style/atom/chip/_chip.scss
+++ b/src/main/style/atom/chip/_chip.scss
@@ -104,6 +104,15 @@ $jhlite-anchor-point-border-size: 3px;
   }
 }
 
+.-not-selectable {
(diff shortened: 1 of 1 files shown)


renanfranca (approved):
Nice improvement 👏 . There is no need to hover to foresight what is possible or not possible to select 😁

Thank you for your work @murdos

qmonmert (approved):

DanielFran:
@murdos  Maybe it should be done the same for the module group name if no module can be used?

Example: front-browser-test
image

murdos:
> @murdos Maybe it should be done the same for the module group name if no module can be used?

@DanielFran: Good idea! However it requires a bit of work to compute the DOM class of the module group name, it's not currently done.

PR #10649, opened 2024-08-23
Cypress/Playwright: split current module in two modules: e2e tests and component-tests

### LINKED ISSUE #10613: Cypress/Playwright: split current module in two modules: e2e tests and component-tests

murdos:
- Fixes #10613


renanfranca (approved):
Great work! Thank you very much @murdos !!!

Issue #10621, opened 2024-08-20
Dockerfile should use ENTRYPOINT instead of CMD


pascalgrimaud:
The current Dockerfile used to build the Docker image, use `CMD` to start the application.

We should use `ENTRYPOINT` instead, to avoid having some issue when deploying to Kubernetes

pascalgrimaud:
agree with you @murdos but it's a new idea for a new module :-)

PR #10659, opened 2024-08-24
Dockerfile: use entrypoint instead of cmd

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/10621

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,5 +14,5 @@ RUN \
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JAVA_OPTS=""
 USER jhipster
-CMD java ${JAVA_OPTS} -Djava.security.egd=file:/dev/./urandom -jar /code/*.jar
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -14,5 +14,5 @@ RUN \
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JAVA_OPTS=""
 USER jhipster
-CMD java ${JAVA_OPTS} -Djava.security.egd=file:/dev/./urandom -jar /code/*.jar
+ENTRYPOINT exec java ${JAVA_OPTS} -Djava.security.egd=file:/dev/./urandom -jar /code/*.jar
```
murdos, line 17: Out of curiosity, what is the purpose of the `exec` command here?
Would it work differently without it?

pascalgrimaud: using exec here, it will force the java process to be the main one with PID value to 1, that's the bug we had :)

PR #10673, opened 2024-08-25
Vue: meta description is missing

qmonmert:
image

### DIFF
--- a/src/main/resources/generator/client/vue/webapp/index.html.mustache
+++ b/src/main/resources/generator/client/vue/webapp/index.html.mustache
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="description" content="JHipster Lite project" />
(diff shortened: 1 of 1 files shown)


src/main/resources/generator/client/vue/webapp/index.html.mustache
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="description" content="JHipster Lite project" />
```
murdos, line 5: Since this is for a generated app, it should not contains any reference to JHipster Lite (in keeping with https://github.com/jhipster/jhipster-lite/issues/8654).
Using `{{ projectName }}` should be more appropriate.

qmonmert: I did like for Angular and React, we can update the 3

murdos: Yeah, I think we should update the 3 :)

PR #10679, opened 2024-08-26
Frontend: update meta description

qmonmert:
maybe also update `<title>` with `{{ projectName }}`? @murdos @pascalgrimaud


murdos:
@qmonmert : Good catch, `<title>` should indeed be updated too! :)

murdos (approved):

qmonmert:
@murdos I will do `<title>` in an other PR because we need to update e2e tests

PR #10690, opened 2024-08-28
add vue internationalization

fabienpuissant:
create vue internalization module using i18next 23.14.0 with language detector see https://i18next.github.io/i18next-vue/introduction.html


fabienpuissant:
setupTest to be done in the react module in order to not mock i18next on each test

src/main/resources/generator/client/vue/i18n/src/main/webapp/app/i18n.ts
```diff
@@ -0,0 +1,22 @@
+import i18n from 'i18next';
+
+import Backend from 'i18next-http-backend';
+import LanguageDetector from 'i18next-browser-languagedetector';
+
+i18n
```
DamnClin, line 6: I like to be able to split trads in multiple files to ease maintenance. Is it possible to do that here?

fabienpuissant: I agree, this is not a native feature, but now we can use Lazy load in memory translations (https://www.i18next.com/how-to/add-or-load-translations#lazy-load-in-memory-translations). 

In the old school way, in older projects, I made a script that flatten all the translation files into the translation.json

Do you think I should use the the plugin ?

DamnClin: There are simpler ways to do it I think:

```typescript
export const buildI18n = () => {
  return createI18n({
    locale: 'fr',
    messages: getLocaleMessages(),
  });
};

const sources = {
  fr: import.meta.glob('@/locales/fr/*.json', {
    eager: true,
    import: 'default',
  }),
};

const getLocaleMessages = () =>
  Object.keys(sources).reduce(
    (messages, locale) => ({
      ...messages,
      [locale]: Object.values(sources[locale]).reduce((message, current) => ({ ...(message ?? {}), ...(current ?? {}) }), {}),
    }),
    {}
  );

```

fabienpuissant: Following your idea, I think it's a good point to use namespaces so that we can have clearly organization of the trads 😄 

Btw Backend was useless and in conflict with locales resources so I removed it

I need to do this on the React module as well

fabienpuissant: Eventually, we will not use namespace because it's too tidious to write namespace on each translation key.
We have translation files for each bounded context which is way more convenient.
All tribute to @Gnuk 😃

src/main/resources/generator/client/common/i18n/Translations.ts
```diff
@@ -0,0 +1,33 @@
...
+export type Translations = Record<string, Translation>;
+
+const toLanguage = ([key, value]: [string, ResourceKey]): [string, ResourceLanguage] => [
+  key,
+  {
+    translation: value,
+  },
+];
+
+const mergeTranslations = (translations: Array<Translations>): Translations =>
```
Gnuk, line 13: I see more usage of `T[]` than `Array<T>`, I don't know if it's the most used. But the interesting thing with brackets is that it's easy to declare both a tuple `[T, U]` and a classic array. So it's not directly related to this PR, but, is there a common usage for Jhipster about it ?

fabienpuissant: They are same, but Array<> prevent potential issues when using key of in the array. But as the Jh lite convention seems to be T[], I replace them with this

DamnClin (approved):
Totally forgot this one... sorry.

LGTM

fabienpuissant:
@murdos 
rebased and adapted but seems ready to go

src/test/java/tech/jhipster/lite/generator/client/vue/i18n/domain/VueI18nModuleFactoryTest.java
```diff
@@ -0,0 +1,180 @@
...
+      vitest()
+    );
+    asserter
+      .hasFile("package.json")
+      .containing(nodeDependency("i18next"))
+      .containing(nodeDependency("i18next-vue"))
+      .containing(nodeDependency("i18next-browser-languagedetector"))
+      .and()
+      .hasFile("src/main/webapp/app/i18n.ts")
+      .containing(
```
murdos, line 37: There's no real need to check the content of your templates.
It's more useful when you're modifying an existing file.

murdos (commented):
@Gnuk : do you have some time to review this PR?

src/test/java/tech/jhipster/lite/generator/client/vue/i18n/domain/VueI18nModuleFactoryTest.java
fabienpuissant: Ok then I removed also in the i18n files related since they are not templated we just load the file into the app (fr.ts, en.ts, Translations.ts, test file)

murdos:
Thanks @fabienpuissant !

pascalgrimaud:
Adding a bounty on this nice feature, and to thanks the work for https://github.com/jhipster/jhipster-lite/pull/10639 too

fabienpuissant:
Thank's :)

Bounty claimed : https://opencollective.com/generator-jhipster/expenses/222013

pascalgrimaud:
@fabienpuissant : approved

PR #10714, opened 2024-08-31
chore: simplify cucumber module steps

murdos:
Rather than handling manually the package.json in cucumber tests, we can just apply the init module. This is more robust and easier to maintain.


renanfranca (approved):
LGTM ;)

Thank you for your work @murdos

PR #10721, opened 2024-09-01
Disable virtual threads by default on generated Spring app

### LINKED ISSUE #10474: Disable virtual threads by default on generated Spring app
There are some edge cases where the usage of java 21 virtual threads is causing issues, so it's better to turn it off, as done by Spring Boot.

Harsh4902:
- turned off virtual thread in by changing `spring.threads.virtual.enabled` to `false` in `application.yaml`
- Fixes #10474

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/core/domain/SpringBootCoreModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/core/domain/SpringBootCoreModuleFactory.java
@@ -78,7 +78,7 @@ public class SpringBootCoreModuleFactory {
         .and()
       .springMainProperties()
         .set(propertyKey("spring.application.name"), propertyValue(baseName))
-        .set(propertyKey("spring.threads.virtual.enabled"), propertyValue(true))
+        .set(propertyKey("spring.threads.virtual.enabled"), propertyValue(false))
         .set(propertyKey(basePackageLoggingLevel), propertyValue("INFO"))
         .and()
       .springLocalProperties()
--- a/src/main/resources/config/application.yml
+++ b/src/main/resources/config/application.yml
@@ -78,7 +78,7 @@ server:
 spring:
   threads:
     virtual:
-      enabled: true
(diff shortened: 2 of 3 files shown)


src/main/resources/config/application.yml
```diff
@@ -78,7 +78,7 @@ server:
 spring:
   threads:
     virtual:
-      enabled: true
+      enabled: false
```
DamnClin, line 81: This is the configuration for the generator (the app you start locally). The goal here was to disable that for generated applications so update in `SpringBootCoreModuleFactory`

Harsh4902: Changed value of  `spring.threads.virtual.enabled` to `false` in `SpringBootCoreModuleFactory`.

src/main/java/tech/jhipster/lite/generator/server/springboot/core/domain/SpringBootCoreModuleFactory.java
```diff
@@ -78,7 +78,7 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
         .and()
       .springMainProperties()
         .set(propertyKey("spring.application.name"), propertyValue(baseName))
-        .set(propertyKey("spring.threads.virtual.enabled"), propertyValue(true))
+        .set(propertyKey("spring.threads.virtual.enabled"), propertyValue(false))
```
murdos, line 81: In fact we should just remove the property, in order to use default Spring boot behaviour

murdos (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/core/domain/SpringBootCoreModuleFactory.java
Harsh4902: Oh, ok.

Harsh4902:
@murdos I am closing this PR, will raise a fresh PR with minimal commit.

PR #10755, opened 2024-09-04
Replace @import by @use inside Tikui style

### LINKED ISSUE #10754: Replace @import to @use
SCSS discourage use of `@import` : https://sass-lang.com/documentation/at-rules/import/

Most of the time it's possible to replace  `@import` to `@use` (with some adaptations) in `SCSS`  files.

qmonmert:
Fix #10754


renanfranca (approved):
Thank you for your work @Gnuk ! LGTM ;)

Good to know that it's possible to use `@use` without explicitly setting an alias with `as name`

PR #10781, opened 2024-09-07
handle optional in Jackson

### LINKED ISSUE #10751: Handle Optional in Jackson
Add jdk8Module

```
@Configuration
class JacksonConfiguration {

  @Bean
  public Module jdk8Module() {
    return new Jdk8Module();
  }
}
```

And configuration:

```
spring.jackson.default-property-inclusion: non_absent
```

Updates to do in `SpringBootCoreModuleFactoryTest` and `SpringBootCoreModuleFactory`

fabienpuissant:
create bean Jdk8Module and ObjectMapper

Fix #10751

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
@@ -23,6 +23,10 @@ public class SpringBootMvcsModulesFactory {
   private static final JHipsterSource MAIN_SOURCE = SOURCE.append("main");
   private static final JHipsterSource TEST_SOURCE = SOURCE.append("test");
 
+  private static final JHipsterSource JACKSON_MAIN_SOURCE = from("server/springboot/jackson/main");
+  private static final JHipsterSource JACKSON_TEST_SOURCE = from("server/springboot/jackson/test");
+  private static final String WIRE_JACKSON_CONFIG = "wire/jackson/infrastructure/primary";
+
   private static final GroupId SPRING_BOOT_GROUP = groupId("org.springframework.boot");
   private static final ArtifactId STARTER_WEB_ARTIFACT_ID = artifactId("spring-boot-starter-web");
 
@@ -84,11 +88,14 @@ public class SpringBootMvcsModulesFactory {
         .and()
       .springMainProperties()
         .set(SERVER_PORT, propertyValue(properties.serverPort().get()))
+        .set(propertyKey("spring.jackson.default-property-inclusion"), propertyValue("non_absent"))
         .and()
       .springTestProperties()
         .set(SERVER_PORT, propertyValue(0))
         .and()
       .files()
+        .add(JACKSON_MAIN_SOURCE.append(WIRE_JACKSON_CONFIG).template("JacksonConfiguration.java"), toSrcMainJava().append(packagePath).append(WIRE_JACKSON_CONFIG).append("JacksonConfiguration.java"))
+        .add(JACKSON_TEST_SOURCE.append(WIRE_JACKSON_CONFIG).template("JacksonConfigurationIT.java"), toSrcTestJava().append(packagePath).append(WIRE_JACKSON_CONFIG).append("JacksonConfigurationIT.java"))
         .add(SOURCE.file("resources/404.html"), to("src/main/resources/public/error/404.html"))
         .batch(MAIN_SOURCE.append(CORS), mainDestination.append(CORS_PRIMARY))
           .addTemplate("CorsFilterConfiguration.java")
--- a/src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactoryTest.java
+++ b/src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactoryTest.java
@@ -103,6 +103,9 @@ class SpringBootMvcsModulesFactoryTest {
         """
         server:
           port: 9000
+        spring:
+          jackson:
+            default-property-inclusion: non_absent
         """
       )
       .and()
@@ -145,6 +148,10 @@ class SpringBootMvcsModulesFactoryTest {
             </dependency>
         """
       )
+      .and()
+      .hasFile("src/main/java/tech/jhipster/jhlitest/wire/jackson/infrastructure/primary/JacksonConfiguration.java")
+      .and()
+      .hasFile("src/test/java/tech/jhipster/jhlitest/wire/jackson/infrastructure/primary/JacksonConfigurationIT.java")
       .and();
   }
 }
--- a/src/main/java/tech/jhipster/lite/generator/server/springboot/webflux/web/domain/SpringBootWebfluxModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/server/springboot/webflux/web/domain/SpringBootWebfluxModuleFactory.java
@@ -14,7 +14,9 @@ import tech.jhipster.lite.shared.error.domain.Assert;
 public class SpringBootWebfluxModuleFactory {
 
   private static final JHipsterSource SOURCE = from("server/springboot/webflux/web");
-
+  private static final JHipsterSource JACKSON_MAIN_SOURCE = from("server/springboot/jackson/main");
+  private static final JHipsterSource JACKSON_TEST_SOURCE = from("server/springboot/jackson/test");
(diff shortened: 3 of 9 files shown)


fabienpuissant:
@DamnClin 
Jackson was not available in the spring boot core module.
In a first version I've just imported jdk8Module from jackson-datatype-jdk8 but in the end, I preferred use the existing starter web

I don't know what do you think about it

fabienpuissant:
> @DamnClin Jackson was not available in the spring boot core module. In a first version I've just imported jdk8Module from jackson-datatype-jdk8 but in the end, I preferred use the existing starter web
> 
> I don't know what do you think about it

Importing the starter-web leads to side effects (especially with Spring cloud), so I just import jackson-datatype-jdk8

DamnClin:
> @DamnClin Jackson was not available in the spring boot core module. In a first version I've just imported jdk8Module from jackson-datatype-jdk8 but in the end, I preferred use the existing starter web
> 
> I don't know what do you think about it

This is probably the wrong module to add it so :). There should be a module with this dependency already on and his must be done here.

We can't add the web starter in this module if it wasn't there since SpringBoot can be use for other needs (like batch), my bad on identifying the sweet spot

src/main/resources/generator/dependencies/pom.xml
```diff
@@ -28,6 +28,7 @@
     <liquibase.version>4.29.2</liquibase.version>
     <reflections.version>0.10.2</reflections.version>
     <json-web-token.version>0.12.6</json-web-token.version>
+    <jackson-datatype-jdk8.version>2.17.2</jackson-datatype-jdk8.version>
```
murdos, line 31: You can remove the changes from this file, they're not used anymore

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
```diff
@@ -23,6 +23,10 @@ public class SpringBootMvcsModulesFactory {
   private static final JHipsterSource MAIN_SOURCE = SOURCE.append("main");
   private static final JHipsterSource TEST_SOURCE = SOURCE.append("test");
 
+  private static final JHipsterSource JACKSON_MAIN_SOURCE = from("server/springboot/jackson/main");
```
murdos, line 26: You should put the new templates in `server/springboot/mvc/web` (the root of the resources for this module), and thus use below the `SOURCE` constant.

We should try to keep a coherent tree directory between modules and their resources.

src/main/resources/generator/server/springboot/jackson/test/wire/jackson/infrastructure/config/JacksonConfigurationTest.java.mustache
```diff
@@ -0,0 +1,23 @@
...
+class JacksonConfigurationTest {
+
+  private final JacksonConfiguration jacksonConfiguration = new JacksonConfiguration();
+
+  @Test
+  void shouldHandleOptional() throws JsonProcessingException {
+    Optional<String> optional = Optional.of("test");
+    ObjectMapper objectMapper = new ObjectMapper();
+    objectMapper.registerModule(jacksonConfiguration.jdk8Module());
+    assertThat(objectMapper.writeValueAsString(optional)).isEqualTo("\"test\"");
```
murdos, line 21: Here you're just testing the `Jdk8Module` which is an external dependency, so there's no real reason to test it again.
What could be more interesting is an integration test where you inject the ObjectMapper provided by Spring Boot, and assert that it indeed correctly map Optional.

murdos (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
fabienpuissant: even if we use this templates for mvc and webflux ?
I put it in a separate module because of that

murdos: Ah, sorry, my review was a bit too quick :)
Then it's fine I guess

murdos (approved):

fabienpuissant:
To avoid errors, use default spring boot jackson configuration and enrich with jdk8Module
Source : https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java

src/main/resources/generator/server/springboot/jackson/main/wire/jackson/infrastructure/config/JacksonConfiguration.java.mustache
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+
+@Configuration
+class JacksonConfiguration {
+
+  @Bean
+  public ObjectMapper objectMapper(Jackson2ObjectMapperBuilder builder) {
```
murdos, line 13: I've just tested it, the following code is sufficient and more extensible (since it doesn't overwrite SpringBoot configuration):

```java
@Bean
  Jdk8Module jdk8Module() {
    return new Jdk8Module();
  }
```

murdos: It will be injected in the list of modules here: https://github.com/spring-projects/spring-boot/blob/74ddea416b8a0d2c12e289116e75aace3c70acf2/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java#L179

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
```diff
@@ -23,6 +23,10 @@ public class SpringBootMvcsModulesFactory {
...
+  private static final JHipsterSource JACKSON_TEST_SOURCE = from("server/springboot/jackson/test");
+  private static final String WIRE_JACKSON_CONFIG = "wire/jackson/infrastructure/config";
```
murdos, line 28: Subfolder of `infrastructure` folder are either `primary` or `secondary`:
```suggestion
  private static final String WIRE_JACKSON_CONFIG = "wire/jackson/infrastructure/primary";
```

src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactoryTest.java
```diff
@@ -145,6 +148,11 @@ private JHipsterModuleAsserter assertMvcModule(JHipsterModule module) {
             </dependency>
         """
       )
+      .and()
+      .hasFile("src/main/java/tech/jhipster/jhlitest/wire/jackson/infrastructure/config/JacksonConfiguration.java")
+      .containing("public ObjectMapper objectMapper(Jackson2ObjectMapperBuilder builder)")
```
murdos, line 153: Usually we test content of a file when it's a template and we want to check the dynamic content.
Here it's not needed since this is a static content handled in the template.

src/main/resources/generator/server/springboot/jackson/main/wire/jackson/infrastructure/config/JacksonConfiguration.java.mustache
fabienpuissant: true xD
Kinda dark magic

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/web/domain/SpringBootMvcsModulesFactory.java
fabienpuissant: image

murdos: You need the update the JacksonConfiguration.java content too:
`package {{packageName}}.wire.jackson.infrastructure.config;` -> `package {{packageName}}.wire.jackson.infrastructure.primary;`

murdos (approved):

Issue #10795, opened 2024-09-08
react core build && dev error


DamnClin:
On main, when running `npm run build` on a `react-core` application got:

```sh
npm run build

> jhipster-sample-application@0.0.0 build
> tsc && vite build --emptyOutDir

✘ [ERROR] "vite-tsconfig-paths" resolved to an ESM file. ESM file cannot be loaded by `require`. See https://vitejs.dev/guide/troubleshooting.html#this-package-is-esm-only for more details. [plugin externalize-deps]

    node_modules/esbuild/lib/main.js:1225:27:
      1225 │         let result = await callback({
           ╵                            ^

    at file:///tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/vite/dist/node/chunks/dep-BaOMuo4I.js:66566:23
    at requestCallbacks.on-resolve (/tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/esbuild/lib/main.js:1225:28)
    at handleRequest (/tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/esbuild/lib/main.js:647:17)
    at handleIncomingPacket (/tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/esbuild/lib/main.js:672:7)
    at Socket.readFromStdout (/tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/esbuild/lib/main.js:600:7)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at Pipe.onStreamRead (node:internal/stream_base_commons:191:23)

  This error came from the "onResolve" callback registered here:

    node_modules/esbuild/lib/main.js:1150:20:
      1150 │       let promise = setup({
           ╵                     ^

    at setup (file:///tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/vite/dist/node/chunks/dep-BaOMuo4I.js:66528:18)
    at handlePlugins (/tmp/9dfbbae6-b568-45d1-bd38-e70d29cdb962/node_modules/esbuild/lib/main.js:1150:21)
[...]

PR #10796, opened 2024-09-08
Add run all in scripts

DamnClin:
Preparing tikui module by adding the possibility to have multiple executions for each npm script target.


DamnClin:
Need https://github.com/jhipster/jhipster-lite/issues/10795

src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java
```diff
@@ -66,13 +66,16 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
         .addDevDependency(packageName("jest-preset-angular"), ANGULAR)
         .addDevDependency(packageName("jest-sonar-reporter"), ANGULAR)
         .addDevDependency(packageName("typescript"), ANGULAR)
+        .addDevDependency(packageName("npm-run-all"), COMMON)
```
murdos, line 69: Please use `npm-run-all2`, `npm-run-all` is not maintained

src/main/java/tech/jhipster/lite/generator/client/react/core/domain/ReactCoreModulesFactory.java
```diff
@@ -59,15 +59,19 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
         .addDevDependency(packageName("vitest-sonar-reporter"), COMMON)
         .addDependency(packageName("react"), REACT)
         .addDependency(packageName("react-dom"), REACT)
-        .addScript(scriptKey("dev"), scriptCommand("vite"))
-        .addScript(scriptKey("build"), scriptCommand("tsc && vite build --emptyOutDir"))
+        .addDependency(packageName("npm-run-all"), COMMON)
+        .addScript(scriptKey("dev"), scriptCommand("npm-run-all dev:*"))
+        .addScript(scriptKey("dev:vite"), scriptCommand("vite"))
+        .addScript(scriptKey("build"), scriptCommand("npm-run-all build:*"))
+        .addScript(scriptKey("build:vite"), scriptCommand("tsc && vite build --emptyOutDir"))
```
murdos, line 66: This could be split with this new approach: `build:tsc` and `build:vite`

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
```diff
@@ -73,13 +73,18 @@ public JHipsterModule buildVueModule(JHipsterModuleProperties properties) {
         .addDevDependency(packageName("vue-tsc"), VUE)
         .addDevDependency(packageName("@types/sinon"), VUE)
         .addDevDependency(packageName("sinon"), VUE)
-        .addScript(scriptKey("build"), scriptCommand("vue-tsc -p tsconfig.build.json --noEmit && vite build --emptyOutDir"))
-        .addScript(scriptKey("dev"), scriptCommand("vite"))
+        .addDevDependency(packageName("npm-run-all"), COMMON)
+        .addScript(scriptKey("build"), scriptCommand("npm-run-all build:*"))
+        .addScript(scriptKey("build:vue"), scriptCommand("vue-tsc -p tsconfig.build.json --noEmit && vite build --emptyOutDir"))
```
murdos, line 78: Here we could also have `build:tsc` and `build:vite`

src/main/resources/generator/dependencies/common/package.json
```diff
@@ -35,6 +35,7 @@
     "mkdirp": "3.0.1",
     "node": "20.17.0",
     "npm": "10.8.3",
+    "npm-run-all": "4.1.5",
```
murdos, line 38: https://www.npmjs.com/package/npm-run-all2
```suggestion
    "npm-run-all2": "6.2.2",
```

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
DamnClin: I don't think this is a good idea since you want to start with `tsc` and, only if this works, build

murdos: My comprehension is that this is exactly what would happen when you'll run `npm run build`: it will launch `npm run build:tsc` and then `npm run build:vite` only if the first command has succeed.

src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java
```diff
@@ -66,13 +66,16 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
         .addDevDependency(packageName("jest-preset-angular"), ANGULAR)
         .addDevDependency(packageName("jest-sonar-reporter"), ANGULAR)
         .addDevDependency(packageName("typescript"), ANGULAR)
+        .addDevDependency(packageName("npm-run-all2"), COMMON)
         .addScript(scriptKey("ng"), scriptCommand("ng"))
         .addScript(scriptKey("start"), scriptCommand("ng serve"))
-        .addScript(scriptKey("build"), scriptCommand("ng build"))
-        .addScript(scriptKey("watch"), scriptCommand("ng build --watch --configuration development"))
-        .addScript(scriptKey("test"), scriptCommand("npm run test:watch"))
+        .addScript(scriptKey("build"), scriptCommand("npm-run-all2 build:*"))
```
murdos, line 72: The command is still `npm-run-all`, `npm-run-all2` is just the name of the package ;)

src/main/java/tech/jhipster/lite/generator/client/vue/core/domain/VueModulesFactory.java
DamnClin: were does the order come from? (haven't look at it at all)

murdos: It might comes from the order of declaration, but this should probably be verified:
> npm-run-all reads the actual npm-script list from package.json in the current directory, then filters the scripts by glob-like patterns, then runs those.

murdos (approved):

qmonmert (approved):

Issue #10822, opened 2024-09-13
vue-core module and coverage


pascalgrimaud:
Following the refactoring of vue-core module by @DamnClin, there are some missing configurations, as these code should be ignored, and not be included in coverage, if we decide it's an implementation details

![image](https://github.com/user-attachments/assets/1e3b4ab9-2a9a-4303-9401-bfad26743f18)

Links:
- https://sonarcloud.io/summary/overall?id=jhipster_jhipster-lite-sample-app
- https://sonarcloud.io/component_measures?id=jhipster_jhipster-lite-sample-app&metric=coverage&view=list

See 
![image](https://github.com/user-attachments/assets/f42ff8f5-b784-4fe0-93e2-abb1e3f5ccee)

pascalgrimaud:
maybe similar to https://github.com/jhipster/jhipster-lite/issues/10752 ?

PR #10824, opened 2024-09-13
Vue: add tests for AxiosHttp

qmonmert:
Related to #10822 

image


murdos:
My feeling was that we should exclude this code from code coverage computation rather than adding tests to cover it.

qmonmert:
But these tests are presents in the generator

pascalgrimaud:
I agree with you @murdos 

But as these tests already exist, we can simply keep them, and put back the coverage to 100% in vitest config

Then, later, in another PR:
- exclude these files properly
- keep consistency between jhlite and generated app

PR #10825, opened 2024-09-13
New API for add lintstaged entries + protobuf backwards compatibility check commit hook

murdos:


murdos:
@fabienpuissant : FYI and your review is welcomed :)

fabienpuissant:
> @fabienpuissant : FYI and your review is welcomed :)

This is waaaay clearer !!

src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java
```diff
@@ -22,11 +22,7 @@ public class AngularModuleFactory {
   public JHipsterModule buildModule(JHipsterModuleProperties properties) {
     //@formatter:off
     return moduleBuilder(properties)
-      .optionalReplacements()
-        .in(path(".lintstagedrc.cjs"))
-          .add(lineBeforeRegex("\\['prettier --write'\\]"), "  '{src/**/,}*.ts': ['eslint --fix', 'prettier --write'],")
-          .and()
-        .and()
+      .preCommitActions(stagedFilesFilter("{src/**/,}*.ts"), preCommitCommands("['eslint --fix', 'prettier --write']"))
```
fabienpuissant, line 25: I wonder if it would be nice to have a PreCommitCommands enum that will contains
PreCommitCommands .LINT("['eslint --fix', 'prettier --write']") 
PreCommitCommands .PROTO_BACKWARD("() => ['mvn proto-backwards-compatibility:backwards-compatibility-check', 'git add *proto.lock']")
 instead of calling the function preCommitCommands with a string. 
This way, we'll have to refactor only the enum if there is some changes

fabienpuissant (approved):

src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java
murdos: Using an enum would prevent any other usage than the 2 existing ones in jhipster-lite, in particular in custom modules developed in custom jhlite instances.

Here I've used a simple string because lintstaged way of declaring commands is very flexible (you can use simple string, array of strings, function, ...) and I don't want to limit what we can do. We can imagine in a second time having different factory methods in the `PreCommitCommands` that would match theses different declaration and ease the declaration.

What we could do instead is introduce a constant in the PreCommitCommands record, but it introduces a coupling between a generic modelling in the domain (`PreCommitCommands`) and some actual specific usages, so I'm not convinced.

PR #10833, opened 2024-09-14
Update vue-core vitest configuration and sonar properties for front

pascalgrimaud:
- fix https://github.com/jhipster/jhipster-lite/issues/10822


qmonmert:
@pascalgrimaud I added tests for HomeRouter #10835 but if it is unused qou can close the PR

src/main/resources/generator/server/sonar/sonar-fullstack-project.properties.mustache
```diff
@@ -17,7 +17,14 @@ sonar.testExecutionReportPaths={{projectBuildDirectory}}/test-results/TESTS-resu
 sonar.javascript.lcov.reportPaths={{projectBuildDirectory}}/test-results/lcov.info
 
 sonar.sourceEncoding=UTF-8
-sonar.exclusions=src/main/webapp/main.ts, src/main/webapp/app/main.ts, src/main/webapp/content/**/*.*, {{projectBuildDirectory}}/classes/static/**/*.*, src/main/webapp/app/index.tsx
+sonar.exclusions=\
+  src/main/webapp/app/home/infrastructure/primary/*.*,\
```
DamnClin, line 21: All infrastructure code, from all contextes, should be excluded since we can't get proper metrics on that

DamnClin: In fact, since we can't merge coverrages in front it not a good idea to add a threshold. It will force people to make implementation details tests just to make CI pass

pascalgrimaud: I agree on that, but as I'm really short on time.

So I'll keep this like this:
- so I can do a patch release, fixing the deploy on GCP
- the coverage will be back at 100% on sample app
- then, someone can work on this for the next version v1.19.0
  - clean tests on infrastructure
  - update sonar properties propertly
  - keep consistency between angular / react / vue

pascalgrimaud: I'll create a dedicated ticket for that

PR #10849, opened 2024-09-15
ci-tests: Run linter on generated apps

murdos:


.github/workflows/github-actions.yml
```diff
@@ -366,7 +366,7 @@ jobs:
           if [[ "${{ steps.compare-hash.outputs.different }}" == "true" || "${{ steps.compare-hash.conclusion }}" == "skipped" ]]; then
             echo "execute_tests=true" >> $GITHUB_OUTPUT
           else
-            echo "execute_tests=false" >> $GITHUB_OUTPUT
+            echo "execute_tests=true" >> $GITHUB_OUTPUT
           fi
```
renanfranca, line 370: Should this step be removed? With this change, the tests will always be executed if the generated app code is different from or equal to the main branch.

I think there should be a specific condition related to the generated apps that runs lint. What do you think?

.github/workflows/github-actions.yml
```diff
@@ -446,6 +446,15 @@ jobs:
...
+      - name: 'Test: run linter for ${{ matrix.app }}'
+        if: steps.tests-requirement-check.outputs.execute_tests == 'true'
+        working-directory: /tmp/jhlite/${{ matrix.app }}/
+        run: |
+          if grep -q '"lint":' package.json; then
+            npm run lint
+          else
+            echo "No lint script in package.json found, skipping step."
+          fi
```
renanfranca, line 457: LGTM 😉.  My only concern is with the if condition, should it be removed?

.github/workflows/github-actions.yml
murdos: This is just for temporarily force the execution of the linter.
Otherwise build would pass without actual checking, and it could break main branches build.

I'll remove it once I've spotted existing lint issues, and fixed them :)

.github/workflows/github-actions.yml
murdos: The if should be kept, some generated app don't have client and thus linter configured.

renanfranca: Oh, I agree. I was referring to the `if` condition from the step and not the `if` inside the run. You've already clarified any doubts I had, so I'm going to leave this PR approved. 👌

renanfranca (approved):
Thank you for your work 😃👍

PR #10877, opened 2024-09-16
Move cypress home test

### LINKED ISSUE #10875: Move cypress home test
Cypress home test is generated in `src/test/webapp/component/common/primary/app/Home.spec.ts`, should be in `src/test/webapp/component/home/Home.spec.ts`

qmonmert:
Fix #10875


src/main/java/tech/jhipster/lite/generator/client/tools/cypress/domain/CypressModuleFactory.java
```diff
@@ -20,6 +20,7 @@ public class CypressModuleFactory {
   private static final JHipsterDestination CYPRESS_E2E_TESTS = to("src/test/webapp/e2e");
 
   private static final String PRIMARY_APP = "common/primary/app";
```
murdos, line 22: Is it still used? Otherwise let's remove it

qmonmert: good catch, I update

PR #10878, opened 2024-09-16
Extensible npm version source

### LINKED ISSUE #10852: Generator: NPM VersionSource should be extensible
When declaring new npm depedency in a module, you have to declare the VersionSource associated with this npm dependency.
The reason is that generator handle different client modules, and that the same dependency can be used with a different version in each client.

The issue is that the VersionSource is currently an enum in jhipster-lite, so it's not extensible. E.g. if a custom-jhlite want to handle a new frontend framework, or if it wants to use different versions from jhipster-lite.

Something similar to module and feature slugs could probably be done, in order to make it extensible.

murdos:
Fixes  #10852


renanfranca:
I am going to review it today 😉👍

murdos:
> I am going to review it today 😉👍

Thanks @renanfranca ! 
To better illustrate what I try to achieve, here's an example use on jhite-extension: https://github.com/jhipster/jhipster-lite-extension/pull/85

renanfranca:
> > I am going to review it today 😉👍
> 
> Thanks @renanfranca ! To better illustrate what I try to achieve, here's an example use on jhite-extension: [jhipster/jhipster-lite-extension#85](https://github.com/jhipster/jhipster-lite-extension/pull/85)

It helped a lot to understand, and it's pretty cool!

renanfranca (approved):
@murdos: Great work, my friend! I took some time to learn new things while reviewing, hehe!

PR #10907, opened 2024-09-19
Handle null in TypeScript optional

### LINKED ISSUE #10897: Handle null in TS Optional
We should add | null in `Optional.ts`, really helpfull

NolwennD:
Fix #10897

### DIFF
--- a/src/main/webapp/app/shared/optional/domain/Optional.ts
+++ b/src/main/webapp/app/shared/optional/domain/Optional.ts
@@ -1,3 +1,5 @@
+const checkIsNaN = <Value>(value: Value) => typeof value === 'number' && isNaN(value);
+
 export abstract class Optional<Value> {
   public static empty<Value>(): Optional<Value> {
     return new EmptyOptional();
@@ -7,11 +9,10 @@ export abstract class Optional<Value> {
     return new ValuatedOptional(value);
   }
 
-  static ofUndefinable<Value>(value: Value | undefined): Optional<Value> {
-    if (value === undefined) {
+  static ofNullable<Value>(value: Value | undefined | null): Optional<Value> {
+    if (value === undefined || value === null || checkIsNaN(value)) {
       return Optional.empty();
     }
-
     return Optional.of(value);
   }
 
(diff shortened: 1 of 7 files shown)


src/main/webapp/app/shared/optional/domain/Optional.ts
```diff
@@ -7,12 +7,11 @@ export abstract class Optional<Value> {
     return new ValuatedOptional(value);
   }
 
-  static ofUndefinable<Value>(value: Value | undefined): Optional<Value> {
-    if (value === undefined) {
-      return Optional.empty();
+  static ofUndefinable<Value>(value: Value | undefined | null): Optional<Value> {
```
murdos, line 10: We should maybe change the name of the method, or add a new method `ofNullable(..)`.
What do you think?

NolwennD: I thank about `ofEmptyable` but I am not really convinced.

DamnClin (approved):

src/main/webapp/app/shared/optional/domain/Optional.ts
murdos: How about having a single method named ofNullable? 
This would then be similar to the java Api that we're using as an aspiration.

I'm not sure we need to worry about breaking things here:
- we have control over jhlite's webapp and it's not used externally 
- in the generated Optional.ts (that is not changed in this PR by the way), people having already applied it have no reason to re-apply it

DamnClin: True (and we need to change the generated one :D)

NolwennD: I rename to `ofNullable`. Not in scope, but I added `NaN` to nullable value.

murdos:
Thanks @NolwennD

Gnuk:
Sorry, I've not seen this PR before, but I think it could be interesting to choose a better name than ofNullable like safeOf() or ofSafe(). Because in java, it's ok, there is not this problem like js with null, undefined or NaN. So that's why I think ofNullable is confusing.

PR #10935, opened 2024-09-21
Init tikui module

DamnClin:
First step for tikui module. We are working on this in background for some time with @Gnuk

After this, TODO are:

- [x] support angular (mainly proxy);
- [x] support react (mainly proxy);
- [ ] add some more style example;
- [ ] create a documentation.

The idea here is to start something in a disabled module so we can move on on this.


src/main/resources/generator/dependencies/common/package.json
```diff
@@ -46,7 +47,12 @@
     "prettier-plugin-packagejson": "2.5.2",
     "recursive-copy-cli": "1.0.20",
     "start-server-and-test": "2.0.8",
+    "stylelint": "^16.7.0",
+    "stylelint-config-concentric-order": "^5.2.0",
+    "stylelint-config-standard-scss": "^13.1.0",
+    "stylelint-order": "^6.0.4",
```
qmonmert, line 53:
```suggestion
    "stylelint": "16.7.0",
    "stylelint-config-concentric-order": "5.2.0",
    "stylelint-config-standard-scss": "13.1.0",
    "stylelint-order": "6.0.4",
```

Gnuk (approved):

renanfranca (approved):
LGTM ;) ! I noticed that Tikui's Quarkus is much more modularized than what I saw in the JHipster Lite project. I only understood the placeholder by looking at the code. I will learn more about the spacing when i have the time to apply the module.

I also observed that in the token, the value is set directly in the :root and SCSS variables are not used. I understood this to be a standard for all tokens, and if a component does not want to use the standard, it defines the value directly in the .scss. 

Do you not recommend creating SCSS variables in the style definitions of the components?

Thank you for your work! @DamnClin and @Gnuk

src/main/resources/generator/client/tikui/tikuiconfig.json.mustache
```diff
@@ -0,0 +1,7 @@
+{
+  "documentation": "tikui",
+  "src": "src/main/style",
+  "dist": "target/classes/public/style",
```
murdos, line 4: For gradle compatibility:
```suggestion
  "dist": "{{ projectBuildDirectory }}/classes/public/style",
```

murdos (commented):
Things that could also be done:
- prettier plugin for pug
- eslint plugin for pug
- amend lint-staged for css/scss =>run stylint --fix and prettier
- amend lint-staged for pug => run eslint and prettier

DamnClin:
> Do you not recommend creating SCSS variables in the style definitions of the components?

As always, in Software Development you are either in a Complicated or Complex context in Cynefin definitions so there is no one size feet all solution. It always depends on the context

DamnClin:
> Things that could also be done:
> 
>     * prettier plugin for pug
> 
>     * eslint plugin for pug
> 
>     * amend lint-staged for css/scss =>run stylint --fix and prettier
> 
>     * amend lint-staged for pug => run eslint and prettier

No idea on how to do that (haven't look at it at all) do you have any example (I can do the module no problem)

murdos:
> 
> No idea on how to do that (haven't look at it at all) do you have any example (I can do the module no problem)

BTW this can be done is a later stage, I was mainly sharing my thoughts while reading the PR.

amending lint-staged can now be done with the `preCommitActions(..)` API on JHipsterModuleBuilder. It can however just add new entry, not altering existing one - and that would be required here for removing css and scss from the main entry added by prettier.

Prettier plugin for pug, besides adding the npm dependency, will require patching the .prettierrc to declare the plugin

Regarding eslint plugin for pug I don't know the details, but I know that it's used on the code of jhlite.

murdos (approved):

DamnClin:
> > No idea on how to do that (haven't look at it at all) do you have any example (I can do the module no problem)
> 
> BTW this can be done is a later stage, I was mainly sharing my thoughts while reading the PR.
> 
> amending lint-staged can now be done with the `preCommitActions(..)` API on JHipsterModuleBuilder. It can however just add new entry, not altering existing one - and that would be required here for removing css and scss from the main entry added by prettier.
> 
> Prettier plugin for pug, besides adding the npm dependency, will require patching the .prettierrc to declare the plugin
> 
> Regarding eslint plugin for pug I don't know the details, but I know that it's used on the code of jhlite.

I'll create dedicated issues for those points

DamnClin:
Updated proxy for angular, cc @qmonmert  is this the "angular" way to do that?

pascalgrimaud:
Adding a bounty to reward the big work done here

DamnClin:
Claimed https://opencollective.com/generator-jhipster/expenses/222055, thanks a lot @pascalgrimaud

pascalgrimaud:
Approved :-)

PR #10948, opened 2024-09-22
Generator: add missing test for RestManagementRepository

qmonmert:


qmonmert:
@pascalgrimaud ok to merge this PR and #10949 ?

murdos:
@qmonmert : these tests are really testing implementation details.
I'd like to not require them by using code coverage from component tests instead: https://github.com/jhipster/jhipster-lite/pull/10950

pascalgrimaud:
I agree : `these tests are really testing implementation details`

But should we totally exclude what is in `infrastructure` folder ? But in this case, we can write bad + wrong code in infrastructure, so Cypress will become mandatory to protect that.

My small opinion (and probably wrong) would be:
- we can merge this one
- it's already done and won't take time to maintain
- for real projects, the user will be free to ignore infrastructure folder or not

qmonmert:
@murdos ok to merge this?

murdos:
Go ahead.
When I'm going to finish https://github.com/jhipster/jhipster-lite/pull/10950, I'll evaluate if I remove it or not.

PR #10957, opened 2024-09-23
Fix invalid if condition in Github Actions

leonard84:
The value of an `if:` entry is already an expression block, so usage of `${{ }}` is not necessary, except in one rare case (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution).
However, it is easy to accidentally create an invalid condition:
`${{ matrix.java-build-tool }} == 'maven'` will always evaluate to true, due to a quirk of how nested expression context are evaluated.
Correct is either
- `${{ matrix.java-build-tool  == 'maven' }}`
- `matrix.java-build-tool == 'maven'`

I've choosen to use the latter as it is simpler. 
I've also cleaned up unnecessary use of `${{ }}` in other if blocks.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -371,7 +371,7 @@ jobs:
       - name: 'Test: list ${{ matrix.app }}'
         if: |
           steps.tests-requirement-check.outputs.execute_tests == 'true' &&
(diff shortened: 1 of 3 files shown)


murdos:
Thanks @leonard84 for the fix and the explanation.
Much appreciated!

PR #10971, opened 2024-09-24
move preset logic to "module" bounded context

### LINKED ISSUE #10651: Preset concept should be moved to the "module" bounded context
A `preset` is collection of preselected modules that could be applied together, so it rather belong to the "module" bounded context.

It actually resides in the "project" bounded context, whereas this bounded context deals with an existing generated project: history of applied modules, download of the project, ...

renanfranca:
Fix #10651


murdos (approved):
Thanks @renanfranca

PR #11003, opened 2024-09-28
Angular: apply prettier on angularoauth2app

qmonmert:


src/main/java/tech/jhipster/lite/generator/client/angular/security/oauth2/domain/AngularOauth2ModuleFactory.java
```diff
@@ -56,14 +56,11 @@ public class AngularOauth2ModuleFactory {
   private static final String TEST_IMPORTS =
     """
     import { By } from '@angular/platform-browser';
-    import LoginComponent from './login/login.component';
-    """;
+    import LoginComponent from './login/login.component';""";
```
murdos, line 59: FYI you can also add a backslash for avoiding line break at end:
```suggestion
    import LoginComponent from './login/login.component';\
""";
```

murdos (approved):

Issue #11002, opened 2024-09-28
Enhance React i18n module


fabienpuissant:
To follow the idea of what have been done in the VueI18nModule, the React one module needs the be adapted.

- Use a setup file to mock react i18n for all tests
- Separate translations to be scoped to each bounded context in separate translation files

fabienpuissant:
@DamnClin @Gnuk 

We need to decide what we want for front translation design (https://github.com/jhipster/jhipster-lite/issues/11111)

Vue Ts configuration with multiple bounded context files

React one json file with all translations

Because rn, React and Vue are not aligned and I would like to make the work the same way 😃

Gnuk:
I think you can split translations by context. If it's easier to be consistent with JSON, we can start with it and maybe later (in the future) we will be able to use typescript type system to check translation keys.

PR #11010, opened 2024-09-29
rework react core like vue core

fabienpuissant:
In order to rework react i18n module, it would be nice to rework the react core module like @DamnClin did for vue core module
Because translation typescript files will be shared between react and vue

=> home context with the component into the primary infrastructure

Obviously impacting react-i18n and react-jwt module

Prerequisites to #11002


murdos (approved):

murdos:
Looks fine.
@pascalgrimaud : do you have any objections? This is somehow a breaking change, but I think it's a good idea to harmonise our clients.

DamnClin (commented):
Maybe add piqure for dependencies injection

DamnClin (approved):

PR #11016, opened 2024-09-29
Move 'base64' and 'slug' shared kernels to 'shared' top-level package

murdos:
Also fixes a few misqualified contexts: maven, gradle and ts-loader

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/MavenModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/MavenModuleFactory.java
@@ -1,6 +1,13 @@
 package tech.jhipster.lite.generator.buildtool.maven.domain;
 
-import static tech.jhipster.lite.module.domain.JHipsterModule.*;
+import static tech.jhipster.lite.module.domain.JHipsterModule.artifactId;
+import static tech.jhipster.lite.module.domain.JHipsterModule.from;
+import static tech.jhipster.lite.module.domain.JHipsterModule.groupId;
+import static tech.jhipster.lite.module.domain.JHipsterModule.javaDependency;
+import static tech.jhipster.lite.module.domain.JHipsterModule.mavenPlugin;
+import static tech.jhipster.lite.module.domain.JHipsterModule.moduleBuilder;
+import static tech.jhipster.lite.module.domain.JHipsterModule.pluginExecution;
+import static tech.jhipster.lite.module.domain.JHipsterModule.to;
 
 import tech.jhipster.lite.module.domain.JHipsterModule;
 import tech.jhipster.lite.module.domain.file.JHipsterSource;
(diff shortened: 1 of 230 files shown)


src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/MavenModuleFactory.java
```diff
@@ -1,6 +1,13 @@
...
 
-import static tech.jhipster.lite.module.domain.JHipsterModule.*;
+import static tech.jhipster.lite.module.domain.JHipsterModule.artifactId;
+import static tech.jhipster.lite.module.domain.JHipsterModule.from;
+import static tech.jhipster.lite.module.domain.JHipsterModule.groupId;
+import static tech.jhipster.lite.module.domain.JHipsterModule.javaDependency;
+import static tech.jhipster.lite.module.domain.JHipsterModule.mavenPlugin;
+import static tech.jhipster.lite.module.domain.JHipsterModule.moduleBuilder;
+import static tech.jhipster.lite.module.domain.JHipsterModule.pluginExecution;
+import static tech.jhipster.lite.module.domain.JHipsterModule.to;
```
renanfranca, line 10: @murdos : The expand import settings are defined in IntelliJ IDEA, or are they handled by Prettier?

renanfranca (approved):
Thank you for your work, @murdos !

src/main/java/tech/jhipster/lite/generator/buildtool/maven/domain/MavenModuleFactory.java
renanfranca: For example, in this commit, it was changed to * instead of staying expanded. I wish there were a way to make this setting the same for everyone.

murdos: Sorry, I had to expand them, otherwise Intellij was not able correctly update imports for feature and module slugs :/

There's a way to uniform imports: we can add a checkstyle rule that forbid `*` imports.

renanfranca: It's cool! Do you use the same configuration as mine?


I like this configuration. It expands when there are fewer and put * when there are too many.  What do you think about only adding these instructions at CONTRIBUTION.md?

murdos: Honestly, I don't like * at all. 
It's often a source of conflicts when multiple developers touch the same file on different branches, and it's not explicit.
I usually forbid them on my projects.

renanfranca: I can follow your lead as every IDEA collapses the import sections.  Let's do the checkstyle configuration l😄😉

Issue #9235, opened 2024-03-15
Jooq module


murdos:
Jooq is a nice alternative to JPA.
It would be great if jhlite supports it.

If integration of jooq-generator (as build plugin) is possible with liquibase/flyway, that would be awesome.

fabienpuissant:
Started to work on it 
Configuration seems ok: https://github.com/fabienpuissant/jhipsterlite-jooq/tree/postgres-liquibase

Need to find relevant example to initialize module

Let me know if you use the plugin that way

murdos:
@fabienpuissant : exemple repo looks fine!

fabienpuissant:
Not convinced about what to do as an axample for the initialization 

Only the configuration without any example is enough ?

Or maybe should I keep a CRUD repository example without the infrastructure primary to avoid RestController ?

murdos:
Jooq module(s) should not contain any sample code, only all the setup.
If you want you can also provide an additional sample module, you can look there are already some.

murdos:
Regarding modules organisation, here's a proposal:
- rename existing postgresql to jpa-postgresl, MySQL to jpa-mysql, ...
- have a feature 'jooq', with modules jooq-postgresql, jooq-mysql, ...
- have another feature 'jooq-codegen-liquibase' with modules for postgresql, MySQL
- and have a feature 'jooq-codegen-flyway'

WDYT?

fabienpuissant:
I was thinking about that since hibernate config was loaded by default
Separate the modules this way seems to be a good idea 

But I don't get last two points, I would include testcontainers codegen by default in jooq-xxxxx module no ?

As I understood, jooq need a codegen anyway. So you want to separate the default codegen and the testcontainer codegen in two modules ?

murdos:
> But I don't get last two points, I would include testcontainers codegen by default in jooq-xxxxx module no ?

If you include testcontainers codegen by default in jooq-xxxxx, it means you already have chosen between liquibase and flyway?
Or I'm missing something?

> As I understood, jooq need a codegen anyway. So you want to separate the default codegen and the testcontainer codegen in two modules ?

Indeed, I would separate things. So you can use `jooq-postgresql` module, and manually use the official jooq codegen maven plugin rather than the testcontainers one (that only support postgres, mysql and mariadb: no support for mssql e.g.).

[...]

PR #11017, opened 2024-09-29
Group existing postgresl, mysql, mariadb and mssql modules into a 'jpa' module

murdos:
Preparing the arrival of Jooq, see #9235

Breaking: renamed slugs to add the prefix 'jpa-'


renanfranca (approved):
LGTM ;) . Only the cucumber test error in the pipeline, but it's related to typescript, I guess. JpaModule makes it better organized 😃

PR #11035, opened 2024-10-02
Upgrade sonarqube v10.7.0

qmonmert:
Fix #11034

### DIFF
--- a/src/main/resources/generator/server/sonar/sonar-project.properties.mustache
+++ b/src/main/resources/generator/server/sonar/sonar-project.properties.mustache
@@ -20,8 +20,10 @@ sonar.exclusions=\
   {{projectBuildDirectory}}/classes/static/**/*.*,\
   src/main/webapp/app/index.tsx,\
   src/main/webapp/main.ts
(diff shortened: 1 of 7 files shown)


src/main/resources/generator/server/sonar/sonar-project.properties.mustache
```diff
@@ -79,3 +81,7 @@ sonar.issue.ignore.multicriteria.S6437.ruleKey=java:S6437
 # Rule: Run container as default user
 sonar.issue.ignore.multicriteria.S6471.resourceKey=src/main/docker/cassandra/**/*
 sonar.issue.ignore.multicriteria.S6471.ruleKey=docker:S6471
+
+# Rule: Methods should not contain selector parameters
+sonar.issue.ignore.multicriteria.S2301.resourceKey=src/main/webapp/**/*
```
murdos, line 86: I don't think we should disable this valid rule on our whole source code: https://sonarsource.github.io/rspec/#/rspec/S2301
Either we fix the code, or we specifically disable it for the problematic code.

PR #11061, opened 2024-10-05
fix(arch-tests): allow wire packages to depend on shared kernels

murdos:


src/main/resources/generator/server/javatool/archunit/test/HexagonalArchTest.java.mustache
```diff
@@ -25,25 +24,25 @@ import org.springframework.stereotype.Controller;
 class HexagonalArchTest {
 
   private static final String ROOT_PACKAGE = "{{packageName}}";
-  private static final String GENERATION_SHARED_KERNEL_PACKAGES = ROOT_PACKAGE.concat(".shared.generation..");
+  private static final String SHARED_PACKAGES = ROOT_PACKAGE.concat(".shared..");
```
DamnClin, line 27: I think we should stick to annotation for that since it's usefull to have shared packages in other locations (especially for business shared packages)

murdos: Agree. I've adjusted the code.

DamnClin (approved):

PR #11088, opened 2024-10-09
add `vuejwtapp` and `vueoauth2app` at the matrix in `github-actions.yml`

### LINKED ISSUE #897: Vue: account with authentication JWT
**TO-DO**
- [x] Create a sample project that adds the jwt authentication to the `vue-core` module and shows it working with login and logout
- [x] https://github.com/jhipster/jhipster-lite/pull/11047
- [ ] Add `vue-jwt` and `vueoauth2app` at the matrix in `github-actions.yml`. Delete the `vueapp` from `generate.sh`. Must be done in different pull request (I can't update the matrix before merge the `generate.sh` with the new `vuejwtapp` and `vueoauth2app`. The reason is that there is a step which tries to generate these two new apps using the `generate.sh` from the main branch) -> https://github.com/jhipster/jhipster-lite/pull/11088

renanfranca:
Fix #897 (roadmap at description)

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -246,7 +246,8 @@ jobs:
           - neo4japp
           - angularapp
           - reactapp
-          - vueapp
+          - vuejwtapp
+          - vueoauth2app
           - kafkaapp
           - pulsarapp
           - reactiveapp
@@ -263,7 +264,7 @@ jobs:
             spring-config-format: properties
           - app: consulapp
             java-version: 22
-          - app: vueapp
+          - app: vuejwtapp
             node-version: 'latest'
           - app: fullapp
             java-build-tool: gradle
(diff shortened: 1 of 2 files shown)


renanfranca:
@pascalgrimaud, @murdos: I removed `vueapp` and created `vuejwtapp` and `vueoauth2app` because the modules `vue-jwt` and `vue-oauth2-keycloak` have files with the same names, and both couldn't be applied in the same app in `generate.sh`.

murdos (approved):
Looks fine!
@pascalgrimaud : do you have some reluctance to add more generated apps in ci?

pascalgrimaud:
@murdos : no, we can add new CI for testing new modules. If I remember well, there should be a check with md5sum or something like that. If the md5sum doesn't change, this job will stop, so in the end, we won't lost lot of time

PR #11107, opened 2024-10-12
Add jooq module with maven

fabienpuissant:
ci test not relevant because need a database up to install with this jooq plugin, cannot test on mssql. Error to connect due to url error.


pascalgrimaud:
for the CI test, you can create a new dedicated config with what you need to test jooq

Another idea would be to add it to mariadb, it should work: https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L232-L239

pascalgrimaud:
for the CI test, you can create a new dedicated config with what you need to test jooq

Another idea would be to add it to mariadb, it should work: https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L232-L239

fabienpuissant:
> for the CI test, you can create a new dedicated config with what you need to test jooq
> 
> Another idea would be to add it to mariadb, it should work: https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L232-L239

But the plugin connected to a database to introspect schema and generate classes depending on what tables exists.
Is it possible in the test ci to have a docker container up with a database ready to be connected to ?

pascalgrimaud:
Here some steps in the GitHub CI:
- we launch the containers here:
  - https://github.com/jhipster/jhipster-lite/blob/main/.github/workflows/github-actions.yml#L484
  -  it will use this [start_docker_compose.sh script](https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start_docker_compose.sh) which will launch the appropriate docker-compose file
- after that, we'll start the application here:
  - https://github.com/jhipster/jhipster-lite/blob/main/.github/workflows/github-actions.yml#L490
  - it will at least test the health management. See https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start.sh#L36

fabienpuissant:
> Here some steps in the GitHub CI:
> 
> * we launch the containers here:
>   
>   * https://github.com/jhipster/jhipster-lite/blob/main/.github/workflows/github-actions.yml#L484
>   * it will use this [start_docker_compose.sh script](https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start_docker_compose.sh) which will launch the appropriate docker-compose file
> * after that, we'll start the application here:
>   
>   * https://github.com/jhipster/jhipster-lite/blob/main/.github/workflows/github-actions.yml#L490
>   * it will at least test the health management. See https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start.sh#L36

I started to investigate on it but in fact the db needs to be up at compile time not when we start the app, so on step "Test: verify: " Regarding the workflow, I don't think we have this use case and I don't find a proper way to do it

Here an example : https://github.com/fabienpuissant/jhipster-lite/actions/runs/11622943610/job/32369374084

src/main/java/tech/jhipster/lite/module/domain/jooqplugin/JooqModuleCodegenConfiguration.java
```diff
@@ -0,0 +1,140 @@
+package tech.jhipster.lite.module.domain.jooqplugin;
```
murdos, line 1: I don't thinks this should be in `module/domain`, this is too specif.
How about moving that in the `jooq/domain`? (and remove the addition in JHipsterModule)

src/main/java/tech/jhipster/lite/generator/server/springboot/database/jooq/domain/JooqModuleFactory.java
```diff
@@ -0,0 +1,238 @@
...
+      .javaDependencies()
+      .addDependency(
+        JavaDependency.builder()
+          .groupId(groupId(ORG_POSTGRESQL))
+          .artifactId(artifactId("postgresql"))
+          .scope(JavaDependencyScope.RUNTIME)
+          .build()
+      )
+      .and()
+      .mavenPlugins()
```
murdos, line 59: NB : gradle should be supported too, by configuration the gradle plugin. See https://www.jooq.org/doc/latest/manual/code-generation/codegen-gradle/
But this must be done in a second PR. Meanwhile you can require maven as a dependency of all jooq module in JooqModuleConfiguration to prevent them from being used with gradle.

src/main/java/tech/jhipster/lite/module/domain/jooqplugin/JooqModuleCodegenConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+        <password>%s</password>
+      </jdbc>
+      <generator>
+        <database>
+          <name>%s</name>
+          <includes>.*</includes>
+          <inputSchema>%s</inputSchema>
+        </database>
+        <target>
+          <packageName>org.jooq.codegen</packageName>
```
murdos, line 41: IMO this should rather be the packageName from properties (e.g. generates classes to com.my.company rather than org.jooq.codegen)

src/test/features/server/springboot/database/jooq-mariadb.feature
```diff
@@ -0,0 +1,41 @@
...
+      | maven-java   |
+      | spring-boot  |
+      | jooq-mariadb |
+    Then I should have files in ""
+      | pom.xml |
+    And I should have files in "documentation"
+      | mariadb.md |
+    And I should have files in "src/main/docker"
+      | mariadb.yml |
+    And I should have files in "src/main/resources/config"
```
murdos, line 14: This one is already created by the `spring-boot` module, so I guess you want have a finer-grained assertion like 
```suggestion
    And I should have "jdbc:mariadb://localhost:3306/" in "src/main/resources/config/application.yml"
```

src/main/java/tech/jhipster/lite/module/domain/jooqplugin/JooqModuleCodegenConfiguration.java
fabienpuissant: yep + solve deplicate issue about database enum

murdos (dismissed):

murdos:
LGTM. @fabienpuissant : could you tease the branch to fix conflicts?

Issue #11106, opened 2024-10-12
Fail to apply typescript and vue-core in a second time


DamnClin:
Step to reproduce:

1. Create an app with `typescript` module;
2. Launch `npm install`
3. Add `vue-core` module.

Error:

```
2024-10-12T16:02:20.575+02:00  INFO 97582 --- [jhlite] [omcat-handler-0] t.j.l.m.domain.JHipsterModulesApplyer    : Apply module: vue-core
[STARTED] Preparing lint-staged...
[COMPLETED] Preparing lint-staged...
[STARTED] Running tasks for staged files...
[STARTED] .lintstagedrc.cjs — 25 files
[STARTED] {src/**/,}*.{ts,vue} — 14 files
[STARTED] *.{md,json,yml,html,css,scss,java,xml,feature} — 7 files
[STARTED] eslint --fix
[STARTED] prettier --write
[FAILED] eslint --fix [FAILED]
[FAILED] eslint --fix [FAILED]
[COMPLETED] Running tasks for staged files...
[STARTED] Applying modifications from tasks...
[SKIPPED] Skipped because of errors from tasks.
[STARTED] Reverting to original state because of errors...
[FAILED] prettier --write [KILLED]
[FAILED] prettier --write [KILLED]
[COMPLETED] Reverting to original state because of errors...
[STARTED] Cleaning up temporary files...
[COMPLETED] Cleaning up temporary files...

✖ eslint --fix:

Oops! Something went wrong! :(

ESLint: 9.12.0

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'eslint-plugin-vue' imported from /tmp/227a3783-d0a9-4f9a-b13a-3db63561a8bc/eslint.config.js
    at packageResolve (node:internal/modules/esm/resolve:854:9)
    at moduleResolve (node:internal/modules/esm/resolve:927:18)
    at defaultResolve (node:internal/modules/esm/resolve:1169:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:540:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:509:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:239:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:96:40)
    at link (node:internal/modules/esm/module_job:95:36)

✖ prettier --write failed without output (KILLED).
husky - pre-commit script failed (code 1)
[...]

PR #11142, opened 2024-10-17
Fix vitest config after prettier format to allow applying front modules

pascalgrimaud:
- related to https://github.com/jhipster/jhipster-lite/issues/11106

### DIFF
--- a/src/main/resources/generator/typescript/vitest.config.ts.mustache
+++ b/src/main/resources/generator/typescript/vitest.config.ts.mustache
@@ -29,6 +29,7 @@ export default defineConfig({
       },
       include: ['src/main/webapp/**/*.ts?(x)'],
       exclude: [
+        // keep this comment to be able to apply other front module after prettier format
         ...configDefaults.coverage.exclude as string[],
       ],
       provider: 'istanbul',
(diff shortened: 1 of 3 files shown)


src/main/resources/generator/typescript/vitest.config.ts.mustache
```diff
@@ -29,6 +29,7 @@ export default defineConfig({
       },
       include: ['src/main/webapp/**/*.ts?(x)'],
       exclude: [
+        // keep this comment to be able to apply other front module after prettier format
```
murdos, line 32: I'm not really found on this comment that will stay on generated project, just for generation purpose.
I'll try another solution and let you know.

pascalgrimaud: Agree, it was just the quickest fix in the mean time. Feel free to propose better solution, and rework the replace after lines for each front

murdos: Here's an alternative (that is not perfect either): https://github.com/jhipster/jhipster-lite/pull/11189

pascalgrimaud: yes, but it's better

Issue #10836, opened 2024-09-14
Consistency between vitest.config.ts and sonar-project.properties


pascalgrimaud:
Following recent discussion about front code and coverage, we need to be consistent between:
- vitest.config.ts
- sonar-project.properties

Then, the existing tests should be reviewed, to be kept or deleted

We need to take care about Angular, React and Vue
No need Svelte as it will be deleted soon

PR #11169, opened 2024-10-20
Consistency between vitest.config.ts and sonar-project.properties

qmonmert:
Related to #10836

### DIFF
--- a/src/main/resources/generator/server/sonar/sonar-fullstack-project.properties.mustache
+++ b/src/main/resources/generator/server/sonar/sonar-fullstack-project.properties.mustache
@@ -18,7 +18,6 @@ sonar.javascript.lcov.reportPaths={{projectBuildDirectory}}/test-results/lcov.in
 
 sonar.sourceEncoding=UTF-8
 sonar.exclusions=\
-  src/main/webapp/app/home/infrastructure/primary/*.*,\
   src/main/webapp/app/*.ts?(x),\
   src/main/webapp/content/**/*.*,\
   {{projectBuildDirectory}}/classes/static/**/*.*,\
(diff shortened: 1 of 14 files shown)


src/main/resources/generator/server/sonar/sonar-fullstack-project.properties.mustache
```diff
@@ -18,11 +18,9 @@ sonar.javascript.lcov.reportPaths={{projectBuildDirectory}}/test-results/lcov.in
 
 sonar.sourceEncoding=UTF-8
 sonar.exclusions=\
-  src/main/webapp/app/home/infrastructure/primary/*.*,\
-  src/main/webapp/app/*.ts,\
+  src/main/webapp/app/*.ts?(x),\
```
murdos, line 21: Have you found any documentation about the pattern format supported by Sonar?
The documentation is not explicit at all.
This StackOverFlow comment suggests that the pattern is Ant fileglob pattern: https://stackoverflow.com/questions/21323276/sonarqube-exclude-a-directory#comment58286503_31639014
If that's true, I guess we want 
```suggestion
  src/main/webapp/app/*.ts*,\
```

src/main/resources/generator/typescript/vitest.config.ts.mustache
```diff
@@ -30,6 +30,7 @@ export default defineConfig({
       include: ['src/main/webapp/**/*.ts?(x)'],
       exclude: [
         ...configDefaults.coverage.exclude as string[],
+        'src/main/webapp/app/*.ts?(x)',
```
murdos, line 33: Are you sure this is a valid glob pattern?
I would have used
```suggestion
        'src/main/webapp/app/*.ts{,x}',
```

qmonmert: I did same as on line 30

qmonmert:
@pascalgrimaud is this what you thought when you created the ticket?

pascalgrimaud:
> @pascalgrimaud is this what you thought when you created the ticket?

I didn't have a closed look, but the list of files in sonar should be really closed to the list in vitest config
If it's not the case, we can have 100% coverage with vitest but 90% coverage with Sonar for example

PR #11270, opened 2024-11-01
Sonar: reactivate circular dependencies rules

qmonmert:
try


murdos:
@qmonmert : what happened? Sonar has fixed the rules?

qmonmert:
@murdos I disable these rules because we had false positives https://community.sonarsource.com/t/advanced-class-cycle-detection-for-java/127500/6

PR #11296, opened 2024-11-03
Vue/ESLint: add @typescript-eslint/consistent-type-imports rule

qmonmert:


murdos:
@qmonmert : should we also add this rule for react generated apps?

qmonmert:
@murdos I will look

PR #11302, opened 2024-11-04
Cypress: Best practices for selectors

### LINKED ISSUE #11158: Add Cypress recommanded selectors to dataSelector
Cypress have [recommendations for data selectors](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements
)

We should update `src/main/resources/generator/client/common/cypress/utils/DataSelector.ts` to add them, probably something like:

```
export const dataSelector = (selector: string): string => `[data-selector="${selector}"],[data-testid="${selector}"],[data-test="${selector}"]`;
```

qmonmert:
Fix #11158 

https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements

image

### DIFF
--- a/src/main/resources/generator/client/common/cypress/utils/DataSelector.ts
+++ b/src/main/resources/generator/client/common/cypress/utils/DataSelector.ts
@@ -1,1 +1,1 @@
-export const dataSelector = (selector: string): string => `[data-selector="${selector}"]`;
+export const dataSelector = (selector: string): string => `[data-cy="${selector}"], [data-test="${selector}"], [data-testid="${selector}"]`;
--- a/src/main/resources/generator/client/common/cypress/utils/DataSelector.ts
+++ b/src/main/resources/generator/client/common/cypress/utils/DataSelector.ts
@@ -1,1 +1,2 @@
-export const dataSelector = (selector: string): string => `[data-cy="${selector}"], [data-test="${selector}"], [data-testid="${selector}"]`;
+export const dataSelector = (selector: string): string =>
+  `[data-selector="${selector}"], [data-cy="${selector}"], [data-test="${selector}"], [data-testid="${selector}"]`;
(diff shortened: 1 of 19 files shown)


src/main/resources/generator/client/common/cypress/utils/DataSelector.ts
```diff
@@ -1 +1 @@
-export const dataSelector = (selector: string): string => `[data-selector="${selector}"]`;
+export const dataSelector = (selector: string): string => `[data-cy="${selector}"], [data-test="${selector}"], [data-testid="${selector}"]`;
```
DamnClin, line 1: I think we should keep data-selector here

DamnClin (approved):

src/main/webapp/app/module/primary/landscape-minimap/LandscapeMiniMap.html
```diff
@@ -1,13 +1,11 @@
...
-    OPEN MINIMAP
-  </button>
+  <button v-if="!isMiniMapOpen" class="jhlite-button-main -block" @click="openMiniMap" data-test="show-minimap-btn">OPEN MINIMAP</button>
   <button
     class="jhlite-button-main -rounded-top -small"
     v-if="isMiniMapOpen"
     @click="closeMiniMap"
     title="Hide minimap"
-    data-selector="hide-minimap-btn"
+    data-test="hide-minimap-btn"
```
murdos, line 8: I would have rather used `data-testid` instead of `data-test` (everywhere)

murdos: because `data-test` is ambiguous, it seems like the content is some data for a test, that will read it to inject id elsewhere.
`data-testid` make it clearer that this is an identifier.

murdos (approved):

PR #11308, opened 2024-11-05
Cypress: fusion of 2 two folders

qmonmert:


murdos (changes requested):
Component tests and E2E tests are distinct type of tests, and there's no reason to merge the 2 folders that would mean E2E tests would be in the `component` folder.

If you check generated apps, you can use e.g. Cypress for component tests, and playwright for E2E tests: https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L186

qmonmert:
@murdos here the tests are for the generator

murdos:
> @murdos here the tests are for the generator

I know. 
But why would you structure the generator differently from the generated apps?

qmonmert:
For me the folder `e2e` and `component` are same (folders with cypress tests): one to test localhost:7471 end the other to test localhost:9000 ...

src/test/webapp/e2e/integration/Patch.spec.ts
murdos, line 1: Why are you removing this E2E?

You don't seem to understand that webapp component tests are testing the webapp with backend requests mocked, whereas E2E tests are testing the webapp with the real backend.
This is not just about using a different port. E2E test are verifying that webapp and backend can correctly communicate.

qmonmert: ok

qmonmert:
On vuejwtapp
image
Tests on e2e and component are same (one with cypress, the other with playwright)

murdos:
> On vuejwtapp image Tests on e2e and component are same (one with cypress, the other with playwright)

Yes, the initial content is identical, because we're generating very basics component and E2E tests.
Once you have generated your app, where everything is setup for you, you can start writing real component and/or E2E tests.

BTW, @qmonmert, what issue are you trying to solve with this PR?
You're just describing what is does (fusion of 2 two folders), but not why you're doing it.

PR #11335, opened 2024-11-10
Add Spring Boot LangChain4j sample

anunnakian:
Add a sample of using LangChain4J with Spring Boot Web


anunnakian:
@pascalgrimaud here is the PR to add a LangChain4j sample ;)

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/sample/jpapersistence/infrastructure/primary/SampleJpaPersistenceModuleConfiguration.java
```diff
@@ -1,34 +0,0 @@
...
-
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
-import tech.jhipster.lite.generator.server.springboot.mvc.sample.jpapersistence.application.SampleJpaPersistenceApplicationService;
-import tech.jhipster.lite.module.domain.resource.JHipsterModuleOrganization;
-import tech.jhipster.lite.module.domain.resource.JHipsterModulePropertiesDefinition;
-import tech.jhipster.lite.module.domain.resource.JHipsterModuleResource;
-
-@Configuration
-class SampleJpaPersistenceModuleConfiguration {
```
pascalgrimaud, line 16: Why deleting this file?

src/main/resources/generator/server/springboot/mvc/sample/langchain4j/main/infrastructure/primary/ChatResource.java.mustache
```diff
@@ -0,0 +1,21 @@
+package {{packageName}}.sample.infrastructure.primary;
+
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
pascalgrimaud, line 8: Need to use requestMapping to be under /api, so it will be visibled by springdoc

src/main/resources/generator/server/springboot/mvc/sample/langchain4j/main/infrastructure/primary/ChatResource.java.mustache
```diff
@@ -0,0 +1,21 @@
...
+public class ChatResource {
```
pascalgrimaud, line 9: No need to be public

src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/sample/jpapersistence/domain/SampleLangChain4jModuleFactoryTest.java
```diff
@@ -10,7 +10,7 @@
 import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
 
 @UnitTest
-class SampleJpaPersistenceModuleFactoryTest {
+class SampleLangChain4jModuleFactoryTest {
```
pascalgrimaud, line 13: Wrong dev here

pascalgrimaud (changes requested):

src/main/java/tech/jhipster/lite/generator/server/springboot/mvc/sample/jpapersistence/infrastructure/primary/SampleJpaPersistenceModuleConfiguration.java
anunnakian: You are to fast ! it was a mistake :D

src/test/java/tech/jhipster/lite/generator/server/springboot/mvc/sample/jpapersistence/domain/SampleLangChain4jModuleFactoryTest.java
anunnakian: Yep sorry :D

pascalgrimaud:
In the last PR, I forgot to tell you to add a new config to test in the CI, the new modules langchain4j

Could you add a new one here?
- https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/generate.sh#L232

Then, add it to the CI:
- https://github.com/jhipster/jhipster-lite/blob/main/.github/workflows/github-actions.yml#L232

pascalgrimaud:
By adding to the CI the generated app, it will show you that the generated app with langchain4j is not totally covered: missing tests for ChatResource

tests-ci/generate.sh
```diff
@@ -111,6 +111,8 @@ cucumber_with_jwt() {
     "spring-boot-cucumber-jwt-authentication"
 }
 
+echo $application
```
qmonmert, line 114:
```suggestion
```

pascalgrimaud:
Cc @renanfranca : what happened when we added a new config in script.sh and it doesn't exist in main branch? The checksum with md5 should be ignored, right?

renanfranca:
> Cc @renanfranca : what happened when we added a new config in script.sh and it doesn't exist in main branch? The checksum with md5 should be ignored, right?

The pipeline will break because the MD5 checksum is generated after it tries and fails to create the non-existent app in generate.sh in the main branch.

In my previous experience [creating a new app in generate.sh](https://github.com/jhipster/jhipster-lite/pull/11088#issuecomment-2404843543) and adding it to the github-actions.yml matrix, I had to add the new app in generate.sh and merge it with the main branch. Only after that could I add the new app to the matrix.

I saw your comment [here](https://github.com/jhipster/jhipster-lite/pull/11088#issuecomment-2406691288) only today. Yes, i think it should ignored non existing apps.

renanfranca:
> > I saw your comment [here](https://github.com/jhipster/jhipster-lite/pull/11088#issuecomment-2406691288) only today. Yes, i think it should ignored non existing apps.
> 
> But will it run on main branch CI? I think so...

Screenshot_20241110_231019_Vivaldi


I am having second thoughts about the importance of checking out the complete project from the main branch if it's only using the test-ci folder to run commands with the downloaded main jar file, which is not updated and generates an error when adding a new app in generate.sh and including the app in the GitHub Actions matrix in a pull request.

@pascalgrimaud and @murdos : What do you think about checking out from the current branch as well at step name: 'Setup: checkout project from main branch' and keeping everything else unchanged?

pascalgrimaud:
I created the ticket here https://github.com/jhipster/jhipster-lite/issues/11341

PR #11353, opened 2024-11-12
ci: use java 23 version instead of 22 for at least one build

murdos:
22 is now obsolete, let's use 23


renanfranca:
I agreed!  I am looking for Java 23 to use the newest features of java 23 and newst Spring boot 😃😃😃👏

PR #11443, opened 2024-11-27
fix tikui styles broken by inter.css

### LINKED ISSUE #11382: Tikui: inter.css brokes style
On http://localhost:9005/atom/atom.html#themebutton icons are not displayed

image

If I remove this https://github.com/jhipster/jhipster-lite/blob/246067ac1a8f527be46f5a0f37b529d236ee72f5/src/main/style/layout.pug#L11 it works

`inter.css` is not loaded in https://github.com/jhipster/jhipster-lite/blob/main/src/main/webapp/index.html so no problem on http://localhost:9000/landscape

I see this PR https://github.com/jhipster/jhipster-lite/pull/10702/files#diff-e87bb40b4aac9492f70669696ae9ad0f8b2989524fd5505a1c2d777566439580
@renanfranca can you look?

renanfranca:
- Fix #11382 

After some research, I found the reason why the `inter` font-family changed the sun (☀️) emoji color to grey only on Chrome and not on Firefox. 

Firefox:
image

Chrome:
image

According to the documentation, the Inter font has a specific behavior with certain emojis due to a specific glyph (U+20E3) that interferes with the rendering of special characters in Chrome[1]. This issue occurs because Chrome has difficulty resolving characters when the Inter font is present, causing some specific emojis not to render correctly[1].

The problem doesn't affect all emojis, only specific characters, as it depends on how Chrome discovers and resolves these special characters when it finds the Inter font in its font stack[1]. When Inter is present, Chrome cannot properly fall back to the system's emoji font in some specific cases[1].

Unfortunately, I can't explain why the moon emoji (🌙) didn't turn grey and only the sun emoji (☀️). However, the technical behavior described explains why some emojis are affected while others continue to work normally.

Citations:
[1] https://github.com/tailwindlabs/tailwindcss/issues/2499
[2] https://rsms.me/inter/
[3] https://github.com/rsms/inter/issues/371
[4] https://forum.obsidian.md/t/the-sun-emoji-is-displayed-as-a-monochrome-glyph/47751 

Use the sun and moon [here](https://rsms.me/inter/lab/?antialias=default&sample=%E2%98%80%EF%B8%8F%F0%9F%8C%99):

image

### DIFF
--- a/src/main/style/atom/theme-button/_theme-button.scss
+++ b/src/main/style/atom/theme-button/_theme-button.scss
@@ -1,3 +1,5 @@
+@use '../../token/fonts';
+
 .jhlite-theme-button {
   display: flex;
 
@@ -21,6 +23,7 @@
     padding: calc(var(--jhlite-element-size) * 0.1);
     width: var(--jhlite-element-size);
     height: calc(var(--jhlite-element-size) * 0.35);
+    font-family: fonts.$jhlite-global-font-emoji-family;
     font-size: calc(var(--jhlite-element-size) * 0.3);
   }
 
--- a/src/main/style/token/_fonts.scss
+++ b/src/main/style/token/_fonts.scss
@@ -2,6 +2,7 @@
 
 $jhlite-font-weight-semi-bold: 600;
 $jhlite-global-font-primary-family: 'Inter', sans-serif;
+$jhlite-global-font-emoji-family: sans-serif;
 $jhlite-global-font-text-size: 2.5vw;
 $jhlite-global-font-text-desktop-size: 16px;
 $jhlite-global-font-glyph-family: jhlite-icons;


qmonmert (approved):
ok, but strangely I don't reproduce on the main branch (on Chrome, maybe new version of Chrome fixes that)

renanfranca:
> ok, but strangely I don't reproduce on the main branch (on Chrome, maybe new version of Chrome fixes that)

I used an old version of Chrome in my wsl2 Ubuntu.

renanfranca:
> ok, but strangely I don't reproduce on the main branch (on Chrome, maybe new version of Chrome fixes that)

@qmonmert : I tested in [Vivaldi browser](https://vivaldi.com) which is based in chromium like google chrome and the sun is grey:

image

And the moon is yellow:
image

qmonmert:
👍

PR #11489, opened 2024-12-02
React/ESLint: enable react/react-in-jsx-scope

qmonmert:


murdos:
@qmonmert : are you sure that this rule should be enabled?

See https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md
> If you are using the [new JSX transform from React 17](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports), you should disable this rule by extending [react/jsx-runtime](https://github.com/jsx-eslint/eslint-plugin-react/blob/8cf47a8ac2242ee00ea36eac4b6ae51956ba4411/index.js#L165-L179) in your eslint config (add "plugin:react/jsx-runtime" to "extends").

qmonmert:
@murdos I see that it is what we do on the JH generator, but maybe it's not a good idea

PR #11525, opened 2024-12-06
ci: use node 22.11.0 rather that latest lts

### LINKED ISSUE #11522: Build failed for react app
Current reactapp failed to build.
We need to fix it, so we can merge new Spring Boot 3.4 then release :)

Do you want to work on this @qmonmert ?

murdos:
In order to avoid version 22.12.0 that contains some regression. Fixes #11522


pascalgrimaud:
@murdos : approved

Issue #11513, opened 2024-12-05
Gradle CI doesn't launch unit tests


fabienpuissant:
It seems that the gradle ci don't launch unit test only IT.

I have tried to put a failing unit test and the ci was green 

Is this the behaviour or a misconfiguration on my side ?

A possible working solution

**github-actions.yml**

```
jobs:
  tests:
    name: tests
    runs-on: ubuntu-latest
    timeout-minutes: 20
    steps:
      - name: 'Setup: checkout project'
        uses: actions/checkout@v4
      - name: 'Setup: environment'
        id: setup
        uses: ./.github/actions/setup
      - name: Setup Gradle
        uses: gradle/gradle-build-action@v2
      - name: 'Test: run unit tests'
        run: ./gradlew clean test --no-daemon
      - name: 'Test: run integration tests'
        run: ./gradlew clean integrationTest --no-daemon
```

renanfranca:
@fabienpuissant : I looked at a random `build` action: [feat(liquibase): async liquibase configuration should not be activated if liquibase is disabled](https://github.com/jhipster/jhipster-lite/actions/runs/12189645586/job/34005487351). At least for this example, the tests were executed:

```bash
> Task :test

JwtReaderTest > should not authenticate user from empty token STANDARD_OUT
    2024-12-05T23:23:09.354Z  INFO   --- [    Test worker] t.j.f.s.a.i.primary.JwtReader            : Invalid empty JWT token

JwtReaderTest > should not authenticate user from blank token STANDARD_OUT
    2024-12-05T23:23:09.359Z  INFO   --- [    Test worker] t.j.f.s.a.i.primary.JwtReader            : Invalid empty JWT token

JwtReaderTest > should not authenticate user from invalid token STANDARD_OUT
    2024-12-05T23:23:09.481Z  INFO   --- [    Test worker] t.j.f.s.a.i.primary.JwtReader            : Invalid JWT token
    2024-12-05T23:23:09.482Z TRACE   --- [    Test worker] t.j.f.s.a.i.primary.JwtReader            : Invalid JWT token: Invalid compact JWT string: Compact JWSs must contain exactly 2 period characters, and compact JWEs must contain exactly 4.  Found: 0
```

Here is the part of the `github-actions.yml` responsible for that:

```yml
      - name: 'Test: verify ${{ matrix.app }}'
        if: steps.tests-requirement-check.outputs.execute_tests == 'true'
[...]

PR #11556, opened 2024-12-09
add unit test step into gradle gitlab ci script

renanfranca:
- Following https://github.com/jhipster/jhipster-lite/issues/11513


atomfrede (approved):
Looks fine. I would also remove the clean for the test tasks (it is the only task on a most likely fresh instance of a runner)

fabienpuissant (approved):

renanfranca:
> Looks fine. I would also remove the clean for the test tasks (it is the only task on a most likely fresh instance of a runner)

Thanks! Done ✌

PR #11602, opened 2024-12-18
chore(generator): configure Error Prone and fix existing issues

murdos:
See http://errorprone.info/


murdos:
@pascalgrimaud : what do you think about this? 
If wanted, I can remove the execution of Error Prone, and keep the second commit that improves the current code.

pascalgrimaud:
I'm fine for using this, if it can improve our code.
We just need to be sure it won't conflict with Sonar analysis, as it can happened when we use several tools like this

PR #11606, opened 2024-12-19
feat(checkstyle): use latest checkstyle version when using checkstyle maven plugin

murdos:


murdos:
Failing build was caused by main being broken due to esbuild and vite v6.0.4: https://github.com/evanw/esbuild/issues/4010
I've manually triggered renovate update for vite v6.0.5 to fix the issue.

PR #11645, opened 2024-12-27
Fix some warnings related to errorprone

pascalgrimaud:
- related to https://github.com/jhipster/jhipster-lite/pull/11602

PR #11647, opened 2024-12-27
chore(tests): remove unused logs spy

murdos:
related to  #11645


DamnClin (approved):

DamnClin:
We switch static import mode?

murdos:
> We switch static import mode?

There's no official guideline, and no tool configured to enforce a specific style, so it depends of each contributor :sweat_smile:

PR #11660, opened 2024-12-29
Vue/ESLint v-html directive can lead to XSS attack

qmonmert:
Fix
image


fabienpuissant:
Yes we can remove the v-html tag and just interpolate with t()

https://github.com/i18next/i18next-vue

I don't know the reason I've put this tag 😄

qmonmert:
> Yes we can remove the v-html tag and just interpolate with t()
> 
> https://github.com/i18next/i18next-vue
> 
> I don't know the reason I've put this tag 😄

👍

PR #11661, opened 2024-12-29
Fix empty char on commands

qmonmert:
Fix
image

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/startupcommand/MavenStartupCommandLine.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/startupcommand/MavenStartupCommandLine.java
@@ -9,6 +9,7 @@ public record MavenStartupCommandLine(String commandLineParameters) implements J
 
   @Override
   public StartupCommandLine commandLine() {
-    return new StartupCommandLine("./mvnw " + commandLineParameters.trim());
+    String commandLineParametersTrim = commandLineParameters.trim();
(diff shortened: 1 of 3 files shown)


murdos:
What exactly is the issue that you're trying to fix?
There's only a screenshot, and I don't understand it.

qmonmert:
> What exactly is the issue that you're trying to fix? There's only a screenshot, and I don't understand it.

image

src/main/java/tech/jhipster/lite/module/domain/startupcommand/GradleStartupCommandLine.java
```diff
@@ -9,6 +9,7 @@ public record GradleStartupCommandLine(String commandLineParameters) implements
 
   @Override
   public StartupCommandLine commandLine() {
-    return new StartupCommandLine("./gradlew " + commandLineParameters.trim());
+    String commandLineParametersTrim = commandLineParameters.trim();
+    return new StartupCommandLine("./gradlew" + (commandLineParametersTrim.isEmpty() ? "" : " " + commandLineParametersTrim));
```
murdos, line 13: A better solution:
image

qmonmert: have you try? because `"./gradlew "` this

murdos: Yes, I've tried from your branch, and my suggestion works.

qmonmert: good catch

PR #11678, opened 2025-01-03
feat(liquibase): liquibase module doesn't need logs-spy module. 

murdos:
only liquibase-async need it.


renanfranca (approved):
Thank you for your work 😃

PR #11679, opened 2025-01-03
feat(logs-spy): logs-spy module should only require spring-boot, not spring-server

murdos:


renanfranca (approved):
LGTM :) . Happy new year!

PR #11684, opened 2025-01-04
Fix: update springdoc-openapi.version to v2.8.0

qmonmert:
Fix #11681 

`type` property does not appear in the JSON `http://localhost:7471/v3/api-docs/all`

I still can't find out why, maybe with this update `Upgrade swagger-core to 2.2.27`

https://github.com/springdoc/springdoc-openapi/releases/tag/v2.8.0


murdos:
I started investigating this too, and I think this is rather related to the switch to OpenAPI 3.1.
I'll investigate a bit more.

pascalgrimaud:
can it be merged @murdos @qmonmert ?

murdos:
Yeah, let's merge it that way :)

PR #11724, opened 2025-01-14
add optional rank at modules domain

### LINKED ISSUE #10934: New state for modules
Today, we have 2 states for modules:
- available, displayed in the UI
- disabled, which are hidden, specially because it's not maintained

I want to add a new state:
- beta
- wip
- or another name

These modules will be displayed with a warning or another color (any suggestions are welcomed)

So we can start new modules and get some feedbacks, and totally rework if needed, without worrying about the use

What do you think?
Good or bad idea?

renanfranca:
- related to #10934

### DIFF
--- a/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleRank.java
+++ b/src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleRank.java
@@ -1,36 +1,35 @@
 package tech.jhipster.lite.module.domain.resource;
 
+/**
+ * Represents the maturity and adoption level of JHipster modules.
+ * Ranks range from experimental to community-validated,
+ * helping users assess module stability and production readiness.
+ */
 public enum JHipsterModuleRank {
   /**
-   * Not really usable as is, unless you have good expertise on the subject
-   * (ex: custom-jhlite)
+   * Experimental or advanced module requiring specific expertise
    */
   RANK_D,
 
   /**
-   * No known usage on production product
-   * (ex: vue-core, the current version)
+   * Module without known production usage
    */
   RANK_C,
 
   /**
-   * One declared usage on production product
-   * (ex: kipe-authorization)
+   * Module with at least one confirmed production usage
    */
   RANK_B,
 
   /**
-   * Multiple declared usages on production product, by multiple person, on various projects
-   * and demonstrated usage on a talk, book or blog post
-   * (ex: java-base)
+   * Module with multiple production usages across different projects
+   * and documented through talks, books or blog posts
    */
   RANK_A,
 
   /**
-   * A and recognized to add some features that are not available otherwise.
-   * Features must be really unique and this must be recognized by at least 10 person
-   * with a vote on any social network (GitHub being a social network).
-   * With those criteria spring-boot-cucumber-mvc is a good candidate (missing votes).
+   * Production-proven module providing unique features,
+   * validated by community feedback (10+ endorsements)
    */
   RANK_S,
 }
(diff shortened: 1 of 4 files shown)


src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleRank.java
```diff
@@ -0,0 +1,9 @@
+package tech.jhipster.lite.module.domain.resource;
+
+import tech.jhipster.lite.shared.error.domain.Assert;
+
+public record JHipsterModuleRank(String rank) {
```
murdos, line 5: How about using an enum instead?

pascalgrimaud, line 5: as we already know the possible value, I would use an enum here: RANK_S, RANK_A, etc

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleResource.java
```diff
@@ -14,6 +15,7 @@ public final class JHipsterModuleResource {
   private final JHipsterModuleApiDoc apiDoc;
   private final JHipsterModuleOrganization organization;
   private final JHipsterModuleTags tags;
+  private final Optional<JHipsterModuleRank> rank;
```
pascalgrimaud, line 18: no need to be optional here, as there is a default value to RANK_D

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleResource.java
```diff
@@ -24,6 +26,7 @@ private JHipsterModuleResource(JHipsterModuleResourceBuilder builder) {
     apiDoc = builder.apiDoc;
     tags = builder.tags;
     organization = builder.organization;
+    rank = Optional.ofNullable(builder.rank).or(() -> Optional.of(new JHipsterModuleRank("D")));
```
pascalgrimaud, line 29: would be something like Optional.ofNullable(builder.rank).orElse(JHipsterModuleRank.RANK_D);

pascalgrimaud (changes requested):

renanfranca:
Thank you for the reviews.  I will work on this tomorrow ✌️😀

pascalgrimaud:
@renanfranca : can you squash all your commits plz?

renanfranca:
> @renanfranca : can you squash all your commits plz?

@pascalgrimaud : sure, it is done 👍

src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleRank.java
```diff
@@ -0,0 +1,36 @@
+package tech.jhipster.lite.module.domain.resource;
+
+public enum JHipsterModuleRank {
+  /**
```
pascalgrimaud, line 4: not sure we should keep all these comments, as we'll need to update them, as soon as there are some change.

For example: as soon as kipe-authorization change his RANK, the comment here won't be correct

murdos: > it is hard to understand rank usage

I'm also a bit sceptical about these rank usages that are very subjective, so I'm curious to see how this will turn out :)

renanfranca: In my understanding, the ranks are curated by the jhipster lite developers to guide those who are using the tool for the first time. Let's give a try ;)

pascalgrimaud: I will not impact the final user, as we'll always display by default "ALL RANK".

Then, for example, the user can select the RANK_S to display only the most used modules. I don't think he/she'll be unhappy because his prefered module is not displayed in RANK_S, because his module will always be in "ALL RANK", so I'm fine with this

renanfranca: @pascalgrimaud : please, let me know if I can keep these improved comments. Thanks :)

pascalgrimaud (approved):

PR #11755, opened 2025-01-19
Tests: clean code

qmonmert:


pascalgrimaud:
@qmonmert : don't forget to add an area, otherwise, it won't be in a section in the release note :)

image

qmonmert:
@pascalgrimaud I did not understand that we need an area, I thought that a label was good enough

pascalgrimaud:
See https://github.com/jhipster/jhipster-lite/blob/main/.github/release-drafter.yml

Here the label which are used in release notes

qmonmert:
ok thanks

pascalgrimaud:
But if it's forgotten, no worry, I'll take care of it before a release

qmonmert:
Thank you for your work

PR #11838, opened 2025-02-01
Fix config: Prettier plugin for ESLint

qmonmert:


murdos (changes requested):
Could you explain why you're doing this change?

This is not the recommended way of integrating prettier with eslint: see https://prettier.io/docs/integrating-with-linters/

We're already using eslint-config-prettier, the recommended integration.

murdos:
> `eslint-config-prettier` is just used to configure `eslint-plugin-prettier` 

It's not. Have you read https://prettier.io/docs/integrating-with-linters/ ?
Either you use `eslint-config-prettier`, or `eslint-plugin-prettier`, but not both.

- `eslint-config-prettier` is used to disable rules in eslint that conflict or are unnecessary with Prettier.
https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#installation never mention eslint-plugin-prettier

- `eslint-plugin-prettier` is used to run prettier as part of the linting process of eslint.
But for us it means that prettier will run twice, since we're executing eslint and prettier independently as recommended by prettier: https://github.com/jhipster/jhipster-lite/blob/0e2eb15d98b0541acb64a338d610c57437c94706/.lintstagedrc.js#L2-L4

> Actually our config is useless

It's not. It's used here: https://github.com/jhipster/jhipster-lite/blob/0e2eb15d98b0541acb64a338d610c57437c94706/eslint.config.js#L60
And if you try to remove this line, and you'll see that result of eslint is different: it'll list 45 problems.

pascalgrimaud:
To be honest, generator-jhipster is not a reference anymore, since maybe 2 years. We should not rely on this, as some implementations/choices are done without real review, real feedback.
It's a very bad idea so copy/paste some code from generator-jhipster, in my opinion.

The best to do is to go to official documentation of the library and check what they suggest, like @murdos did

qmonmert:
@pascalgrimaud I did not copy/paste without testing, I don't work like that... Maybe I did a mistake.
Sorry guys for the inconvenience.
I close this PR.

PR #11866, opened 2025-02-06
use typed translation on react i18next on each context

### LINKED ISSUE #11111: Vue internationalization choices?
Playing with `vue-i18next` module I have some questions on the choices that were made:

* Why do we need explicit declarations? If I understand it right, declaring translations is:
  * Create translation files;
  * Create a XXXTranslations.ts file to export those translations;
  * Declare it in i18n.ts.
* Why do we use ts file instead of json files for translations?

I was expecting something to just drop json files somewhere in context to get translations for the context and I'm confused with the tradeoffs :D. cc @fabienpuissant

fabienpuissant:
Fix #11111,#11002

### DIFF
--- a/src/main/resources/generator/client/common/i18n/i18n.spec.ts
+++ b/src/main/resources/generator/client/common/i18n/i18n.spec.ts
@@ -1,13 +1,13 @@
 import i18n from '@/i18n';
-import { mergeTranslations } from '@/Translations';
+import { mergeTranslations, type Translations } from '@/Translations';
 
 describe('i18n configuration', () => {
   it('loads en translation', () => {
-    expect(i18n.getResourceBundle('en', '')['home']['translationEnabled']).toBe('Internationalization enabled');
+    expect((i18n.getResourceBundle('en', '') as Translations).home.translationEnabled).toBe('Internationalization enabled');
   });
 
   it('loads fr translation', () => {
-    expect(i18n.getResourceBundle('fr', '')['home']['translationEnabled']).toBe('Internationalisation activée');
+    expect((i18n.getResourceBundle('fr', '') as Translations).home.translationEnabled).toBe('Internationalisation activée');
   });
 
   describe('mergeTranslations function', () => {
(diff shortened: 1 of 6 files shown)


src/main/resources/generator/client/common/i18n/i18n.spec.ts
```diff
@@ -3,10 +3,12 @@ import { mergeTranslations } from '@/Translations';
 
 describe('i18n configuration', () => {
   it('loads en translation', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     expect(i18n.getResourceBundle('en', '')['home']['translationEnabled']).toBe('Internationalization enabled');
```
qmonmert, line 7:
```suggestion
    expect((i18n.getResourceBundle('en', '') as Translations).home.translationEnabled).toBe('Internationalization enabled');
```
with
`import { mergeTranslations, type Translations } from '@/Translations';`

src/main/resources/generator/client/common/i18n/i18n.spec.ts
```diff
@@ -3,10 +3,12 @@ import { mergeTranslations } from '@/Translations';
...
   });
 
   it('loads fr translation', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     expect(i18n.getResourceBundle('fr', '')['home']['translationEnabled']).toBe('Internationalisation activée');
```
qmonmert, line 12:
```suggestion
    expect((i18n.getResourceBundle('fr', '') as Translations).home.translationEnabled).toBe('Internationalisation activée');
```
with `import { mergeTranslations, type Translations } from '@/Translations';`

src/main/resources/generator/client/common/i18n/i18n.spec.ts
```diff
@@ -1,13 +1,14 @@
 import i18n from '@/i18n';
+import type { Translations } from '@/Translations';
 import { mergeTranslations } from '@/Translations';
```
qmonmert, line 3:
```suggestion
import { mergeTranslations, type Translations } from '@/Translations';
```

fabienpuissant:
Thank's @qmonmert 😄

qmonmert:
> Thank's @qmonmert 😄

🤝

murdos:
Thanks @fabienpuissant !

PR #11898, opened 2025-02-12
Vite: set build.sourcemap to true

qmonmert:
Hide this message
image

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -22,6 +22,7 @@ export default defineConfig({
   ],
   build: {
     outDir: '../../../target/classes/static',
+    sourcemap: true,
   },
   root: 'src/main/webapp',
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -22,6 +22,7 @@ export default defineConfig({
   ],
   build: {
     outDir: '../../../target/classes/static',
+    sourcemap: true,
```
murdos, line 25: The "warning" (it's not really one BTW) only appears when you run vite locally for testing.
However activating sourcemap like you do will also impact the build for the production build.
And I don't know if it's desirable;

qmonmert: yes you're right 
image
https://github.com/ifaxity/vite-plugin-istanbul

PR #11981, opened 2025-02-24
Angular: Configure environment-specific defaults

qmonmert:
Related to https://angular.dev/tools/cli/environments#configure-environment-specific-defaults

image


src/main/java/tech/jhipster/lite/generator/client/angular/core/domain/AngularModuleFactory.java
```diff
@@ -114,8 +114,8 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
           .addFile("AngularLogo.svg")
           .and()
         .batch(SOURCE.file("src/main/webapp/environments"), to("src/main/webapp/environments"))
-          .addTemplate("environment.prod.ts")
-          .addTemplate("environment.prod.spec.ts")
+          .addTemplate("environment.development.ts")
```
murdos, line 117: I was wondering if we should not use the term `local` rather than `development`, since this is the word we use for the Spring profile for starting the application locally.

pascalgrimaud: Agree too, I don't like dev/prod term, as it's too confusing

qmonmert: it's the terms used by Angular but we can change

murdos (approved):

PR #12075, opened 2025-03-10
Add rank when building modules

pascalgrimaud:


src/main/java/tech/jhipster/lite/module/domain/resource/JHipsterModuleResource.java
```diff
@@ -26,7 +26,7 @@ private JHipsterModuleResource(JHipsterModuleResourceBuilder builder) {
     apiDoc = builder.apiDoc;
     tags = builder.tags;
     organization = builder.organization;
-    rank = Optional.ofNullable(builder.rank).orElse(JHipsterModuleRank.RANK_D);
+    rank = JHLiteModuleSlug.getRank(slug.get()).orElse(JHipsterModuleRank.RANK_D);
```
murdos, line 29: The issue is that this will not work for slugs of custom modules from a custom jhlite.
A proper solution is to extends the module slug factory interface.

pascalgrimaud: Indeed, all custom modules will be rank D :-D

PR #12094, opened 2025-03-13
Document dependencies package for Angular

qmonmert:
Same as:
- https://github.com/jhipster/jhipster-lite/blob/0105d3df64c795b54cf4322543e0720566c2016a/src/main/resources/generator/dependencies/react/package.json#L2
- https://github.com/jhipster/jhipster-lite/blob/0105d3df64c795b54cf4322543e0720566c2016a/src/main/resources/generator/dependencies/vue/package.json#L2

### DIFF
--- a/src/main/resources/generator/dependencies/angular/package.json
+++ b/src/main/resources/generator/dependencies/angular/package.json
@@ -1,4 +1,8 @@
 {
+  "name": "jhlite-dependencies",
+  "version": "0.0.0",
+  "description": "JHipster Lite : used for Angular+TypeScript dependencies",
+  "license": "Apache-2.0",
   "dependencies": {
     "@angular-builders/jest": "19.0.0",
     "@angular/build": "19.2.1",


src/main/resources/generator/dependencies/angular/package.json
```diff
@@ -1,4 +1,8 @@
 {
+  "name": "jhlite-dependencies",
+  "version": "0.0.0",
+  "description": "JHipster Lite : used for Angular+TypeScript dependencies",
```
murdos, line 4:
```suggestion
  "description": "JHipster Lite : used for Angular dependencies",
```

qmonmert: we have a specific version of TypeScript in this package.json

murdos: Sure, but then why not also talking about jest, zonejs, rxjs?
The name of the NpmVersionSource is ANGULAR, not ANGULAR_AND_TYPESCRIPT: https://github.com/jhipster/jhipster-lite/blob/a86bfc71dec6348d4e80cc940b303c09077eb9fe/src/main/java/tech/jhipster/lite/module/domain/npm/JHLiteNpmVersionSource.java#L8

PR #12106, opened 2025-03-15
Rename SpringDoc to Springdoc

qmonmert:
Standardize


murdos (changes requested):
https://springdoc.org/ never use the case `SpringDoc`.
I think we should stick to `Springdoc`

qmonmert:
@murdos so we replace all SpringDoc with Springdoc?

murdos (approved):

PR #12119, opened 2025-03-16
chore: add Intellij icon

murdos:
An icon is then automatically displayed in IntelliJ :)
image


renanfranca (approved):
@murdos : That's so cool! I loved it 😍!

PR #12125, opened 2025-03-16
Rename factory variable

qmonmert:


murdos:
Sorry, but I don't agree with the renaming. Just using `factory` is a bit too vague, it's better to have something more explicit for e.g. with `approvalTesting`.

qmonmert:
@murdos so we rename all https://github.com/jhipster/jhipster-lite/blob/e909489bc37d1d65a4f3e343ab07e089583659b4/src/main/java/tech/jhipster/lite/generator/buildtool/maven/application/MavenApplicationService.java#L11 
?

PR #12132, opened 2025-03-17
bug: missing eslint-config-prettier configuration in eslint.config.js

### LINKED ISSUE #12124: bug: missing eslint-config-prettier configuration in eslint.config.js
For both react and vue, eslint-config-prettier is not configured, which leads to conflict between prettier and eslint.

qmonmert:
Fix #12124

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/typescript/core/domain/TypescriptModuleFactory.java
+++ b/src/main/java/tech/jhipster/lite/generator/typescript/core/domain/TypescriptModuleFactory.java
@@ -1,14 +1,10 @@
 package tech.jhipster.lite.generator.typescript.core.domain;
 
-import static tech.jhipster.lite.module.domain.JHipsterModule.from;
-import static tech.jhipster.lite.module.domain.JHipsterModule.moduleBuilder;
-import static tech.jhipster.lite.module.domain.JHipsterModule.packageName;
-import static tech.jhipster.lite.module.domain.JHipsterModule.scriptCommand;
-import static tech.jhipster.lite.module.domain.JHipsterModule.scriptKey;
-import static tech.jhipster.lite.module.domain.JHipsterModule.to;
+import static tech.jhipster.lite.module.domain.JHipsterModule.*;
 import static tech.jhipster.lite.module.domain.npm.JHLiteNpmVersionSource.COMMON;
 import static tech.jhipster.lite.module.domain.packagejson.NodeModuleFormat.MODULE;
 
+import java.util.function.Consumer;
 import tech.jhipster.lite.module.domain.JHipsterModule;
 import tech.jhipster.lite.module.domain.file.JHipsterSource;
 import tech.jhipster.lite.module.domain.npm.NpmLazyInstaller;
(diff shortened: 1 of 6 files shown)


murdos (changes requested):
Actually the typescript module doesn't depend on prettier, so it's not really fair to add eslint-config-prettier here.
Furthermore eslint-config-prettier recommends to add prettier config in last position.

qmonmert:
@murdos we add ESLint in TypescriptModuleFactory, so I don't think that we must add eslint-config-prettier in PrettierModuleFactory ... WDYT?

murdos:
> @murdos we add ESLint in TypescriptModuleFactory, so I don't think that we must add eslint-config-prettier in PrettierModuleFactory ... WDYT?

My first thought was to add `eslint-config-prettier` dependency (and config) in react and vue modules, since they both require typescript and prettier.

But indeed that would not work if you just apply `prettier` and `typescript` module: the conflict between both tools will exist.

I'm tempted to keep the addition of eslint-config-prettier in PrettierModuleFactory like you did, but to also make `typescript` depends on `prettier`. That would resolve the case. WDYT?
In that case we would be able to remove the dependency to prettier from vue and react.

src/main/java/tech/jhipster/lite/generator/typescript/core/domain/TypescriptModuleFactory.java
```diff
@@ -62,7 +58,18 @@ public JHipsterModule buildModule(JHipsterModuleProperties properties) {
           .addTemplate("eslint.config.js")
           .and()
         .and()
+      .apply(patchEslintConfigPrettier())
```
murdos, line 61: You can just update the template `eslint.config.js` that is added a few lines above (l. 58), this would be much simpler :)

qmonmert:
> > @murdos we add ESLint in TypescriptModuleFactory, so I don't think that we must add eslint-config-prettier in PrettierModuleFactory ... WDYT?
> 
> My first thought was to add `eslint-config-prettier` dependency (and config) in react and vue modules, since they both require typescript and prettier.
> 
> But indeed that would not work if you just apply `prettier` and `typescript` module: the conflict between both tools will exist.
> 
> I'm tempted to keep the addition of eslint-config-prettier in PrettierModuleFactory like you did, but to also make `typescript` depends on `prettier`. That would resolve the case. WDYT? In that case we would be able to remove the dependency to prettier from vue and react.

I agree to add a dependency between Typescript and Prettier.
To resume: if I add Prettier (with eslint-config-prettier), I need Typescript (with eslint) before.
Ok for you?

src/main/java/tech/jhipster/lite/generator/prettier/infrastructure/primary/PrettierModuleConfiguration.java
```diff
@@ -19,7 +18,7 @@ JHipsterModuleResource prettierModule(PrettierApplicationService prettier) {
       .slug(PRETTIER)
       .propertiesDefinition(initPropertiesDefinition())
       .apiDoc("Prettier", "Format project with prettier")
-      .organization(JHipsterModuleOrganization.builder().addDependency(INIT).build())
+      .organization(JHipsterModuleOrganization.builder().addDependency(INIT).addDependency(TYPESCRIPT).build())
```
murdos, line 21: Sorry, I was maybe not clear: it's the other way: Typescript module should depends on Prettier.

qmonmert: 😕 and eslint-config-prettier is added in TypescriptModuleFactory or PrettierModuleFactory ?

murdos: Eslint-config-prettier is added by the typescript module (because it adds eslint).

murdos (approved):

murdos:
Thanks @qmonmert !

pascalgrimaud:
Adding a bounty for all your last PRs, @qmonmert 
thanks for your work

qmonmert:
@pascalgrimaud thanks https://opencollective.com/generator-jhipster/expenses/243647

pascalgrimaud:
@qmonmert : approved

PR #12194, opened 2025-03-26
Optimize imports

qmonmert:


src/test/resources/tech/jhipster/lite/generator/client/vue/core/domain/VueModuleFactoryTest.shouldBuildVueModule.tsconfig.json.approved.txt
```diff
@@ -0,0 +1,13 @@
+{
```
murdos, line 1: I don't understand where these approved files come from, and how they relate to this PR.
Do you know?

qmonmert: my bad

pascalgrimaud:
Not sure about this, as we don't have the same config about import static

murdos (approved):

renanfranca:
> Not sure about this, as we don't have the same config about import static

I think we should change our intellij idea configuration to not use wild card...

renanfranca:
> Not sure about this, as we don't have the same config about import static

I think we should change our intellij idea configuration to not use wild card...

pascalgrimaud:
we should be consistent between us :)

murdos:
> we should be consistent between us :)

We can enforce a no-star policy with checkstyle. That way we can be sure the code is consistent and we don't forget to change our IntelliJ config.

renanfranca:
> > we should be consistent between us :)
> 
> We can enforce a no-star policy with checkstyle. That way we can be sure the code is consistent and we don't forget to change our IntelliJ config.

Nice 😎, thanks!

qmonmert:
I close the PR

murdos:
> > we should be consistent between us :)
> 
> We can enforce a no-star policy with checkstyle. That way we can be sure the code is consistent and we don't forget to change our IntelliJ config.

@pascalgrimaud : Do you agree to adding such a checkstyle verification?

pascalgrimaud:
I think we could have this checkstyle verification only if we have a formater which can fix the import correctly after each commit
Otherwise, I afraid each new contributor must configure their IDE before contributing

So let's discuss before adding this new rule

PR #12196, opened 2025-03-27
Rework test file

qmonmert:


src/test/webapp/unit/module/primary/project-actions/ProjectActionsComponent.spec.ts
```diff
@@ -50,18 +50,18 @@ describe('Project actions', () => {
     it('should disable formatting without project path', async () => {
       const wrapper = wrap({ folderPath: '' });
 
-      await flushForm(wrapper);
```
murdos, line 53: Could you explain to us why it's not necessary anymore?

PR #12236, opened 2025-04-03
fix the rank button alignment to be centered

renanfranca:
Here is a demo video showing the alignment was fixed:

https://github.com/user-attachments/assets/e4a9a3ff-1322-4853-9ca3-c11cba838b3b

I needed to hide the short name (e.g., 'S') only after the full name (e.g., 'RANK S') was completely shown. That's why I had to create a component `ToggleButtonExpandableVue` to encapsulate the logic to hide the short name using v-if only after the opacity transition from the full name finished.

I tried to fix the problem using only CSS, and I encountered the following strange behavior (a glitch). I can call it a "visual bug"; it is unpleasant and annoying :). Here is a demo video showing the glitch:

https://github.com/user-attachments/assets/f6277c5d-26b5-40eb-992b-fc934854b8d8

### DIFF
--- a/src/test/webapp/unit/module/primary/landscape/LandscapeRankModuleFilterComponent.spec.ts
+++ b/src/test/webapp/unit/module/primary/landscape/LandscapeRankModuleFilterComponent.spec.ts
@@ -12,7 +12,7 @@ import { applicationBaseNamePropertyDefinition } from '../../domain/Modules.fixt
 const wrap = (props?: { moduleRankStatistics: ModuleRankStatistics }): VueWrapper => {
   return mount(LandscapeRankModuleFilterVue, {
     props: props || {
-      moduleRankStatistics: RANKS.map(rank => ({ rank, quantity: 1 })), // Force enable all ranks
+      moduleRankStatistics: RANKS.map(rank => ({ rank, quantity: 1 })),
     },
   });
 };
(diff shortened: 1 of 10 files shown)


src/test/webapp/unit/module/primary/landscape/LandscapeRankModuleFilterComponent.spec.ts
```diff
@@ -148,6 +148,46 @@ describe('LandscapeRankModuleFilterComponent', () => {
     }
   });
 
+  it('should display buttons without colors by default', async () => {
```
qmonmert, line 151:
```suggestion
  it('should display buttons without colors by default', () => {
```
I think

src/test/webapp/unit/module/primary/landscape/LandscapeRankModuleFilterComponent.spec.ts
```diff
@@ -148,6 +148,46 @@ describe('LandscapeRankModuleFilterComponent', () => {
...
+    };
+    for (const rank of RANKS) {
+      const rankButton = wrapper.find(wrappedElement(`rank-${rank}-filter`));
+
+      expect(rankButton.classes()).not.toContain('-rank-color');
+      expect(rankButton.classes()).not.toContain(rankColorClasses[rank]);
+    }
+  });
+
+  it('should display buttons without colors when a selected rank is deselected', async () => {
```
qmonmert, line 169:
```suggestion
  it('should display buttons without colors when a selected rank is deselected', () => {
```
same

src/test/webapp/unit/shared/toggle-button-expandable/infrastructure/primary/ToggleButtonExpandable.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+        },
+      });
+
+      const shortName = wrapper.find('[data-testid="short-name-S"]');
+
+      expect(shortName.exists()).toBe(true);
+      expect(shortName.text()).toBe('S');
+    });
+
+    it('should display full name when initialized as active', async () => {
```
qmonmert, line 21:
```suggestion
    it('should display full name when initialized as active', () => {
```

src/test/webapp/unit/module/primary/landscape/LandscapeRankModuleFilterComponent.spec.ts
renanfranca: I couldn't remove the async from this test because it uses await inside:
```java
  it('should display buttons without colors when a selected rank is deselected', () => {
        const wrapper = wrap();

    const rankAButton = wrapper.find(wrappedElement('rank-RANK_A-filter'));
await rankAButton.trigger('click');
await rankAButton.trigger('click');
```

qmonmert (dismissed):

PR #12253, opened 2025-04-05
Rewrite some tests with it.each

qmonmert:


murdos (commented):
I'm not convinced, tests are much less readable and comprehensive.

pascalgrimaud:
Agree with @murdos, I think the new version is less readable, so I prefer to keep the current version

qmonmert:
ok I close

Issue #12038, opened 2025-03-05
Integrate Spring Boot's Docker Compose support


murdos:
Add Docker Compose support so dependent containers are started automatically: https://docs.spring.io/spring-boot/reference/features/dev-services.html#features.dev-services.docker-compose

This is a better developer experience because you do t have to start anything manually. 
It might help simplify our CI jobs too.

pascalgrimaud:
@murdos : do you think it should be a new module? and not update the existing ones

murdos:
> @murdos : do you think it should be a new module? and not update the existing ones

I would update existing ones, since you should add each new service (keycloack, database, redis, ...) in a central docker-compose.yml.

murdos:
BTW not all docker service should be added IMO to the  docker-compose that Spring Boot will use. E.g. sonar should not be started by Spring Boot.

adem-hassine:
Hi @murdos @pascalgrimaud,

I’ve reviewed and incorporated all the mentioned details for the integration. As this is would be the initial draft PR, I’d like to gather your feedback on the overall approach and kindly request your guidance, especially regarding the implementation of tests & coverage.

To summarize, here are the key points I'd like to highlight:

- Added an additional handler to be applied.
- Each module requiring Spring Boot Docker Compose integration must explicitly include the relevant Java dependency.

PR #12288, opened 2025-04-10
Integrate Spring Boot's Docker Compose support

adem-hassine:
Closes https://github.com/jhipster/jhipster-lite/issues/12038


adem-hassine:
Hi @murdos,

I wanted to highlight the changes compared to the previous PR:

- New CI Test Script : I’ve defined a new test CI script where the build tool wrapper is used instead of [java -jar](https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start.sh#L28) to run the server. As since docker-compose is considered a development tool, it is not included when generating the JAR. This necessitated the creation of a new script file. 

-  Readiness Timeout Fixes : Some services were experiencing readiness timeout issues. To resolve this, I updated their docker-compose configurations to include proper health checks and disabled the default Spring Docker readiness behavior.

murdos:
Hi @adem-hassine, thanks for the detailed report!
 
> I wanted to highlight the changes compared to the previous PR:
> 
>     * New CI Test Script : I’ve defined a new test CI script where the build tool wrapper is used instead of [java -jar](https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start.sh#L28) to run the server. As since docker-compose is considered a development tool, it’s excluded when generating the JAR, along with the root Docker Compose file. This necessitated the creation of a new script file.

I think we should make a choice between testing generated applications in development/local conditions, or in conditions similar to the production. In generator-jhipster there were some ci tests that were testing dev, other prod.
I don't think we should bother with that, and I think we should only test the generated jar, as before.
@pascalgrimaud : WDYT?

>     * Readiness Timeout Fixes : Some services were experiencing readiness timeout issues. To resolve this, I updated their docker-compose configurations to include proper health checks and disabled the default Spring Docker readiness behavior.

Great! You made a huge job here too! :)

pascalgrimaud:
I agree with @murdos :
- this feature should not change the current behavior of the CI
- the CI should use java -jar, instead of ./mvnw
- so we can still use the current script to start manually the différent service

Why? Because testing ./mvnw is not relevant, it will be used only locally and the important part is in Real environnement
So we should focus on java jar

adem-hassine:
Hi @murdos , 

No problem at all and thanks for the feedback! 
> > ```
> > 1. For now, I’ve rolled back the changes related to the CI and kept only the `java -jar` execution in the ci. That said, this means future CI tests won’t run and test the dev mode.
> > ```
> 
> Yeah, we discussed that with @pascalgrimaud, and we assume it.
> 
Got it, thanks for confirming! 
> > ```
> > 2. After this rollback, I noticed something unexpected with the Gradle app: the Docker Compose dependency remains available on the JAR classpath. This led me to look for a way to prevent that behavior. However, since the developmentOnly scope is currently not supported by the [GradleCommandHandler.java](https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/module/infrastructure/secondary/javadependency/gradle/GradleCommandHandler.java#L232), I wanted to get your opinion on how best to proceed. For better visibility, I’ve pushed a sample generated from this PR and added a [small commit ](https://github.com/adem-hassine/jhipster-lite-fullappgradle-sample/commit/7cbe1f1f0dc11fb92ca62c46a987adb11d750a49) demonstrating how it can work for both development and production JAR scenarios.
> > ```
> 
> This seems to be a broader issue with java dependency scopes and their mapping to maven and gradle, that I think we should address separately. I think we should exclude gradle for now (more on that below) in order to not block the work on docker-compose.
> 
> During our last discussion with @pascalgrimaud, I changed a bit my mind, and I think it would be better if:
> 
> * each module that generates a docker compose file would contribute to the global `docker-compose.yml`
> * a new module for the spring boot and docker compose integration was added (it would add the dependency and spring configuration).
>   That way spring boot and docker compose could be an opt-in feature.
> 
> To circumvent the gradle issue, we would enforce a dependency to maven-java for the new module for the spring boot+docker compose integration, until the gradle issue is resolved.
> 
> I'm sorry for the change of direction. If you want, I can iterate over your branch and finish the work. Just let me know.
> 

Noted, no worries as this approach will certainly lead to a better implementation. I'm enjoying the process of contributing to JHipster Lite and am ready to proceed with the necessary changes.
> > ```
> > 3. The dependency on the Cassandra service in the migration Docker Compose file caused an issue. Since the Compose file is [launched separately](https://github.com/jhipster/jhipster-lite/blob/main/tests-ci/start_docker_compose.sh#L47), it creates a conflict when running the [Cassandra migration service](https://github.com/jhipster/jhipster-lite/pull/12288/files#diff-7883dd514d5619f69e09b3a33a78f9518f44ff99e57c6d2a59d27d578a416246R17). To resolve this, I created a copy of the Docker file to separate the one used by spring-boot-docker-compose from the one used manually.
> > ```
> 
> How about launching docker services in the CI from the global docker compose file? This could both simplify the CI scripts, and avoid the double docker file for cassandra?

Yes, we can use the global Docker Compose file and update the CI accordingly. However, maintaining a single file that references an external Docker Compose file will still introduce inconsistencies in the startup commands, e.g.: running the [Cassandra service](https://github.com/jhipster/jhipster-lite/blob/677a64c31ae55a33aaf15fa348d95b25141d562b/src/main/java/tech/jhipster/lite/generator/server/springboot/database/cassandra/domain/CassandraModuleFactory.java#L52) using the script provided in the README will fail.

murdos:
Thanks so much @adem-hassine for your hard work!
Don't forget to claim your bounty, you deserve it :)

adem-hassine:
Hi @murdos ,
I'm really glad to have made my first feature contribution to JHipster Lite — thank you for your time and valuable insights. I hope to have more opportunities to collaborate on future issues!

I've claimed the [bounty](https://opencollective.com/generator-jhipster/expenses/250234).
Much respect 🙏,

pascalgrimaud:
@adem-hassine : approved :-)

PR #12306, opened 2025-04-12
Improve coverage

qmonmert:
Before
image
image

After
image
image


murdos (changes requested):
I don't think that these changes improve the quality of code.
If you want to improve frontend coverage, I encourage you to try to make a few more component tests with cypress, that should naturally improve coverage.

PR #12323, opened 2025-04-14
Infinitest available on IntelliJ

qmonmert:
https://infinitest.github.io/

image
image


murdos:
@qmonmert : have you tried it? In my experience it's pretty unstable, so I'm not sure we should recommend it.

PR #12378, opened 2025-04-25
Rework on tests

qmonmert:
Same as https://github.com/jhipster/jhipster-lite/blob/dd576d2e5d988df21cc0a57487ca84099665d548/src/test/java/tech/jhipster/lite/module/domain/properties/JHipsterModulePropertiesTest.java#L23

image

### DIFF
--- a/src/test/java/tech/jhipster/lite/generator/client/tikui/domain/TikuiModuleFactoryTest.java
+++ b/src/test/java/tech/jhipster/lite/generator/client/tikui/domain/TikuiModuleFactoryTest.java
@@ -17,7 +17,7 @@ class TikuiModuleFactoryTest {
   private static final TikuiModuleFactory factory = new TikuiModuleFactory();
 
   @Nested
-  class PrettierTest {
+  class Prettier {
 
     @Test
     void shouldBuildModuleWithPugFormat() {
@@ -45,7 +45,7 @@ class TikuiModuleFactoryTest {
   }
 
   @Nested
-  class VueTest {
+  class Vue {
 
     @Test
     void shouldBuildModuleOnVueProject() {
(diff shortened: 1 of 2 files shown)


src/test/java/tech/jhipster/lite/generator/client/tikui/domain/TikuiModuleFactoryTest.java
```diff
@@ -17,6 +18,7 @@ class TikuiModuleFactoryTest {
   private static final TikuiModuleFactory factory = new TikuiModuleFactory();
 
   @Nested
+  @DisplayName("Prettier")
   class PrettierTest {
```
murdos, line 22: There 's a much simpler solution: rename the class and remove the 'Test' suffix.

There's no need to use a Test suffix for nested test classes

murdos (changes requested):

src/test/java/tech/jhipster/lite/generator/client/tools/cypress/domain/CypressModuleFactoryTest.java
```diff
@@ -16,6 +17,7 @@ class CypressModuleFactoryTest {
   private static final CypressModuleFactory factory = new CypressModuleFactory();
 
   @Nested
+  @DisplayName("Component")
```
murdos, line 20: The full name is "component tests", not "component".

Here I think we should just remove the nested classes, and just have the 2 tests at top level, they don't bring much value.

murdos (approved):

murdos:
Thanks @qmonmert :)

qmonmert:
> Thanks @qmonmert :)

 🤝

PR #12419, opened 2025-04-29
Remove unnecessary public visibility

qmonmert:


murdos:
👍 what would be great is an archunit test that would enforce this practice.

PR #12430, opened 2025-05-01
Rename: SimpleCache

qmonmert:


src/main/java/tech/jhipster/lite/generator/server/springboot/cache/simple/infrastructure/primary/SimpleCacheModuleConfiguration.java
```diff
@@ -11,7 +11,7 @@
 import tech.jhipster.lite.module.domain.resource.JHipsterModuleResource;
 
 @Configuration
-class CacheModuleConfiguration {
+class SimpleCacheModuleConfiguration {
 
   @Bean
   JHipsterModuleResource simpleCacheModule(SpringBootCacheSimpleApplicationService caches) {
```
murdos, line 17: Then I guess you'll come later to rename the service. Could you do it in the same PR?
```suggestion
  JHipsterModuleResource simpleCacheModule(SimpleCacheApplicationService caches) {
```

qmonmert: done

PR #12431, opened 2025-05-01
Rename: Ehcache

qmonmert:
image


qmonmert:
> Well, according to https://www.ehcache.org/, the proper case is "Ehcache". So this one was correct.

good catch

PR #12436, opened 2025-05-01
Rename: SpringBootMvcModuleFactory

qmonmert:


pascalgrimaud:
the reason for Mvcs was because there is Tomcat and Undertow here. But honestly, both names are OK, so I don't care

qmonmert:
image
@pascalgrimaud it's to have the same nomenclature for service, factory and configuration

PR #12443, opened 2025-05-02
Rename Spring boot docker to Docker

qmonmert:


src/main/java/tech/jhipster/lite/generator/server/springboot/database/redis/domain/RedisModuleFactory.java
```diff
@@ -101,7 +101,7 @@ private JavaDependency testContainerDependency() {
       .build();
   }
 
-  private JavaDependency springBootDockerComposeIntegration() {
+  private JavaDependency dockerComposeIntegration() {
```
murdos, line 104: Previous name was more explicit, and is more similar to the artifiactId.

src/main/java/tech/jhipster/lite/generator/server/springboot/docker/domain/DockerModuleFactory.java
```diff
@@ -14,7 +14,7 @@
 import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
 import tech.jhipster.lite.shared.error.domain.Assert;
 
-public class SpringBootDockerModuleFactory {
+public class DockerModuleFactory {
```
murdos, line 17: What is the problem with the current name? The configuration contains things that are specific to Spring, like SPRING_OUTPUT_ANSI_ENABLED environment variable.

murdos (changes requested):

PR #12488, opened 2025-05-08
Remove commas

qmonmert:


pascalgrimaud:
removed 'area: enhancement label', as same behavior in the code

PR #12506, opened 2025-05-09
Rename packages

qmonmert:
Related to #12442

### DIFF
--- a/src/main/java/tech/jhipster/lite/generator/client/hexagonaldocumentation/application/FrontHexagonalArchitectureDocumentationApplicationService.java
+++ b/src/main/java/tech/jhipster/lite/generator/client/hexagonaldocumentation/application/FrontHexagonalArchitectureDocumentationApplicationService.java
@@ -1,7 +1,7 @@
-package tech.jhipster.lite.generator.client.hexagonal_documentation.application;
+package tech.jhipster.lite.generator.client.hexagonaldocumentation.application;
 
 import org.springframework.stereotype.Service;
-import tech.jhipster.lite.generator.client.hexagonal_documentation.domain.FrontHexagonalDocumentationModuleFactory;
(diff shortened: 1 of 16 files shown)


pascalgrimaud:
honestly, I'd prefer with `_` because it's more readable, but I don't care so we can merge it

murdos:
> honestly, I'd prefer with `_` because it's more readable, but I don't care so we can merge it

For the record, I do prefer too `_` for the same reason.

renanfranca:
@qmonmert : Do you have a strong opinion about not using the underscore "_"? If so, can you share it, please?

qmonmert:
@renanfranca I have not a strong opinion but I saw that https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names but I'm open for others choices

qmonmert:
@renanfranca feel free to revert but I didn't see a lot of Open Source Project with underscore in the package name

renanfranca:
>  @renanfranca feel free to revert 

Thanks, I will revert as soon as possible :)


> but I didn't see a lot of Open Source Project with underscore in the package name

For me, that is the first project, which uses underscore at package name. But it is the first project that does not use the prefix `get` and not use verbs in some java methods. After a while, it makes so much sense!

That's makes me not use `get` or verbs on private methods at my work. I don't have the opportunity to talk with all my colleagues to make it the standard for the codebase.

pascalgrimaud:
No need to revert back to underscore
No need to remove other package with underscore

If we want to do this, let's discuss first then let's code a new arch unit test, to appli this new rule. Otherwise, let's stay like this

We should focus on code that brings values to our user

renanfranca:
> If we want to do this, let's discuss first then let's code a new arch unit test, to appli this new rule. Otherwise, let's stay like this

I created the following discussion https://github.com/jhipster/jhipster-lite/discussions/12594

PR #12507, opened 2025-05-09
Rename logger

qmonmert:
Standardize
image


pascalgrimaud:
update label, as it's not good

PR #12558, opened 2025-05-15
Thymeleaf/Tests rename files

qmonmert:


pascalgrimaud:
@murdos @qmonmert : missing labels here. You can have a look at [release-drafter.yml](https://github.com/jhipster/jhipster-lite/blob/main/.github/release-drafter.yml) to see how it works. If you have some questions, just ask :smile: 

To make a good release note:
- each PR should be labeled with one and unique area
  - features: it's a something new for our user
  - enhancement: improvement of the code
  - refactoring: nothing change, it's just refactoring
  - bug: bug fixes
- then, we can apply one or many other labels, related to "theme"

image

murdos:
@pascalgrimaud : you're absolutely right, sorry for not having checked labels! I'll be more careful now :)

PR #12567, opened 2025-05-17
Rework Thymeleaf module

qmonmert:


src/main/java/tech/jhipster/lite/generator/server/springboot/thymeleaf/springboot/application/SpringBootThymeleafApplicationService.java
```diff
@@ -1,7 +1,7 @@
-package tech.jhipster.lite.generator.server.springboot.thymeleaf.application;
+package tech.jhipster.lite.generator.server.springboot.thymeleaf.springboot.application;
 
 import org.springframework.stereotype.Service;
-import tech.jhipster.lite.generator.server.springboot.thymeleaf.domain.SpringBootThymeleafModuleFactory;
+import tech.jhipster.lite.generator.server.springboot.thymeleaf.springboot.domain.SpringBootThymeleafModuleFactory;
```
renanfranca, line 4: Can you, please, share the reason to add `.springboot` again at the package name?

qmonmert: we can't add package-info in `template` folder without updating the structure like that. The package name `.springboot` is maybe not the best ...

renanfranca: Thanks! I can see it now 😉 

What do you think of the name package name `.init` instead of `.spring`?

qmonmert: I updated with `.init` 🙂

renanfranca (approved):
Thank you for your work @qmonmert 😃

PR #12582, opened 2025-05-19
Angular: rename OAuth2 to OAuth2Keycloak

qmonmert:
Same as Vue
https://github.com/jhipster/jhipster-lite/blob/main/src/main/java/tech/jhipster/lite/generator/client/vue/security/oauth2_keycloak/application/VueOAuth2KeycloakApplicationService.java


src/main/java/tech/jhipster/lite/shared/slug/domain/JHLiteModuleSlug.java
```diff
@@ -18,7 +18,7 @@ public enum JHLiteModuleSlug implements JHipsterModuleSlugFactory {
   ANGULAR_CORE("angular-core", RANK_A),
   ANGULAR_HEALTH("angular-health", RANK_C),
   ANGULAR_JWT("angular-jwt", RANK_C),
-  ANGULAR_OAUTH_2("angular-oauth2", RANK_C),
```
pascalgrimaud, line 21: I think we should not change the slug unless it's totally necessary, as it's a breaking change

qmonmert: @pascalgrimaud ok I update

PR #12637, opened 2025-05-26
Standardize feature name

qmonmert:


murdos:
I'm not convinced that having explicitly "module" was necessary.

murdos:
@qmonmert : I'm not sure there's a need to revert. But yeah I think it was implicit, and there's not necessarily the need to always add the word "module"

PR #12797, opened 2025-06-18
The surrounding quotes actually unquote this

qmonmert:
image


src/main/resources/generator/server/springboot/custom-jhlite/tests-ci/generate.sh.mustache
```diff
@@ -30,7 +30,7 @@ applyModules() {
       -H "accept: */*" \
       -H "Content-Type: application/json" \
       -d "$payload" \
-      "http://localhost:{{serverPort}}""$api""
+      "http://localhost:{{serverPort}}"$api"
```
murdos, line 33: Are you sure? I believe there's an extra quote here.

```suggestion
      "http://localhost:{{serverPort}}$api"
```

Also why are we not aligned with the generate.sh of the generator?

PR #12928, opened 2025-07-12
Remove duplicated entry in gitattributes

qmonmert:


pascalgrimaud:
@qmonmert : maybe I'm tired but I dont see the other txt? Only the one you deletes

qmonmert:
@pascalgrimaud https://github.com/jhipster/jhipster-lite/blob/57fa579791932ae723d9251c30bd794406f89b35/src/main/resources/generator/init/gitattributes#L55

qmonmert:
@pascalgrimaud 
image

PR #12940, opened 2025-07-14
Add sonar-typescript module

qmonmert:
Fix #12735


qmonmert:
@pascalgrimaud is this what you expected?

src/main/java/tech/jhipster/lite/generator/typescript/sonar/domain/SonarTypescriptModuleFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import static tech.jhipster.lite.module.domain.nodejs.JHLiteNodePackagesVersionSource.COMMON;
+
+import tech.jhipster.lite.module.domain.JHipsterModule;
+import tech.jhipster.lite.module.domain.docker.DockerImages;
+import tech.jhipster.lite.module.domain.file.JHipsterDestination;
+import tech.jhipster.lite.module.domain.file.JHipsterSource;
+import tech.jhipster.lite.module.domain.properties.JHipsterModuleProperties;
+import tech.jhipster.lite.shared.error.domain.Assert;
+
+public class SonarTypescriptModuleFactory {
```
murdos, line 13: IMO this should be in the existing `SonarQubeModuleFactory`, and under the same feature, that should probably be renamed from `sonar-qube-java` to `sonar-qube`.

The main reason is that you don't want to apply both `sonar-typescript` and `sonar-qube-java-backend` (or `sonar-qube-java-backend-and-frontend`), because they generate the same files (`sonar.md`, `sonar-project.properties`, ...)

qmonmert: issue named was "Add sonar-typescript module" so I did a new module.
@pascalgrimaud WDYT? agree with @murdos ?

qmonmert: Indeed some files are common but the way to launch Sonar is different so for me two modules seems good...

pascalgrimaud:
@qmonmert : I'm reviewing it. Can you resolve conflict plz @qmonmert ?

src/main/java/tech/jhipster/lite/generator/typescript/sonar/domain/SonarTypescriptModuleFactory.java
murdos: Yes, but because of the common files, you should not be able to apply both.

murdos: > issue named was "Add sonar-typescript module" so I did a new module. 

Well, there are multiple generators (aka. an  hexagon with a factory and a configuration) that provides multiples modules.
So for me a new module does mean new generator/hexagon.

pascalgrimaud: I'm letting you take care of my comment @qmonmert
Then, I'll take care of the refactoring

pascalgrimaud:
Thanks, I'm merging this and will do the refactoring

qmonmert:
thanks @pascalgrimaud

PR #13079, opened 2025-08-06
Rename seed4j-needle

qmonmert:


murdos:
Great, thanks @qmonmert !
I think we should also provide an openrewrite recipe that will rename needles in existing projects.

PR #13164, opened 2025-08-21
use vitest on angular core instead of jest

### LINKED ISSUE #12783: Angular: switch to vitest
With Angular 20, there's no an experimental support for vitest.

I would suggest to use it, in order to have all our clients use vitest.
This would also help to make `angular-core` depends on the `typescript` module.

fabienpuissant:
https://angular.fr/unit-tests/vitest

I tried the native but there is a lot of errors due to environment tests and zone.js

The second implementation with analogjs works fine

But I'm struggling  hard with the tests generated on different modules..
If anyone can help with it bc I tried like everything to mock window location (definePropery, spyOn, globalStub) but I always get Cannot redefine property: location 

NB : In this PR, the angular-core module does not depend on typescript it is only preparing the transition to fix #12783


wmarques:
I think using analogjs implementation is fine for now as the official support is not yet stable. We can do the switch after and it should be pretty straightforward (just removing analogjs dependency from package.json and change some config).

qmonmert:
For me we should wait for official support Angular/Vitest (without analogjs). There is no emergency.

murdos:
Closing this PR since migration has been done as part of migration to Angular 21 (#14158)

PR #13232, opened 2025-09-04
Docker Hub: publish main

pascalgrimaud:
- related to https://github.com/seed4j/seed4j/issues/13170

it should publish a new Docker image when we merge new code to main branch


pascalgrimaud:
let's try this

PR #13246, opened 2025-09-06
Add root folder restriction in forced project configuration

### LINKED ISSUE #13244: Bug: should not allow generate project at root of forced project folder
When using this property:

```
seed4j:
  forced-project-folder: /tmp/seed4j
```

We should not allow to generate project at root folder

clinm:
fixes #13244


pascalgrimaud:
Thanks @clinm :-)

PR #13266, opened 2025-09-09
Delete deploy.yml github workflow

clinm:
Fixes #13258


clinm:
Easy enough to make me suspicious ;)

PR #13339, opened 2025-09-20
Replace start-server-and-test with concurrently and wait-on

qmonmert:


package.json
```diff
@@ -37,8 +37,8 @@
     "preview": "vite preview",
     "start": "npm run dev",
     "test": "npm run test:watch",
-    "test:component": "start-server-and-test dev http://localhost:9000 'cypress open --e2e --config-file src/test/webapp/component/cypress-config.ts'",
-    "test:component:headless": "start-server-and-test dev http://localhost:9000 'cypress run --headless --config-file src/test/webapp/component/cypress-config.ts'",
+    "test:component": "concurrently -k -s first -n dev,test \"npm run dev\" \"wait-on http://localhost:9000 && cypress open --e2e --config-file src/test/webapp/component/cypress-config.ts\"",
+    "test:component:headless": "concurrently -k -s first -n dev,test \"npm run dev\" \"wait-on http://localhost:9000 && cypress run --headless --config-file src/test/webapp/component/cypress-config.ts\"",
```
pascalgrimaud, line 41: I think you could replace `\"` by simple quote `'`

pascalgrimaud: I'm merging this and will apply the fix in another PR !

PR #13349, opened 2025-09-21
replace to single quote for test:component:x

pascalgrimaud:
- following https://github.com/seed4j/seed4j/pull/13339


pascalgrimaud:
Oh Windows doesn't like this, so let's stay with the original code from @qmonmert

PR #13371, opened 2025-09-23
ci: build and test with java 25 by default

murdos:


murdos:
It's a bit too early for IMO requiring java 25 for running the seed4j engine and generating apps requiring java 25.

I think we should wait a few months, to make sure it's available in main distributions.

PR #13421, opened 2025-10-01
Add Tailwind CSS module for Angular

wmarques:
I don't know if it is needed but I use Tailwind in all my projects, so I thought I'll add Tailwind support into the project 😄 

Don't hesitate to close this if you think it's overkill or not needed !


qmonmert:
Add a `angular-tailwind.feature` 😉

wmarques:
Thanks @qmonmert , should be good now

qmonmert (approved):
👍

murdos:
@wmarques : you also need to add it to the `generate.sh` script, so the module is actually tested during CI, e.g. here: https://github.com/seed4j/seed4j/blob/main/tests-ci/generate.sh#L418

PR #13431, opened 2025-10-02
Move to Java 25

NolwennD:
It seems that integration of this PR should wait https://github.com/devcontainers/images/pull/1566 to be included in a next version of devcontainers but I’m not sure.

Changes of `StringReader` to `Reader` come from  modernizer.


renanfranca:
Thank you for your work @NolwennD 😃

PR #14387, opened 2026-05-14
Improve pipeline speed

clinm:
- add cache for playwright browsers
- use alpine image for docker image retrieving the sonar token

Result : https://github.com/seed4j/seed4j/actions/runs/25864602660/usage?pr=14387

- Everything under 10 minutes
- Longest part on average is the sonar image download (https://github.com/seed4j/seed4j/pull/14388)

PR #14388, opened 2026-05-14
ci: cache Sonar Docker image across matrix runs for generated apps

murdos:
Add restore/load/save steps around the 'Test: starting Sonar' step so that the sonarqube image (~500 MB) is pulled from Docker Hub only once per unique Dockerfile hash, then reused by all 24 matrix runners via actions/cache@v5.


clinm:
I'm glad you took that part ! I considered it but had no idea where to start :)

murdos:
Closing: this was an experiment inspired by #14387

PR #14420, opened 2026-05-24
Fix workflow and cache issues

clinm:
# Windows issue
The tests-windows job has some issues : 
image

One correlation factor is the use of previous cache. All jobs I checked had the error when the cache was missing.
The problem is that I cannot find anything related to the cypress cache in previous commit

After some research, I found https://github.com/cypress-io/cypress/pull/33887 and rolling back to node 24.15.0 does fix the job https://github.com/seed4j/seed4j/actions/runs/26359102398/job/77591255298?pr=14420

Node LTS release made May 21 : https://nodejs.org/fr/blog/release/v24.16.0

# Linux issue
Tests are flaky

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -8,7 +8,8 @@ inputs:
   node-version:
     description: 'Node version to use'
     required: true
-    default: 'lts/*'
+    default: '24.15.0'
+    # Pinned to 24.15.0 until https://github.com/cypress-io/cypress/pull/33887 is released. Then 'lts/*' can be used again
   node-package-manager:
     description: 'Package manager to use for installing node dependencies'
     required: true
--- a/src/test/webapp/component/integration/Landscape.spec.ts
+++ b/src/test/webapp/component/integration/Landscape.spec.ts
@@ -5,6 +5,7 @@ describe('Landscape', () => {
   beforeEach(() => cy.intercept({ path: '/api/project-folders' }, { body: '/tmp/seed4j/1234' }));
 
   it('should change theme after toggle switch theme button', () => {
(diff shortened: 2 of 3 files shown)


clinm:
@murdos 
For Windows : Cypress team already has a fix for the issue, i guess it is a matter of days before a release. We could wait ?
For Linux : I'm not sure my changes fixe once for all the flaky tests, but i ran around 30 times without failure. I can make another PR clean up this one, as you wish

.github/actions/setup/action.yml
```diff
@@ -8,7 +8,7 @@ inputs:
   node-version:
     description: 'Node version to use'
     required: true
-    default: 'lts/*'
+    default: '24.15.0'
```
murdos, line 11: It's a good practice to leave a comment so we don't forget to revert this change once the upstream issue is fixed. Something like:

```suggestion
    # Pinned to 24.15.0 until https://github.com/cypress-io/cypress/pull/33887 is released. Then 'lts/*' can be used again
    default: '24.15.0' 
```

.github/workflows/github-actions.yml
```diff
@@ -124,8 +124,15 @@ jobs:
         run: docker compose -f src/main/docker/sonar.yml up -d
       - name: 'Test: run backend tests'
         run: |
-          chmod +x mvnw
-          ./mvnw clean verify -Dsonar.qualitygate.wait=true
```
murdos, line 128: All backend tests are not ran anymore. I guess this was for debug, and you're going to revert that?

murdos (commented):
Good catch! 
I think we can fix `main` with these fixes without waiting for a cypress release.

clinm:
@murdos Linux tests are still flaky, but i guess it's better than nothing

murdos (approved):

PR #14619, opened 2026-07-26
update: interactiveMode already set to false

clinm:
See : https://github.com/actions/setup-java/issues/547

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -37,10 +37,6 @@ runs:
         distribution: 'temurin'
         java-version: '${{ inputs.java-version }}'
         check-latest: true
-    - name: 'Disable maven interactive mode'
-      shell: bash
-      run: |
-        sed -i 's/<servers>/<interactiveMode>false\<\/interactiveMode>\n  <servers>/g' ~/.m2/settings.xml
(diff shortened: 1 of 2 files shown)


murdos (dismissed):

murdos:
Yeah, I noticed the other day that my feature request on action/java finally got fixed.
Thanks for applying this change 🙂

NolwennD:
`generate-main-branch-jar` job fails with a duplicated key error because it checkout `main` and the `sed` fix is still present!

renanfranca:
> `generate-main-branch-jar` job fails with a duplicated key error because it checkout `main` and the `sed` fix is still present!

@murdos : I think one way out is to use the [ci ignore] in this PR title, so the pipeline will be skipped. Do you agree?

Here are the references:
- https://github.com/seed4j/seed4j/blob/main/CONTRIBUTING.md#-git-commit-guidelines
- https://github.com/seed4j/seed4j/blob/d6699b9a6ade03515d54d15e4e1a3327e7bfe40b/.github/workflows/github-actions.yml#L43-L65

murdos:
I've just merged this branch as is. Let's see if it fixes main branch.

PR #14626, opened 2026-07-30
Fix error prone warnings (ep2)

NolwennD:
Same as the previous PR (#14621) but on a larger scope.

Methodology will be explain (by Claude) in the first comment.

TLDR version: generate an app (from CI list), run error prone analysis with javac + options, fix warnings in templates, generate again, mvn verify and so on!

### DIFF
--- a/src/main/resources/generator/server/pagination/domain/main/AppPageable.java.mustache
+++ b/src/main/resources/generator/server/pagination/domain/main/AppPageable.java.mustache
@@ -5,7 +5,7 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 import {{packageName}}.shared.error.domain.Assert;
 import {{packageName}}.shared.generation.domain.ExcludeFromGeneratedCodeCoverage;
 
-public class {{baseName}}Pageable {
+public final class {{baseName}}Pageable {
 
   private final int page;
   private final int pageSize;
--- a/src/main/resources/generator/server/pagination/rest/main/RestAppPage.java.mustache
+++ b/src/main/resources/generator/server/pagination/rest/main/RestAppPage.java.mustache
@@ -91,43 +91,43 @@ public final class Rest{{baseName}}Page<T> {
       this.content = content;
     }
 
-    public Rest{{baseName}}PageBuilder<T> pageSize(int pageSize) {
+    private Rest{{baseName}}PageBuilder<T> pageSize(int pageSize) {
       this.pageSize = pageSize;
 
       return this;
     }
 
-    public Rest{{baseName}}PageBuilder<T> currentPage(int currentPage) {
+    private Rest{{baseName}}PageBuilder<T> currentPage(int currentPage) {
       this.currentPage = currentPage;
 
       return this;
     }
 
-    public Rest{{baseName}}PageBuilder<T> totalElementsCount(long totalElementsCount) {
+    private Rest{{baseName}}PageBuilder<T> totalElementsCount(long totalElementsCount) {
       this.totalElementsCount = totalElementsCount;
 
       return this;
     }
 
-    public Rest{{baseName}}PageBuilder<T> pageCount(int pageCount) {
+    private Rest{{baseName}}PageBuilder<T> pageCount(int pageCount) {
       this.pageCount = pageCount;
 
       return this;
     }
 
-    public Rest{{baseName}}PageBuilder<T> hasPrevious(boolean hasPrevious) {
+    private Rest{{baseName}}PageBuilder<T> hasPrevious(boolean hasPrevious) {
       this.hasPrevious = hasPrevious;
 
       return this;
     }
 
-    public Rest{{baseName}}PageBuilder<T> hasNext(boolean hasNext) {
+    private Rest{{baseName}}PageBuilder<T> hasNext(boolean hasNext) {
       this.hasNext = hasNext;
 
       return this;
     }
 
-    public Rest{{baseName}}Page<T> build() {
+    private Rest{{baseName}}Page<T> build() {
       return new Rest{{baseName}}Page<>(this);
     }
   }
--- a/src/main/resources/generator/server/springboot/cache/ehcache/test/JavaCacheConfigurationIT.java.mustache
+++ b/src/main/resources/generator/server/springboot/cache/ehcache/test/JavaCacheConfigurationIT.java.mustache
@@ -9,14 +9,10 @@ import org.ehcache.config.ResourceType;
 import org.ehcache.jsr107.Eh107Configuration;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.ApplicationContext;
 import {{packageName}}.IntegrationTest;
 
 @IntegrationTest(properties = { "application.cache.ehcache.max-entries=1000", "application.cache.ehcache.time-to-live-seconds=86400" })
 class CacheConfigurationIT {
-
-  @Autowired
-  private ApplicationContext applicationContext;
 
   @Autowired
   private CacheConfiguration cacheConfiguration;
@@ -37,6 +33,6 @@ class CacheConfigurationIT {
 
     assertThat(config.getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getSize()).isEqualTo(1000);
 
-    assertThat(config.getExpiryPolicy().getExpiryForCreation(42L, "entry").getSeconds()).isEqualTo(86400);
+    assertThat(config.getExpiryPolicy().getExpiryForCreation(42L, "entry").toSeconds()).isEqualTo(86400);
   }
 }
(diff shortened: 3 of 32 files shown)


NolwennD:
# Methodology: Auditing ErrorProne Warnings in Generated (Mustache-Templated) Code

## Problem statement

Seed4J itself compiles with ErrorProne + NullAway enabled (see `pom.xml`'s `maven-compiler-plugin` configuration and `.mvn/jvm.config`). However, the Spring Boot applications it *generates* — produced from `.mustache` templates under `src/main/resources/generator/` — do **not** have ErrorProne wired into their own `pom.xml`/`build.gradle`. This means any bug pattern ErrorProne would catch in Seed4J's own code goes completely unchecked in the code it hands out to end users.

The goal of this audit is to run the exact same ErrorProne analysis against real generated output, trace every finding back to the originating `.mustache` template, and fix the template — without modifying the generated project's build configuration itself (since we can't assume any given user will ever enable ErrorProne on their generated app).

## Step 1 — Generate a real application

Build Seed4J, start it as a server, and drive it through its own REST API (the same mechanism the CI uses) to materialize a real application on disk:

```bash
./mvnw package -DskipTests
java -jar target/seed4j-*-exec.jar --server.port=1339 &
./tests-ci/generate.sh <app-name> maven yaml npm   # e.g. fullappmaven, oauth2app, cassandraapp...
```

This produces a standalone Maven (or Gradle) project under `/tmp/seed4j/<app-name>` with its own `pom.xml` and dependency tree.

## Step 2 — Resolve two classpaths

**a) The generated app's own compile classpath** (its dependencies are resolved from its own `pom.xml`, independently of Seed4J's):

```bash
cd /tmp/seed4j/<app-name>
mvn dependency:build-classpath -Dmdep.outputFile=app_cp.txt
```

**b) The ErrorProne "processor path"** — the annotation-processor-style classpath containing ErrorProne itself and NullAway, resolved once (their versions come from Seed4J's own `pom.xml`, currently `error_prone_core:2.50.0` and `nullaway:0.13.7`) via a disposable throwaway `pom.xml`:

```xml
<dependencies>
  <dependency><groupId>com.google.errorprone</groupId><artifactId>error_prone_core</artifactId><version>2.50.0</version></dependency>
  <dependency><groupId>com.uber.nullaway</groupId><artifactId>nullaway</artifactId><version>0.13.7</version></dependency>
</dependencies>
```
```bash
mvn dependency:build-classpath -Dmdep.outputFile=processor_cp.txt
```

## Step 3 — Compile with `javac` directly, mirroring Seed4J's own compiler configuration exactly

This is the crux of the technique: instead of touching the generated project's build, invoke the JDK's `javac` binary directly, replaying the identical flags Seed4J's own `maven-compiler-plugin` uses (`compilerArgs` in `pom.xml`) plus the JVM-level flags from `.mvn/jvm.config` (required because ErrorProne hooks into `javac`'s internal APIs, which are otherwise inaccessible since JDK 16+):

```bash
javac \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.processing=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED \
  -J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED \
  -J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED \
  -XDcompilePolicy=simple \
  --should-stop=ifError=FLOW \
  --release 25 \
  -parameters \
  -classpath "$(cat app_cp.txt)" \
  -processorpath "$(cat processor_cp.txt)" \
  "-Xplugin:ErrorProne \
    -Xep:BadImport:OFF -Xep:UnnecessaryLambda:OFF -Xep:UnnecessaryStringBuilder:OFF -Xep:ImmutableEnumChecker:OFF \
    -Xep:NullAway:ERROR \
    -XepOpt:NullAway:JSpecifyMode=true \
    -XepOpt:NullAway:OnlyNullMarked=true \
    -XepOpt:NullAway:CustomContractAnnotations=com.seed4j.shared.nullness.domain.Contract \
    -XepOpt:NullAway:ExcludedFieldAnnotations=com.seed4j.Logs,org.mockito.Mock" \
  -d out/main \
  @main_sources.txt        # a text file listing every .java path, one per line (find src/main/java -name "*.java" > main_sources.txt)
```

Run it twice: once for `src/main/java`, once for `src/test/java` (the second invocation adds the first run's `-d` output directory to `-classpath`, so test code can resolve main classes).

Note the two `javac` gotchas discovered along the way:
- `-Xplugin:ErrorProne` alone fails with `InvalidCommandLineOptionException` unless `-XDcompilePolicy=simple` is also passed (ErrorProne doesn't support the default `by-todo` compile policy).
- NullAway refuses to start unless either `AnnotatedPackages` or `OnlyNullMarked` is set — Seed4J's own config uses the latter.

## Step 4 — Trace each finding back to its `.mustache` template

For every `warning:`/`error:` line, find the template that generated the offending file — usually by matching on class/method names between the generated `.java` and the `.mustache` sources:

```bash
find src/main/resources/generator -iname "<GeneratedClassName>.java.mustache"
```

Read the template around the corresponding construct (same package layout, same line numbers minus the `{{ }}` placeholders) and confirm the match before editing.

## Step 5 — Classify and apply the fix

Not every ErrorProne finding gets the same treatment:

- **Real code fix** (visibility, `equals()` pattern, missing `@Override`, resource leak, locale-unsafe casing, etc.) → apply it in the template. Where possible, cross-check against an equivalent piece of Seed4J's *own* code (e.g. its own `RestSeed4JModuleBuilder`, `AssertionErrorMessagesTest`) to match the project's established idiom rather than inventing a new one.
- **Fix that only changes behavior via a `@SuppressWarnings(...)` annotation** → do **not** apply it. Since we don't control whether a given generated app will ever run ErrorProne, an annotation that exists solely to silence a specific bug-checker is dead weight in code that will very likely never be checked by that tool.
- **No safe fix exists without changing intended behavior** (e.g. `MemoizersTest`, where discarding the return value is deliberate) → leave untouched, unless a genuine behavior-preserving rewrite (not an annotation) is available.

### Worked example: `MemoizersTest.java.mustache`

`server/javatool/memoizers/test/MemoizersTest.java.mustache` triggers a `ReturnValueIgnored` **error** (not just a warning) at lines 36-37:

```java
memoizer.apply(1);
memoizer.apply(1);

assertThat(factory.callsCount()).isEqualTo(1);
assertThat(memoizer.apply(1)).isNull();
```

The two calls at lines 36-37 are made purely for their side effect (populating the memoization cache so `factory.callsCount()` can be asserted afterwards), and their return value is deliberately discarded.

Two candidate fixes were considered:

- **Option A — real rewrite (the one to prefer if this is ever fixed):** consume the return values instead of discarding them. This makes the test strictly more thorough (it now also asserts both calls return `null`) and needs no annotation:
  ```java
  assertThat(memoizer.apply(1)).isNull();
  assertThat(memoizer.apply(1)).isNull();

  assertThat(factory.callsCount()).isEqualTo(1);
  ```
- **Option B — mirror Seed4J's own precedent (rejected):** `src/test/java/com/seed4j/shared/memoizer/domain/MemoizersTest.java` — Seed4J's own test for its own `Memoizers` utility — is identical to the template line for line, except it carries `@SuppressWarnings("ReturnValueIgnored")` on the test method. Following that precedent would mean adding the same annotation to the template. This was rejected under the Step 5 rule: the annotation only has an effect if the generated app ever runs ErrorProne, which we can't assume, so it would just be dead code in every generated project.

**Decision:** the template was left untouched (per explicit instruction during this audit), with Option A recorded as the fix to apply if/when this file is revisited.

## Step 6 — Verify

Rebuild Seed4J's jar, restart the server, regenerate the app from the now-patched templates, and re-run Step 3 to confirm the warning count drops to only the intentionally-skipped items. Then run the generated project's own full build to catch any regression the template edit might have introduced:

```bash
cd /tmp/seed4j/<app-name>
./mvnw verify
```

(This step caught a real regression once: switching a `Builder`'s constructor to `private` without also marking the enclosing class `final` broke Checkstyle's `FinalClass` rule; and using `split(regex, -1)` instead of `split(regex, 0)` to silence `StringSplitter` silently changed behavior on blank input and broke a test.)

murdos (approved):

